PR #32, opened 2026-08-10
ci: point human-facing action-run links at the public forge URL

spdrman:
## Problem

The `gitea/ci` commit status this repo posts onto GitHub carried a `target_url`
of `http://gitea:3000/<org>/<repo>/actions/runs/<n>`. That host only exists on
the `gitea-net` Docker network, so the link is dead for whoever clicks it from
the pull request.

## Cause

One value was serving two different needs.

The act_runner registers against the internal `http://gitea:3000`, so inside a
job `${{ github.server_url }}` resolves to that address. **That is correct and
is not changed here.** Runner traffic stays on the Docker network, never leaves
the host and comes back through Caddy, and the runner keeps no inbound HTTP
surface.

It is simply not a URL a human can follow. Gitea's own links are already
correct, because `ROOT_URL` is `https://git.opsite.ca/`. I confirmed that
against the live API: `GET /repos/iasbuilt/webapp/actions/tasks` returns
`"url": "https://git.opsite.ca/iasbuilt/webapp/actions/runs/314"`. The bad link
was never emitted by Gitea, it was built here and posted to the GitHub
commit-status API.

## Change

Human-facing link sites now read a separate, explicitly public variable:

| Expression | Value | Use |
|---|---|---|
| `${{ github.server_url }}` | `http://gitea:3000` | API and git calls from inside a job |
| `${{ vars.GIT_PUBLIC_URL }}` | `https://git.opsite.ca` | anything a human clicks |

Only `target_url` sites changed. Every other `github.server_url` use stays
internal, and each switched site carries a comment saying which of the two it
is, so the next person does not helpfully unify them again.

## The variable

`GIT_PUBLIC_URL` is an **org-scoped Gitea Actions variable**, already live on
both `iasbuilt` and `libviprs`, so there is no per-repo setup and nothing to
configure before merging.

It is not called `GITEA_PUBLIC_URL` because Gitea reserves the `GITEA_` and
`GITHUB_` prefixes for Actions variable names and rejects them with HTTP 400
`invalid secret name`. Verified against this instance (1.24.7):

```
POST /api/v1/orgs/iasbuilt/actions/variables/GITEA_PUBLIC_URL -> 400 invalid secret name
POST /api/v1/orgs/iasbuilt/actions/variables/GIT_PUBLIC_URL   -> 204
```

Provisioning is codified in the IaC repo in `iasbuilt/IaC` PR #103, so a rebuild
recreates it.

## Not changed

- The runner registration / `gitea_act_runner_instance_url`.
- `GITEA__server__ROOT_URL`, which is already correct.

## Verification

YAML parses. The only remaining `github.server_url` mentions in the changed
files are inside the explanatory comments. Not waiting on CI, the shared runner
is backed up.

## Second defect fixed here

This repo also built the path from `${{ github.run_id }}` rather than
`${{ github.run_number }}`. Gitea's run URL is indexed by the run **number**,
not the internal id, so that link would 404 even once the host was corrected.
Evidence from this repo's latest run:

```
id=337  run_number=5  url=https://git.opsite.ca/libviprs/libviprs-dep/actions/runs/5
GET /libviprs/libviprs-dep/actions/runs/337 -> 404
```

Both `target_url` sites now use `run_number`.


spdrman:
Closing as superseded.

This edits `.gitea/workflows/*` to point human-facing action-run links at the public forge URL. **CI has moved to GitHub Actions wholesale** for every repo in the org (`libviprs/libviprs#585`), and `.gitea/` is now deleted here along with the mirror relay.

So the files this touches no longer exist, and the problem it solves cannot arise, because nothing reports from Gitea any more.

The instinct was right and worth recording: a CI link a human clicks should resolve for that human. That is now true by construction, since the runs live on GitHub alongside the repo.

Its siblings are closed for the same reason: `libviprs/libviprs#529` and `libviprs/libviprs-tests#136`.

Decision recorded rather than asked, per the standing instruction on this epic.

PR #35, opened 2026-08-27
zstd: build and publish static + shared libzstd

### LINKED ISSUE #34: zstd: build and publish static + shared libzstd, so libviprs stops compiling C under --features packfile
libviprs compiles the zstd C library from source on every build that enables
`--features packfile`, and it should come from here instead, the way PDFium does.

## What libviprs actually pulls today

Measured on `libviprs` @ `f62a56a` with `cargo tree`:

```
zstd-sys v2.0.16+zstd.1.5.7
└── zstd-safe v7.2.4
    └── zstd v0.13.3
        └── zip v7.2.0
            └── libviprs v0.4.0
```

`zstd-sys` is not in the default build (`cargo tree -i zstd-sys` with default
features reports no such package); it arrives with `--features packfile`, which
enables `dep:zip`. Its build script compiles the vendored zstd C sources and emits
`rustc-link-lib=static=zstd`, 216 object files.

It is the only true `-sys` crate in the tree. Two near-misses worth recording so
nobody re-derives them: `core-foundation-sys` only links macOS system frameworks
and compiles nothing, and `libbz2-rs-sys` is a **pure-Rust** reimplementation
despite the `-sys` suffix, so a rule keyed on crate names would reject it wrongly.
The reliable signal is a build script emitting `rustc-link-lib`.

## Why this one fits the existing pattern cleanly

`zstd-sys` already supports linking a prebuilt library. It has a `pkg-config`
feature and honours `ZSTD_SYS_USE_PKG_CONFIG`, with `rerun-if-env-changed`
wired up, so pointing it at an archive published from this repo needs no fork.

(For contrast, `blake3` cannot be done this way. Its build script has no
`pkg_config`, no external `rustc-link-lib`, and no environment variable naming a
prebuilt library, so it always compiles its own vendored C. That is a separate
decision and not part of this issue.)

## What to build

Follow `pdfium/` as the template: a `zstd/` directory with its own build driver,
`VERSION` as the single source of truth, README, verification script and pytest
suite, publishing archives that carry **both** a static archive and a shared
library per platform, across `{linux, musl} x {amd64, arm64}` plus macOS, matching
what `pdfium/README.md` documents.

zstd is far smaller than PDFium and needs no Chromium toolchain, so the build
itself is a plain CMake or Makefile invocation. The work is almost entirely in
matching this repo's conventions rather than in compiling the library.

## One thing that has to change outside `zstd/`

CI is hardcoded to the pdfium directory and will not see a new one:

```
.github/workflows/ci.yml   ruff check pdfium/ ; ruff format --check pdfium/
                           pytest pdfium/tests/ -v
                           shellcheck pdfium/patches/*.sh
pyproject.toml             testpaths = ["pdfium/tests"]
```

So a `zstd/` directory added without touching these lands completely untested. The
fix should make the paths derive from what is on disk rather than adding a second
hardcoded copy, so the third dependency cannot repeat this.

Wiring libviprs itself to consume the published archive is deliberately **not** in
scope here; that is a change to `libviprs`' own `Cargo.toml` and belongs in that
repo once these artifacts exist.

spdrman:
Closes #34.

libviprs recompiles the zstd C sources on every build that enables `--features packfile`, because `zstd-sys` vendors them. This adds a `zstd/` directory that publishes the library instead, built to the same shape as `pdfium/`: a build driver, `VERSION` as the single source of truth, a README, a verification script and a pytest suite, publishing one archive per platform carrying **both** `libzstd.a` and `libzstd.so` (`.dylib` on mac).

## What I actually built and verified

Every archive in the matrix was built end to end on this machine and passed `zstd/scripts/verify_archive.sh`. Nothing here is wired-but-unbuilt.

| Archive | Built via | Smoke test | Size |
| --- | --- | --- | --- |
| `zstd-linux-x64.tgz` | Docker, `linux/amd64` (QEMU) | ran | 772 KB |
| `zstd-linux-arm64.tgz` | Docker, `linux/arm64` (native) | ran | 729 KB |
| `zstd-musl-x64.tgz` | Docker, `linux/amd64` on Alpine (QEMU) | ran | 797 KB |
| `zstd-musl-arm64.tgz` | Docker, `linux/arm64` on Alpine (native) | ran | 760 KB |
| `zstd-mac-arm64.tgz` | native CMake on macOS 15 / Apple Silicon | ran | 559 KB |
| `zstd-mac-x64.tgz` | native CMake, `-arch x86_64` slice | link-only, no Rosetta | 688 KB |

"Smoke test ran" means the build compiled a compress/decompress round-trip against the staged static archive *and* against the staged shared library and executed both, checking `ZSTD_versionString()` against `VERSION`. Every combo printed `smoke: zstd 1.5.7 round-tripped 65536 -> 41 -> 65536 bytes` twice.

## Why 1.5.7

`zstd-sys v2.0.16+zstd.1.5.7` vendors zstd 1.5.7, so publishing that exact version makes adopting this a link-time change and nothing else. The reasoning lives in `zstd/README.md` next to the number, and `zstd/tests/test_zstd_version.py` fails if `VERSION`, the pinned sha256 and the README's download URLs ever disagree. The release tarball's sha256 is checked before anything compiles, and a version with no pinned hash is refused rather than downloaded unverified.

## What it deliberately does not inherit from pdfium

No Chromium checkout, no GN, no patches, no two-phase ninja, no memory scheduler, no ETA machinery. zstd is one CMake configure with `ZSTD_BUILD_STATIC=ON` and `ZSTD_BUILD_SHARED=ON`, and both libraries fall out of a single pass. Each combo builds in a container pinned to the **target** architecture (`FROM --platform=linux/arm64 …`) rather than cross-compiling, which is what lets the smoke test run the artifacts it just produced. That trade only works because the build is minutes: a QEMU-emulated PDFium build would not be.

What is kept is the structure: per-platform Docker builds, a staging directory packaged as a `.tgz`, verification before publishing, streamed per-job logs under `bin/logs/<plat>-<arch>.log`, and `--upload` through `gh` with `--clobber`.

## Verification, and proof it bites

`zstd/scripts/verify_archive.sh` runs over the packaged tarball and the driver runs it on every archive before `--upload` can publish one, so verification is not something only CI does. It asserts the specific thing rather than that a file exists:

- `libzstd.a` has `!<arch>` magic (never `!<thin>`), is large enough to be real, holds at least ten objects, and carries a symbol index that **defines** `ZSTD_compress`, `ZSTD_decompress`, `ZSTD_versionNumber`, `ZSTD_createCCtx`, `ZDICT_trainFromBuffer`.
- Every object inside the archive, and the shared library, is built for the architecture the filename claims, read straight out of ELF `e_machine` / Mach-O `cputype`, so an arm64 archive verifies on an x64 runner with no cross binutils.
- The shared library is `ET_DYN` / `MH_DYLIB` and exports the API, not an executable or a stray relocatable object.
- `libzstd.pc` is relocatable and carries no build-machine paths.

The script understands both GNU and BSD `ar`. That is not theoretical: the first run against a real mac archive rejected it with `unrecognised object magic '7a737464'` for all 30 members, because BSD `ar` stores long member names inside the member data (`#1/<len>`). The verifier caught a genuine defect in itself before any test did.

Every guard was mutated and watched go red. Highlights:

| Mutation | Result |
| --- | --- |
| Accept `!<thin>` archives | `test_thin_archive_rejected` — `AssertionError: assert 0 == 1` |
| `object_arch_ok()` always returns 0 | 3 failed: both wrong-architecture tests plus shellcheck |
| Symbol-index check weakened to "the file exists" | `test_missing_api_symbol_rejected` — the archive was still rejected, but for the *wrong reason* (`does not export` instead of `does not define`), and the message-specific assertion caught it |
| Remove the pkg-config relocatability check | `test_absolute_prefix_rejected`, again caught by the specific message |
| Remove the `ET_DYN`/`MH_DYLIB` check | `test_not_a_shared_object_rejected` |
| Truncate the staged `libzstd.a` inside the Dockerfile | `docker build` failed at `[15/16] RUN sh /tmp/smoke.sh` and the driver exited 1 with no archive produced |
| Tell the smoke test to expect 9.9.9 | `smoke: linked zstd is 1.5.7, expected 9.9.9`, exit 1 |
| Bump `VERSION` to 1.5.8 | 3 failed: no pinned hash, plus both README URL checks |
| Add a second `cmake --build` pass | `test_both_libraries_from_one_pass` |
| Drop `sha256sum -c -` from the Dockerfile | `test_checksum_is_enforced` |

## CI now finds new dependencies by itself

`ci.yml` and `pyproject.toml` named `pdfium/` in every step, so `zstd/` would have landed lint-free, test-free and shellcheck-free. Rather than adding a second hardcoded copy, the paths are discovered:

- `ruff check .` / `ruff format --check .` — ruff walks the tree and honours `.gitignore`
- bare `pytest` — `pyproject.toml` now sets no `testpaths`, only `norecursedirs = ["bin", …]` to keep collection out of unpacked upstream source trees. pytest exits 5 on an empty collection, so a discovery regression fails rather than passing vacuously
- `git ls-files '*.sh' | xargs shellcheck`, with an explicit empty-list guard, because `xargs` with no input would shellcheck zero files and exit 0

`tools/install-hooks.sh` gets the same treatment so the pre-commit hook keeps mirroring CI. `zstd/tests/test_ci_coverage.py` (13 tests) pins all of it, and each of those was mutated too: hardcoding ruff back to `pdfium/` fails 2, narrowing the shellcheck glob fails 3, restoring `testpaths = ["pdfium/tests"]` fails 2, reverting the hook fails 2, dropping the empty-discovery guard fails 1. It parses the workflow as text rather than with PyYAML on purpose — the test job installs pytest and nothing else, and a guard that `importorskip`s itself out of existence in CI is not a guard.

It also widened shellcheck's reach by accident-of-doing-it-properly: the old job checked one file (`pdfium/patches/*.sh`), and the new one checks all five tracked scripts, so `pdfium/build_mac_native.sh`, `pdfium/scripts/verify_archive.sh` and `tools/install-hooks.sh` are linted for the first time. All clean, no changes needed.

## One thing I found and folded in

`pdfium/tests/test_release_workflow.py` opens with `pytest.importorskip("yaml")`, and the CI test job installed only pytest — so those 13 tests skipped on every run while the job reported green. The first CI run on this branch collected `254 items / 1 skipped` for exactly that reason. One line (`pip install pytest pyyaml`) makes them real again; the run after that collects 269 items with nothing skipped.

The guard for it derives from the tree rather than naming pyyaml: `test_every_importorskip_is_installed_by_the_test_job` scans every `tests/` directory for `importorskip` and fails if anything a module skips itself over is missing from the test job's install line. Removing pyyaml from `ci.yml` again turns it red.

Gates: `ruff check .` clean, `ruff format --check .` 25 files formatted, `shellcheck` clean over all 5 tracked scripts, `pytest` 269 passed (166 pdfium, 103 new). CI on this branch is green on all four jobs.

## For whoever wires up libviprs

Out of scope here, but the archive is shaped to make it easy:

- `lib/pkgconfig/libzstd.pc` ships with `prefix=${pcfiledir}/../..`, so it resolves relative to wherever the archive is unpacked. `PKG_CONFIG_PATH=<dir>/lib/pkgconfig pkg-config --modversion libzstd` prints `1.5.7` against an unpacked archive — verified, not assumed. That is the form `zstd-sys`' `pkg-config` feature and `ZSTD_SYS_USE_PKG_CONFIG=1` consume.
- The library is built with `ZSTD_MULTITHREAD_SUPPORT=ON`, matching distro packages, so a static link needs `-lpthread`. `Libs.private` records it on linux; it is empty on mac because pthread lives in libSystem.
- `ZSTD_LEGACY_SUPPORT=OFF`, matching `zstd-sys`' default (no `legacy` feature), so v0.4–v0.7 frames are not decodable — same as today.
- Pick `musl-*` for Alpine and `linux-*` for glibc. Static-linking a glibc `libzstd.a` into a musl binary fails at link time.

## Not done, deliberately

zstd has no release workflow of its own. `release.yml` is pdfium-specific (its tag, its VERSION file, its mac universal job), and the runbook scoped my outside-`zstd/` changes to CI and the docs. The driver already does version resolution, verification and `gh release upload --clobber`, so adding one is a YAML job rather than new build logic. `README.md` says so where it documents cutting a release.


spdrman:
Reviewed and verified independently rather than read. Composing this onto current `main` (`76eca8d`, two commits newer than the base CI ran on) and running the full gate:

| check | result |
| --- | --- |
| `ruff check .` | All checks passed |
| `ruff format --check .` | 25 files already formatted |
| `shellcheck` over `git ls-files '*.sh'` | clean, 5 scripts |
| `pytest` | **269 passed**, nothing skipped |

269 with nothing skipped is the number this PR claims, and it is the one that matters: it is the evidence the `pyyaml` fix is real, since the 13 `test_release_workflow.py` tests were skipping silently before it. The only thing `main` has moved on since the base is `pdfium/README.md`, which this branch does not touch, so the composition is clean as well as green.

**Supply chain.** I fetched `https://github.com/facebook/zstd/releases/download/v1.5.7/zstd-1.5.7.tar.gz` and hashed it:

```
actual:  eb33e51f49a15e023950cd7825ca74a4a2b43db8354825ac24fc1b7ee09e6fa3
pinned:  eb33e51f49a15e023950cd7825ca74a4a2b43db8354825ac24fc1b7ee09e6fa3
```

It matches upstream. `source_sha256()` raising on an unpinned version rather than downloading unverified is the right default, and `test_zstd_version.py` holding `VERSION`, the hash and the README URLs together means the three cannot drift.

**The mutation table.** This org has had a lane publish a mutation table it had not run, with one row the opposite of what was printed, so I re-ran two rows rather than trusting the table:

- Accept `!<thin>` archives: `test_thin_archive_rejected` failed with `AssertionError: assert 0 == 1`. Matches.
- `object_arch_ok()` always returns 0: **3 failed**, exactly `test_object_of_the_wrong_architecture_rejected`, `test_wrong_architecture_rejected` and `test_passes_shellcheck` (SC2317, unreachable code). Matches, including the count and the shellcheck row.

Both reproduce as published. I am treating the rest of the table as good on that basis.

**On the CI discovery change**, which is the part with reach beyond `zstd/`: deriving the paths instead of adding a second hardcoded copy is the right call, and the three guards that make it stick (pytest exiting 5 on empty collection, the explicit empty-list guard before `xargs shellcheck`, and `test_every_importorskip_is_installed_by_the_test_job` deriving from the tree rather than naming pyyaml) each close a way the check could have gone quietly vacuous. That last one is the good one: it fixes the class rather than the instance.

## One thing to land straight after this, which is not a change request

Merging this reddens `libviprs-tests`' `Hook Mirror` job. That repo's `tools/install-hooks.sh` hardcodes this repo's CI commands:

```
"ruff check pdfium/"   "ruff format --check pdfium/"
"shellcheck pdfium/patches/*.sh"   "pytest pdfium/tests/ -v"
```

and exempts `pip install pytest`, which this PR turns into `pip install pytest pyyaml`. I ran that guard locally with all seven repos laid down and this branch as `libviprs-dep`, and `the_hook_mirrors_the_dep_ci` fails on the new multi-line shellcheck block.

I am merging anyway rather than blocking, for a specific reason: that job is **already red on `libviprs-tests` main** and has been for several merges, because `libviprs-org`'s `sync` job grew a multi-line block of its own. So this adds a second reason to an already-failing gate rather than breaking a green one. Nobody noticed because no repo in the org requires a status check to pass (libviprs-org#59). The companion PR fixing both halves goes up next, verified against this branch.

Merging.

Issue #47, opened 2026-09-12
G1.3: Implement the ACadSharp adapter and primitive flattening, and enforce the limits the ABI declares


spdrman:
Part of the ACadSharp NativeAOT epic in this repository (linked as a sub-issue). Map `LINE`, `POLYLINE`, `ARC`, `CIRCLE`, `ELLIPSE`, `SPLINE`, `TEXT`, `MTEXT`, `BLOCK` (as `INSERT` expansion), `DIMENSION` and basic `HATCH` onto the batch protocol frozen in the ABI issue, surface ACadSharp reader notifications as structured `Warning` records, and enforce the limits the ABI declares.

Block expansion is allowed for v1. If expansion causes unacceptable memory or output amplification on the hostile corpus below, extend the wire protocol with block-definition and instance records (a new wire version) rather than hiding the cost.

## The one rule that is easy to get wrong

Do not tessellate curves in the shim. Arc, circle, ellipse and spline cross the ABI as curves with their parameters. `libviprs` owns zoom and precision aware tessellation and is the only layer that knows either. A shim that tessellates has thrown away the information the consumer needs.

## Flattening

```
LINE       -> Line
POLYLINE   -> Polyline (LWPOLYLINE and 2D/3D POLYLINE alike; bulges stay as arcs, not as segments)
ARC        -> Arc
CIRCLE     -> Circle
ELLIPSE    -> Ellipse
SPLINE     -> Spline
INSERT     -> the block's visible primitives, transformed, depth-bounded
DIMENSION  -> Lines + Text from the dimension's block, or a Warning if it has none
HATCH      -> Polygon per boundary loop, or Lines for pattern-only hatches
TEXT/MTEXT -> Text
anything else -> Warning(UNSUPPORTED_ENTITY) carrying the entity type name and handle
```

Every ACadSharp reader notification (`DwgReader.Read(path, onNotification)` with `NotificationEventArgs`) becomes a `Warning` record with a code and the message. None is dropped.

## DWG version gate

The pinned ACadSharp reads AC1014 through AC1032 (the DwgReader column of upstream's README). A file whose six-byte version signature is outside that range returns `UNSUPPORTED_FORMAT` before any parsing, not `CORRUPT_INPUT` after. Craft the fixture: the bytes `AC1009` followed by zeros is enough, since ACadSharp's own writer cannot produce an AC1009 or AC1012 file.

## Corpus

[...]

PR #52, opened 2026-09-13
G1.2: Freeze the VIPRS CAD C ABI and the batch protocol

### LINKED ISSUE #46: G1.2: Freeze the VIPRS CAD C ABI and the batch protocol, with a C and a Rust conformance consumer
Part of the ACadSharp NativeAOT epic in this repository (linked as a sub-issue). This is the contract everything downstream is written against, including the `acadsharp-rs` crate that does not exist yet, so it is frozen here rather than allowed to evolve alongside the adapter.

## Deliverables

- `acadsharp/include/viprs_acadsharp.h`: the single source of truth for the ABI.
- `acadsharp/docs/ABI.md`: ownership, threading, error model, capability metadata and the fixed-width rules, written without reference to Rust so a second consumer could be built from it alone.
- `acadsharp/docs/WIRE.md`: the batch protocol.
- `acadsharp/tests/conformance/c/`: a C consumer built against the header.
- `acadsharp/tests/conformance/rust/`: a Rust consumer crate whose declarations are generated from, or tested against, the header rather than maintained by hand.
- Protocol parser tests on both sides of the boundary.

## The rules the header has to encode

A VIPRS-owned C ABI, never ACadSharp's object model. No `CadDocument*`, `CadEntity*`, `Layer*` or `Block*` crosses the boundary. C ABI only. Every `[UnmanagedCallersOnly]` export turns failure into a result code, and no managed exception escapes. Every struct opens with `uint32_t struct_size; uint32_t struct_version;` and otherwise holds only fixed-width scalars. Strings are written into caller-provided UTF-8 buffers with explicit byte lengths and a required-length out parameter.

Never use C or Rust enum layout as ABI, C or C# bool layout as ABI, pointer-sized integer fields other than opaque pointers, or implicit null termination.

Entry points, as in the design document:

```c
uint32_t viprs_acad_abi_version(void);
uint64_t viprs_acad_abi_fingerprint(void);
uint32_t viprs_acad_capabilities_v1(viprs_acad_capabilities_v1 *out, uint8_t *acadsharp_version_utf8, uint64_t cap, uint64_t *required);
uint32_t viprs_acad_open_path_utf8(const uint8_t *path, uint64_t path_len, const viprs_acad_limits_v1 *limits, viprs_cad_handle **out);
uint32_t viprs_acad_open_memory(const uint8_t *data, uint64_t data_len, const viprs_acad_limits_v1 *limits, viprs_cad_handle **out);
uint32_t viprs_acad_view_count(viprs_cad_handle *h, uint32_t *out_count);
uint32_t viprs_acad_view_info_v1(viprs_cad_handle *h, uint32_t index, viprs_view_info_v1 *out, uint8_t *name_utf8, uint64_t name_cap, uint64_t *name_required);
uint32_t viprs_acad_decode_begin(viprs_cad_handle *h, uint32_t view_index, const uint32_t *cancel_flag, viprs_decode_handle **out);
uint32_t viprs_acad_decode_next_batch(viprs_decode_handle *d, uint8_t *buf, uint64_t cap, uint64_t *written, uint8_t *done);
void     viprs_acad_decode_close(viprs_decode_handle *d);
void     viprs_acad_close(viprs_cad_handle *h);
```

Two things here go beyond the design document, and both are cheaper now than as a version bump later:

- `viprs_acad_limits_v1` (struct_size, struct_version, `max_input_bytes`, `max_entities`, `max_string_bytes`, `max_polyline_points`, `max_block_depth`, `max_output_bytes`), passed to both open calls; a null pointer means the documented defaults. Cancellation is a caller-owned `const uint32_t *cancel_flag` the decoder polls between batches. No callbacks into the caller in v1.
- `viprs_acad_capabilities_v1` carrying `dwg_version_min` and `dwg_version_max` as the numeric AC10xx codes (1014 and 1032 for ACadSharp 3.7.1; the README's DwgReader column is the source), plus the pinned ACadSharp version as a UTF-8 string through a caller buffer.

Result codes are plain `uint32_t` constants, not an enum: `0 OK`, `1 INVALID_ARGUMENT`, `2 UNSUPPORTED_FORMAT`, `3 CORRUPT_INPUT`, `4 UNSUPPORTED_ENTITY`, `5 OUT_OF_MEMORY`, `6 CANCELED`, `7 INTERNAL_ERROR`, `8 ABI_MISMATCH`, and `9 LIMIT_EXCEEDED`, which the design document lacks and the limits surface needs. Downstream never parses error strings for control flow.

The fingerprint is defined, not hand-assigned: the first eight bytes of the sha256 of the published header, big-endian. The driver generates `AbiFingerprint.g.cs` from the header at build time so `viprs_acad_abi_fingerprint()` returns the hash of the header it was built from. That is what lets both the verifier and a consumer detect a header that drifted from its library.

## The batch protocol

Fixed little-endian header: magic `VACB` (4 bytes), `uint16_t wire_version`, `uint16_t flags`, `uint32_t payload_length`. Then records, each `uint16_t type`, `uint16_t reserved`, `uint32_t length` (bytes, header included), payload. Unknown types are skipped by `length`. Target 64 KiB to 1 MiB per batch. Record types: `DocumentBegin`, `ViewBegin`, `Line`, `Polyline`, `Arc`, `Circle`, `Ellipse`, `Spline`, `Polygon`, `Text`, `Warning`, `ViewEnd`, `DocumentEnd`. Coordinates are `double` in drawing units. Curves carry their parameters (arc: centre, radius, start and end angle; ellipse: centre, major axis, ratio, parameter range; spline: degree, knots, control points, weights) and are never tessellated in the shim. `Text` carries position, height, rotation and UTF-8 bytes with a length. `Warning` carries a code, a UTF-8 message with a length, and an optional entity handle.

- Exceptions cannot escape: a test-only export (`viprs_acad__test_throw`, compiled only in the test configuration) throws a managed exception; the C consumer must receive `INTERNAL_ERROR` and keep running. Failing state: remove the catch-all in the export wrapper and the process aborts.
- Fingerprint drift: change one byte in the header, rebuild only the consumer, and the consumer's handshake must report `ABI_MISMATCH`. Separately, the CI check comparing `sha256sum acadsharp/include/viprs_acadsharp.h` with the value embedded in the shim must go red on the same edit. Observe both once before calling them gates.
- Parser: a synthetic batch with an unknown record type in the middle is skipped and the record after it is read; a `payload_length` past the buffer end is `CORRUPT_INPUT`; a record `length` smaller than its own header is rejected, not looped on.
- Layout: the Rust consumer asserts `size_of` and `offset_of` for every struct against values taken from the header (bindgen at build time, or a checked-in table with a test that regenerates it and diffs). Failing state: reorder two fields in the header and the test must name the struct.
- Arguments: every export called with null handles, null pointers or zero lengths returns `INVALID_ARGUMENT`; none faults.
- Presence of the limits and cancellation surface: `viprs_acad_open_memory` with a `limits` pointer whose `struct_size` is wrong returns `INVALID_ARGUMENT`; a `cancel_flag` set before the first `decode_next_batch` returns `CANCELED`. The semantic tests (a limit actually stopping a decode part-way) land with the adapter issue and are listed there.

Until the adapter exists, back the exports with a synthetic in-memory document selected by a magic byte sequence on `open_memory`, so every entry point and every parser path is exercisable from the C and Rust consumers on this issue alone.

## Acceptance

- [ ] `viprs_acadsharp.h` exposes no ACadSharp type and no managed pointer; `grep -c -E 'Cad(Document|Entity)|Layer\b|Block\b' acadsharp/include/viprs_acadsharp.h` is 0.
- [ ] No managed exception escapes an export, shown by the test above.
- [ ] Protocol parser tests exist on both sides, including the unknown-type skip and the two malformed-length rejections.
- [ ] The C conformance consumer compiles and runs against the published header in CI.
- [ ] The Rust conformance consumer compiles and runs against the same header, and its declarations are generated or layout-tested rather than hand-maintained.
- [ ] Header hash and ABI fingerprint drift fails CI, and that failure has been observed once by editing the header.
- [ ] `ABI.md` and `WIRE.md` describe the contract without reference to Rust.
- [ ] `viprs_acad_limits_v1`, `cancel_flag`, `viprs_acad_capabilities_v1` and `LIMIT_EXCEEDED` are in the header.
- [ ] Nothing here names a Windows target or toolchain.

## Couplings

The packaging issue ships this header in `include/` and records its sha256 and fingerprint in `LINKINFO.json`; the `acadsharp-rs` epic consumes those two fields and the `abi_version` at first use, so their names must not change once this lands. If the conformance consumers become a `ci.yml` job here, that job needs the libviprs-tests Hook Mirror pairing (mirror or defer it in `tools/install-hooks.sh`), landed right behind, because CI is required on both repos.

spdrman:
Closes #46

The contract everything downstream is written against, plus two consumers that call it from outside the runtime. `acadsharp-rs` does not exist yet, so the ABI is frozen here rather than left to evolve alongside whatever adapter happens to back it.

## The header changed, and it had to

**`viprs_acadsharp.h` did not compile as C.** `viprs_acad_capabilities_v1` is both a struct and an entry point, and in C a typedef name and a function name are the same kind of identifier, so

```c
typedef struct viprs_acad_capabilities_v1 { ... } viprs_acad_capabilities_v1;
uint32_t viprs_acad_capabilities_v1(viprs_acad_capabilities_v1 *out, ...);
```

is rejected outright: `error: 'viprs_acad_capabilities_v1' redeclared as different kind of symbol`. It compiles as C++, where the function name hides the class name, which is how a header ships broken. Everyone who tried it tried it the wrong way, me included, until I pointed a C compiler at it.

The fix keeps both names the issue freezes and drops the typedef, so the three data structs are tags and a consumer writes `struct viprs_acad_limits_v1`. I did all three rather than only the one that collides, because a contract where you have to remember which struct needs the keyword is a contract somebody gets wrong. The opaque handles keep their typedefs: no fields, and no call shares their name. `test_abi_header.py` grew three cases so it cannot come back.

That moved the fingerprint, which is the point of having one. Both recorded constants moved with it in the same commit.

## The fingerprint is generated, and I watched it track the header

An MSBuild target hashes `include/viprs_acadsharp.h` and writes `AbiFingerprint.g.cs` into the intermediate directory. Nothing is checked in, nothing is typed. When I made the change above, the value went from `0xD855AA37BEEE7ACB` to `0x03BB8E542F149EC0` on the next build with nobody touching a number.

## Neither consumer carries its own copy of the boundary

The C one includes the published header and asserts `sizeof`, `_Alignof` and `offsetof` for every struct through `_Static_assert`. The other generates every declaration from the same file in `build.rs`, computes its own sha256 there, and asserts `size_of` and `offset_of` for every struct. It has no dependencies, so a conformance run is offline and has no supply chain to audit.

The layout tables are written out by hand on purpose. Generated alongside the struct definitions, a reordered field would move both and the test would pass.

## Every mutation I ran, and what it printed

All of it in containers on linux/arm64. Nothing touched a host toolchain.

| Removed | What happened |
| --- | --- |
| The catch-all in `viprs_acad__test_throw` | `Unhandled exception. System.InvalidOperationException: thrown on purpose by the test export / at Exports.TestThrow(UInt32) + 0x44 / Aborted`. Consumer exit **134**, output cut off mid-line. Restored: the same five calls come back as `INTERNAL_ERROR` and the process keeps going. |
| One byte appended to the header, consumers rebuilt, shim not | C: `fingerprint=0x03BB8E542F149EC0 expected=0x62E70348DF8DE05C`, `FAIL fingerprint handshake: VIPRS_ACAD_ABI_MISMATCH (8)`, exit **8**. Generated consumer: identical, exit **8**. `pytest acadsharp/tests/test_abi_fingerprint.py`: **2 failed, 9 passed**, with no compiler present. |
| `dwg_version_min` and `dwg_version_max` swapped (same width, struct size unchanged) | Generated consumer: `offset_of!(viprs_acad_capabilities_v1, dwg_version_min) moved, so viprs_acad_capabilities_v1 was reordered / left: 20 / right: 16`. C: `static assertion failed: "offsetof(struct viprs_acad_capabilities_v1, dwg_version_min) is not 16, so viprs_acad_capabilities_v1 was reordered"`. `pytest test_abi_layout.py`: **3 failed**, including the managed struct. |
| `length < 8` in the C parser | **104 checks, 4 failures.** Lengths 0 and 4 stop being refused; 7 is still caught by the multiple-of-four rule. |
| The `payload_length` bound in the C parser, widened so it never fires | **104 checks, 1 failure**: `FAIL a payload_length past the buffer end is CORRUPT_INPUT`. |
| `length < 8` in the generated consumer's parser | `panicked at src/wire.rs: slice index starts at 8 but ends at 0`, exit **101**. Refusing is what keeps that from being a panic. |
| The `struct_size` check in `viprs_acad_capabilities_v1` | **104 checks, 1 failure**: `FAIL a struct_size this build does not know is INVALID_ARGUMENT`. |
| The cancel-flag poll in `decode_next_batch` | **104 checks, 2 failures**: the flag set before the first call stops being `CANCELED`, and the call writes a batch. |
| The null-buffer and zero-capacity check in `decode_next_batch` | **104 checks, 2 failures**. Worth reading twice: it does **not** fault. The runtime turns the write through a null into a managed exception and the catch-all makes it `INTERNAL_ERROR`, so the second line of defence is real. The check still earns its place, because `INVALID_ARGUMENT` and `INTERNAL_ERROR` are different promises, and a caller that switches on the number is told a different thing. |

Unmutated, the C consumer is **108 checks, 0 failures** and the generated one **88 checks, 0 failures**. The repository gate is **490 passed, 22 skipped**, with `ruff check`, `ruff format --check` and `tools/shellcheck-all.sh` clean, all four in containers.

Two of those rows are only red because of something the mutation run itself found, which is the argument for doing it: the short-record case in both consumers called the parser once, and with the rule removed one call returns a record and nothing looks wrong. The cursor only fails to advance on the call after it. Both cases drain now, through a bounded loop whose counter is the test's own and not the parser's.

## Three things I folded in rather than filing

The per-record byte counts in WIRE.md were payload sizes written in a field that means the whole record: `Line` said 64 and is 72, `Arc` said 88 and is 96, `Circle` 72 and is 80. They are totals now, and both consumers check the bytes the library actually emits against the same table, so a payload list nobody compares against the producer cannot drift again.

A batch does not always stay under 1 MiB. A batch never splits a record, and a `Polyline` at the default `max_polyline_points` is 24 MB on its own, so an oversized record becomes a batch of its own. Better written down than discovered.

And `open_path_utf8` had no case that opened anything. It has a code path of its own, sniffing the file rather than a caller's buffer, and the only things running it were the four ways it refuses. A route exercised only by its failures is a route nobody has run.

## The configuration is called AbiTest, not Test

That cost a build to work out. The SDK turns the configuration name into a preprocessor symbol, the vendored upstream source carries `#if TEST` regions that do not compile on their own, and `dotnet publish -c Test` fails inside a dependency with four `CS0246`s that never mention configurations.

## The seam for #47

`IDocumentSource` is six members and `SourceFactory` has one branch per source, so the adapter issue adds a line to one small file rather than editing something two lanes touched. Primitives are built through named factories rather than assembled as bags of doubles, because a source that gets the field order wrong is not something either side's compiler can see.

Handles are counter-issued identifiers cast to a pointer, never addresses. Both consumers pass `0xDEAD` and get `INVALID_ARGUMENT` instead of a fault.

## Adversarial review round: four findings fixed

**C3, Critical. `decode_next_batch` wrote twelve bytes past a short buffer and returned OK.**
`NextBatch` bounded every record against `cap` and then wrote its batch header with no check at all. A call with `cap` below twelve wrote past the caller's buffer and returned `VIPRS_ACAD_OK` with `*written` 12, so a consumer taking a slice of `*written` read out of bounds on top of the write that had already happened. The trigger is a call after `*done` came back 1, which the header neither forbade nor documented.

`cap` below twelve is now `VIPRS_ACAD_LIMIT_EXCEEDED` with twelve in `*written` and nothing written. A call after `done` is legal: it writes an empty final batch and reports `done` again, so a loop that asks once too often terminates instead of failing. I rejected succeeding with `*written` 0, because then a successful call sometimes produces bytes a parser can read and sometimes does not, and every consumer would have to learn the difference. Those empty batches are also not charged against `max_output_bytes`, or "calling after done is legal" would be a rule with an expiry date. All of it is in the header.

**H1, High. `Handles.Remove` evicted before it cast, so a wrong-typed close orphaned the object.**
`decode_close` called with a document handle found the document, took it out of the table, failed the cast and did nothing. Every decode that document tracked stayed alive holding the whole parsed drawing, and the later `close` found nothing and silently did nothing. From outside it looked exactly like a close that worked. It casts first and evicts only on success now, which is the guard `Get<T>` already had, in the right order.

**L2. The cancel flag is read with `Volatile.Read`.**
It happens once per call today so nothing can hoist it, but the moment anything reads it inside a loop a plain load can be read once and reused for the rest of the decode.

**B-2, High. Nothing read a scalar out of a record payload.**
Both parsers walked record headers and never looked inside, so swapping an arc's `radius` with its `start_angle` changed nothing any test could see, in any branch. The synthetic document now carries placement probes rather than geometry: the k-th `double` after the prologue is `100 * type + k + 0.25`, so no field of a record holds the same number as another, none holds zero, and none holds a value that would look right one slot over. Both consumers read every documented field at its documented offset, check the variable-length invariants, and check the stream's own record and warning totals against their own tallies. ABI.md now says those numbers are contractual, which is the opposite of what it said before.

**One I found while fixing C3.** The header said any `cap` below twelve was `LIMIT_EXCEEDED`, but `cap` 0 returns `INVALID_ARGUMENT` under the zero-length rule and both consumers already asserted that. The prose said one thing and the code did another; the header now separates the two cases.

### What each fix printed when I took it back out

| Removed | C consumer | Generated consumer |
| --- | --- | --- |
| the `cap >= 12` refusal in `NextBatch` | **9 failures** | **9 failures** |
| the already-finished short-circuit | **1 failure**: a thousand calls after `done` stop being OK once they are charged to `max_output_bytes` | **1 failure**, same |
| the cast-before-evict order in `Handles.Remove` | **2 failures**: neither wrong-typed close leaves the handle alone | **2 failures**, same |
| `radius` and `start_angle` swapped in the Arc encoder | **1 failure**: `Arc: got 504.250000, wanted 503.250000` | **1 failure**: `Arc: got 504.25, wanted 503.25` |

That last row is the one the reviewer predicted would stay green in all four branches. It does not any more.

**And the handle mutation found a hole in the test I had just written for it.** The live-handle count walks back down to the baseline under the bug as well, because the wrong-typed close evicts the handle without disposing it, so my final "nothing was orphaned" check passed against the broken build. The count is necessary and not sufficient. What tells the two apart is whether the handle still works after a close that was supposed to do nothing, so both consumers now call `view_count` on the document after a wrong `decode_close`, and `decode_next_batch` on the decode after a wrong `close`. That is the third time a mutation has caught a weak assertion of mine on this branch, and the first two are in the history for the same reason.

## What is deliberately not here

No job in `.github/workflows/ci.yml`. A job that compiles these needs the libviprs-tests Hook Mirror pairing that ADR 0001 already declined for the managed side, and `test_acadsharp_ci_decision.py` fails the moment somebody adds one without revisiting the decision. The drift gate the issue wants in CI is in `pytest -v` instead, which is already required: `test_abi_fingerprint.py` records the header's sha256 and reds on a one-byte edit with no compiler present.

## Running the consumers

```sh
# publish the shim, in a container
docker run --rm --platform linux/arm64 -v "$PWD":"$PWD" -v /path/to/ACadSharp-3.7.1:/acadsrc \
  -w "$PWD/acadsharp/native" <sdk-image-with-clang> \
  dotnet publish Viprs.ACadSharp.Native.csproj -r linux-arm64 -c AbiTest \
    -p:AcadSharpProject=/acadsrc/src/ACadSharp/ACadSharp.csproj

acadsharp/tests/conformance/c/run.sh
acadsharp/tests/conformance/rust/run.sh
```

Both scripts take `VIPRS_LIB_DIR`, `VIPRS_CONFORMANCE_IMAGE` and `VIPRS_CONFORMANCE_PLATFORM`, and both skip the exception section when the library they are given has no test exports in it.


spdrman:
Three things folded into this PR rather than filed, since they turned up while proving the thing works.

**The header did not compile as C.** `viprs_acad_capabilities_v1` is both a struct and an entry point, and in C a typedef name and a function name are the same kind of identifier. The frozen header typedef'd the struct to its own name, so a C compiler rejects the pair outright. It compiles as C++, where the function hides the class name, which is exactly how a header ships broken. I dropped the typedef on all three data structs rather than only the one that collides, because a contract where you have to remember which struct needs the keyword is a contract somebody gets wrong. Both recorded constants moved with it, and the header's own rule list and `test_abi_header.py` now carry the reason.

**Two assertions I wrote could not fail.** The short-record case in both consumers called the parser once. With the `length < 8` rule removed, one call returns a record and nothing looks wrong: the cursor only fails to advance on the call after it. Both cases drain now, through a bounded loop whose counter belongs to the test rather than to the parser, and the same mutation reds four checks where it previously reds none.

**The per-record byte counts in WIRE.md were payload sizes** written in a field that means the whole record: `Line` said 64 and is 72, `Arc` said 88 and is 96, `Circle` 72 and is 80. They are totals now, and both consumers check the bytes the library actually emits against the same table. While fixing that I also had to admit a batch does not always stay under 1 MiB, because a batch never splits a record and a `Polyline` at the default `max_polyline_points` is 24 MB on its own.

And one gap the mutation pass exposed by omission: `open_path_utf8` had no case that opened anything. It sniffs the file rather than a caller's buffer, so it has a code path of its own, and the only things running it were its four refusals.

One thing deliberately left out: there is no job in `.github/workflows/ci.yml`. A job that compiles the consumers engages the libviprs-tests Hook Mirror pairing that ADR 0001 already declined for the managed side, and `test_acadsharp_ci_decision.py` fails the moment somebody adds one without revisiting that decision. The drift gate the issue wants in CI is in `pytest -v` instead, which is already required: `test_abi_fingerprint.py` records the header's sha256 and goes red on a one-byte edit with no compiler present. I observed both halves, and the PR body has what each printed.

spdrman:
All four findings fixed, each mutation-tested, plus two things the mutation run found in my own work.

**C3** is fixed by refusing a `cap` below twelve rather than writing the header anyway. I also had to decide what a call after `done` does, and the header now says: it writes an empty final batch and reports `done` again, rather than succeeding with `*written` 0. A zero-length success would mean a successful call sometimes produces bytes a parser can read and sometimes does not, and every consumer would have to learn the difference. Those empty batches are not charged to `max_output_bytes` either, or "calling after done is legal" would have an expiry date at about six hundred calls.

**H1** casts before evicting now, which is the guard `Get<T>` already had, in the right order.

**L2** reads the flag with `Volatile.Read`.

**B-2** is the biggest change. The synthetic document now carries placement probes instead of geometry, the k-th `double` after the prologue being `100 * type + k + 0.25`, and both consumers read every documented field at its documented offset, check the variable-length invariants, and check the stream's own record and warning totals against tallies they keep themselves. The reviewer's exact prediction, swapping `radius` and `start_angle` in the Arc encoder, now prints `Arc: got 504.25, wanted 503.25` in both consumers.

Two things I found on the way, both folded in rather than filed:

The header claimed any `cap` below twelve was `LIMIT_EXCEEDED`, but `cap` 0 is `INVALID_ARGUMENT` under the zero-length rule and both consumers already asserted it. The prose and the code disagreed, in the file that is the contract.

And the wrong-typed-close test I wrote for H1 could not fail on its most important check. Putting the bug back reds two of its assertions, but the last one, the one claiming nothing was orphaned, stayed green: under the bug the wrong close evicts the handle without disposing it, so the live count still returns to baseline while both objects are stranded. The count is necessary and not sufficient. Both consumers now check the handle still works after a close that was supposed to do nothing, which is what actually separates the two.

Header edits moved the fingerprint twice, and both recorded constants moved with it in the same commits. It is `0x765D22C58CF48B52` now, generated as always.

Issue #48, opened 2026-09-12
G1.4: Build, package and verify the ACadSharp native archives across the five targets, and freeze LINKINFO.json


spdrman:
Part of the ACadSharp NativeAOT epic in this repository (linked as a sub-issue). Extend `acadsharp/build_acadsharp.py` from the spike's single target to the whole matrix, package each target the way `pdfium/` and `zstd/` do, write the two manifests the `acadsharp-rs` crate will consume, and ship a verifier that has been watched refusing broken archives.

## Targets, and this is the whole matrix

| archive | Rust triple in `LINKINFO.json` |
|---|---|
| `acadsharp-linux-x64.tgz` | `x86_64-unknown-linux-gnu` |
| `acadsharp-linux-arm64.tgz` | `aarch64-unknown-linux-gnu` |
| `acadsharp-musl-x64.tgz` | `x86_64-unknown-linux-musl` |
| `acadsharp-musl-arm64.tgz` | `aarch64-unknown-linux-musl` |
| `acadsharp-mac-arm64.tgz` | `aarch64-apple-darwin` |

A target may ship shared-only where static is not safely supportable. That is a recorded outcome in the manifest, not a failure.

The archive names follow this repo's convention, not the design document's `acadsharp-native-<version>-<triple>` form. `pdfium/tests/test_naming.py` and `zstd/tests/test_zstd_naming.py` pin `<dep>-<platform>-<cpu>.tgz` with platform in `linux | musl | mac` and cpu in `x64 | arm64`, extracting to a directory of the same name, and `zstd/scripts/verify_archive.sh` infers platform and cpu from that filename. A third scheme would break the repo's own tests and its verifier convention. The version lives in the release tag (`acadsharp-<VERSION>`), and the Rust triple lives inside the manifest where the consumer reads it.

## Archive layout

```
acadsharp-<platform>-<cpu>/
  lib/
    libacadsharp_native.a           # present when static was built; certified or not is in LINKINFO
    libacadsharp_native.so          # .dylib on mac
  include/
    viprs_acadsharp.h
  metadata/
    BUILDINFO.json
    LINKINFO.json
    CHECKSUMS.txt                   # sha256 of every other file in the archive
  LICENSES/
    ACadSharp-LICENSE               # MIT, Copyright (c) 2021 Albert Domenech
    THIRD_PARTY_NOTICES             # .NET runtime and any NuGet package the publish pulled in
  README.md
```

[...]

Issue #43, opened 2026-09-12
zstd/README.md links a zstd-1.5.7 release that does not exist, and nothing publishes it


spdrman:
#34 asked for zstd to be built and published the way pdfium is, and #35 landed the build side: `zstd/build_zstd.py`, `zstd/scripts/verify_archive.sh`, `zstd/VERSION` (`1.5.7`) and the tests. The publish side is documented and not done:

- `zstd/README.md:21-29` says "Archives are attached to the [`zstd-1.5.7`](https://github.com/libviprs/libviprs-dep/releases/tag/zstd-1.5.7) release" and lists six download URLs under `releases/download/zstd-1.5.7/`. `gh release view zstd-1.5.7 -R libviprs/libviprs-dep` returns "release not found", and `gh release list` shows only the pdfium and musl-cross releases. All six URLs 404.
- `.github/workflows/release.yml` builds and publishes pdfium only: it resolves `pdfium/VERSION`, tags `pdfium-<version>`, runs `pdfium/build_pdfium.py`. There is no zstd job in it or in `build.yml`.
- `python3 zstd/build_zstd.py --upload` (README line 101; the driver's `upload_release` at about line 867) is the only publish path, and it is a local command that needs the `gh` CLI and a macOS host for the two mac slices.

The README is the consumer's entry point and every link on it is dead. libviprs, the consumer #34 was for, still compiles zstd from `zstd-sys` under `--features packfile`; nothing there points at this repo yet, so nothing is broken downstream, but #34's "done" is a build script nobody has run to completion.

Pick one, and say which in the PR:

- Publish. Run `python3 zstd/build_zstd.py --parallel` for the four Linux and musl cells in Docker, the two mac cells on a Mac, then `--upload`. Confirm the six assets and their sha256s on the release page, and put the digests into `zstd/README.md` next to the URLs the way `pdfium/README.md` does. Then add a zstd path to `release.yml` (or a `release-zstd.yml`) so the next version bump is a `release` branch push like pdfium's rather than a laptop ritual. Keep the pdfium jobs' shape: resolve `zstd/VERSION`, create the `zstd-<version>` release, build per cell, verify, upload.
- Or strip. Remove the release link and the six URLs from `zstd/README.md`, say the archives are not published yet, and reopen the publish half as its own issue, with the consuming change in libviprs (`ZSTD_SYS_USE_PKG_CONFIG` pointed at the archive) listed as the follow-up.

[...]

PR #53, opened 2026-09-13
G1.5: Publish the ACadSharp NativeAOT release through its own workflow

### LINKED ISSUE #49: G1.5: Publish the ACadSharp NativeAOT release through its own workflow on the release branch
Part of the ACadSharp NativeAOT epic in this repository (linked as a sub-issue). Publish a pinned ACadSharp NativeAOT release so a consumer can resolve it without a floating dependency, through a workflow of its own.

## The workflow

`.github/workflows/release-acadsharp.yml`. Not an extension of `release.yml`, which is pdfium-shaped end to end (chromium branch input, a lipo'd universal job, its own summary table). PR #44 adds `release-zstd.yml` beside it for exactly this reason, so one dependency's flake cannot redden another's release, and it is the shape to copy: `resolve-version`, a version preflight that refuses an unpinned version before anything builds, `create-release` up front so the fan-out never races on `gh release create`, per-cell build then verify then upload, `--clobber` on upload so a re-run replaces its own assets, and a `concurrency` group per ref with `cancel-in-progress: false`. If #44 has not merged when this starts, copy its shape from that branch anyway; the per-dependency split is the agreed pattern.

Triggers: `push` to `release` and `workflow_dispatch` with an optional `version` input. Never `push` to `main`.

Tag: `acadsharp-<VERSION>`, for example `acadsharp-3.7.1-viprs.1`, from `acadsharp/VERSION` through the driver's `release_tag()`.

Cells: the five archives the packaging issue defines. Linux and musl cells run where the spike's ADR decided (native `ubuntu-24.04-arm` runners, cross-compilation, or emulation); the mac cell runs on `macos-15`. `fail-fast: false`, so one cell's failure does not discard another's archive.

## Preflight, before any build job

- `acadsharp/VERSION` parses as `<upstream>-viprs.<n>`; the upstream half has a `SOURCE_SHA256` entry in the driver; an empty or malformed file fails here.
- The .NET SDK `global.json` asks for is installable on the runner (`actions/setup-dotnet` with `global-json-file`); a missing SDK line fails here, not in five build jobs.

## Release notes, generated

From the manifests, not typed: ACadSharp version and commit, .NET SDK version, the supported DWG read range (AC1014 to AC1032 for 3.7.1), one line per target with `static_certified` and the archive's sha256, and, in a final job with `if: always()` that reads the matrix results, one line per target that did not publish, naming the job. A release that quietly ships four of five targets is worse than one that ships four and says so.

## README

`acadsharp/README.md` gets the download URLs and their digests only once the release exists, and a test fails if a URL appears without its digest or a digest without its URL. #43 is the standing example of the failure this prevents: `zstd/README.md` advertised six URLs for a release that had never been cut, and nothing checked. Do not repeat it in the other direction either, by publishing and leaving the README silent.

`acadsharp/tests/test_release_workflow.py`, with the same classes as the zstd test PR #44 adds (`TestWorkflowShape`, `TestTriggers`, `TestVersionPreflight`, `TestBuildJobsVerifyBeforeTheyUpload`, `TestMatrixCoversEveryArchive`, `TestUploadsGoToTheResolvedTag`), asserting:

- the workflow parses, and every job that uploads has `contents: write`;
- it fires on a push to `release`, can be dispatched, and does not fire on `main`. Failing state: add `main` to the trigger list in a scratch copy and the test names it;
- the version comes from `acadsharp/VERSION`, the tag matches `release_tag()`, and the unpinned-version refusal runs in `resolve-version`, before any job that needs it. Failing state: move that step into a build job and the ordering assertion fails;
- every build cell runs build, then `acadsharp/scripts/verify_archive.sh` on the named archive, then upload, in that order, and the build step does not pass `--upload` to the driver. Failing state: swap verify and upload in a scratch copy;
- the matrix cells are exactly the driver's default matrix plus mac, each carrying the cpu the archive name uses;
- uploads go to the tag `resolve-version` computed, and `create-release` runs before the fan-out.

On the runner, once, before this issue closes: dispatch the workflow with the `version` input on a throwaway pre-release tag (for example `3.7.1-viprs.0`), watch all five cells publish, download each asset on a clean machine with no .NET and run the C smoke against it, then delete the throwaway release and tag. That is the definition of done for the whole epic exercised end to end, and the reason this issue is last.

## Acceptance

- [ ] `release-acadsharp.yml` exists, triggers only on `release` and dispatch, and `test_release_workflow.py` pins every point above.
- [ ] A `VERSION` with no pinned upstream digest, or an uninstallable SDK pin, fails in `resolve-version` before any build.
- [ ] Every cell builds, verifies, then uploads, and a failed cell does not cancel the others.
- [ ] The release notes are generated from the manifests and name every target that did not publish.
- [ ] `acadsharp/README.md` lists the six URLs with digests, and a test holds URLs and digests together.
- [ ] A throwaway release was cut end to end, its assets ran the smoke on a machine with no .NET, and it was then deleted.
- [ ] `MANUAL.md`'s "Build and publish a release" section covers acadsharp.
- [ ] Nothing here names a Windows target, artifact or toolchain.

## Couplings

This is a new workflow file, not `ci.yml`, so the libviprs-tests Hook Mirror job does not read it and no pairing is needed. The `acadsharp-rs` epic's release lane tests "all targets published by libviprs-dep", so the asset names and the tag format are part of that crate's contract from the first real release on. The consuming change in `libviprs` and the `acadsharp-rs` compatibility gate are both outside this repo and are not filed until a real release exists.

spdrman:
Publishes the ACadSharp NativeAOT archives through a workflow of their own, on the same `release`-branch trigger pdfium and zstd already use.

`release.yml` is pdfium-shaped end to end: a chromium branch input, a `lipo`'d universal job, its own summary table. There is nothing in it acadsharp could reuse, and an acadsharp flake reddening a pdfium release run is exactly what PR #44 split `release-zstd.yml` out to avoid. So this is a third file with the same shape, not a fourth branch inside the first.

Closes #49

## The job graph

```
resolve-version ──┬── create-release ──┬── build-linux  (4 cells, fail-fast: false)
                  │                    │     linux/x64    ubuntu-latest
                  │                    │     linux/arm64  ubuntu-24.04-arm
                  │                    │     musl/x64     ubuntu-latest
                  │                    │     musl/arm64   ubuntu-24.04-arm
                  │                    │
                  │                    └── build-mac    (1 cell)
                  │                          mac/arm64   macos-15
                  │
                  └────────────────────────────────────────► release-notes  (if: always())
```

Each build cell runs build, then `verify_archive.sh` on the archive it just made with the cell's platform and cpu passed in, then `gh release upload --clobber`. `--upload` is deliberately not passed to the driver: that would publish before the verifier looked at it, which makes the verification decorative. `concurrency` is per ref with `cancel-in-progress: false`, because a run halfway through five uploads is worth finishing.

## Three refusals, all in `resolve-version`

A malformed `acadsharp/VERSION` (through `split_version`), an upstream half with no `SOURCE_SHA256` entry (through `source_sha256`), and a `global.json` that either names no `sdk.version` or names an SDK the runner cannot install (an explicit read, then `actions/setup-dotnet` with `global-json-file`). All three would otherwise be discovered five times over, on five claimed runners, after the tag already existed. `resolve-version` gets no `permissions:` block at all: it publishes nothing, so it holds no token that could.

## Nothing is emulated

This is the one place the zstd shape is deliberately not copied. `release-zstd.yml` registers `setup-qemu-action` and builds foreign-architecture cells under emulation. ADR 0001 measured a cross-architecture publish here producing the object file and then failing at the native link with `unrecognised emulation mode: aarch64linux`, and recorded that the .NET runtime documents `qemu-user-static` as unsupported. So arm64 cells take `ubuntu-24.04-arm`, and `test_no_cell_is_emulated` refuses an emulator if the next person to copy that file brings one back. Without that guard the copy would have silently reversed a measured decision.

## Notes are generated, never typed

`create-release` writes a preamble from `acadsharp/VERSION`, `acadsharp/native/global.json`, `acadsharp/include/viprs_acadsharp.h` and `acadsharp/build_acadsharp.py`, and hands it to `release-notes` as an artifact so the generator exists once rather than in two `run:` blocks that can drift. `release-notes` runs whatever happened above, downloads what actually reached the release page, hashes each asset and reads `static_certified`, the Rust triple and the ACadSharp commit out of each archive's `metadata/LINKINFO.json`. Every target that did not publish gets a line naming the job that should have produced it.

`test_nothing_version_bearing_is_typed_into_the_workflow` reads the current version, the SDK pin, the source digest and the read range out of the manifests and asserts none of them appears as a literal in the workflow, so bumping a pin cannot leave the release page stating the old one.

## What the tests were watched failing at

29 mutations, each applied, run, then restored. The first 21 were run again against the composed tree after the flag fix, since it touches the steps they exercise. What each printed:

| mutation | test that went red |
| --- | --- |
| add `main` to the trigger list | `test_does_not_fire_on_main`: *publishing must not happen on a push to main*; `test_it_publishes_from_exactly_one_branch`: *release is the only branch a publish comes from* |
| move the unpinned-version refusal into `build-linux` | 4 red, including `test_no_build_job_carries_the_refusal_itself`: *build-linux carries the unpinned-version refusal. It belongs in resolve-version, before anything builds*, and `test_the_refusal_cannot_be_switched_off`: *no step in resolve-version calls source_sha256, so there is no refusal to switch off and this guard is reading nothing* |
| swap Verify and Upload | `test_build_then_verify_then_upload[build-linux]`: *expected Build < Verify < Upload, got 2/4/3* |
| add `--upload` to the Build step | `test_the_build_step_does_not_upload[build-linux]`: *--upload inside the Build step publishes before Verify runs, which makes the verification decorative* |
| drop `contents: write` from `build-linux` | `test_every_publishing_job_can_write_releases`: *build-linux calls `gh release`, so it needs permissions.contents: write* |
| delete the `musl/arm64` cell | 3 red, including `test_the_notes_job_knows_every_matrix_cell_and_whose_job_it_is`: *TARGETS list and the build matrices disagree* |
| move `linux/arm64` to `ubuntu-latest` and add the QEMU action | `test_every_arm64_container_cell_runs_on_an_arm64_runner` and `test_no_cell_is_emulated`: *ADR 0001 decides native runners per architecture* |
| type `10.0.401` into the notes | `test_nothing_version_bearing_is_typed_into_the_workflow`: *types '10.0.401', which acadsharp/native/global.json already says* |
| change the tag prefix to `acad-` | `test_tag_matches_the_drivers_release_tag` |
| drop `--clobber` | `test_a_re_run_replaces_its_own_assets[build-linux]` |
| drop `if: always()` from `release-notes` | `test_the_notes_job_runs_whatever_happened` |
| `cancel-in-progress: true` | `test_runs_are_serialised_per_ref_without_cancelling` |
| drop `create-release` from the build jobs' `needs` | `test_the_release_exists_before_the_fan_out_uploads`: *parallel `gh release create` calls race* |
| delete one line from the notes job's `TARGETS` | `test_the_notes_job_knows_every_matrix_cell_and_whose_job_it_is` |
| README: a download URL with no digest | `test_every_download_url_carries_its_digest` and `test_readme_either_links_archives_or_says_there_are_none`: *it currently does both* |
| README: a digest with no URL | `test_every_digest_names_what_it_digests` |
| break the header's read-range regex | `test_the_workflow_uses_the_same_regex` and `test_the_preamble_generator_produces_the_pins_it_promises` |
| stop recording missing targets in the notes loop | `test_the_notes_generator_names_every_target_that_did_not_publish`: *linux/x64 published nothing and the notes never say so* |
| blank the per-archive digest | `test_the_notes_generator_reports_what_did_publish`: *the archive's own sha256 is not in the notes* |
| a syntax error in a notes generator | `test_every_inline_script_compiles` and the generator run |
| revert the Build step to `--cpu` | 3 red, including `test_the_build_step_command_actually_plans_the_right_archive`: *the driver refused acadsharp/build_acadsharp.py --platform linux --cpu x64* |
| set one cell to `arch: arm64, cpu: x64` | `test_every_cell_carries_both_vocabularies_and_they_agree`: *builds arm64 and then names the archive x64, so it would verify and upload a target it did not build* |
| delete `arch` from one cell | 3 red, including *has no arch, so the Build step has nothing to pass* |

Every one of those was restored and the suite re-run green afterwards.

- **The generators are run, not just read.** The notes are produced by Python that lives inside the workflow, so nothing compiles it until a release run does and a typo is found on the release page. `TestTheInlineGeneratorsRun` compiles every inline block and then executes both generators against this tree, with no network and no runner. Asserting on the workflow's text alone would never have proved the missing-target loop reaches every cell: deleting the `missing.append` leaves every text-level assertion green, and only the run catches it.
- **`test_nothing_names_a_microsoft_platform_target_or_toolchain` borrows `test_acadsharp_targets.py`'s regex** rather than restating the runtime identifiers. Spelling them out a second time trips that file's own directory-wide guard, which is how I found out it existed. That guard walks `acadsharp/` only, and this workflow is not in it.
- **`--prerelease` was dropped from the create step.** I had it there for the throwaway run, but pdfium and zstd do not use it, and a release marked pre-release is never "latest", which a consumer resolving the tag would trip over.
- **Five archives, not six.** The acceptance list says "the six URLs", copied from zstd's shape. Issue #48's own table has five: there is no Microsoft-platform artifact and the mac slice is Apple Silicon only. The README table lists five and says so.

## The README says there is no release, because there is not

#43 is the standing example in one direction: `zstd/README.md` advertised six URLs for a release that had never been cut and nothing checked. Publishing and leaving the README silent is the same failure with the sign flipped. So `acadsharp/README.md` carries the `No archives are published yet` marker and a table of the five archives with their Rust triples, and the test holds the two halves together both ways: a download URL with no sha256 on its line fails, and so does a sha256 on a line with no link. `test_readme_either_links_archives_or_says_there_are_none` makes "neither" impossible, so the pairing test cannot go quiet.

## The seam with #48, and the flag it caught

#48 has landed on `origin/feat/48-packaging`, and composing the two found a real break: this workflow was passing `--cpu` to the driver, which takes `--arch`. Every cell would have died on argparse the first time this ran. My side moved, and it was not a judgement call: `build_zstd.py` and `build_pdfium.py` both take `--platform` and `--arch`, and #48's driver follows them exactly. `--cpu` is the archive-name token, not a CLI flag.

I read the driver, then ran it in a composed worktree rather than trusting either reading:

```
resolve_jobs(None, None)      -> [('linux','amd64'), ('linux','arm64'), ('musl','amd64'), ('musl','arm64')]
archive_name('linux','amd64') -> acadsharp-linux-x64.tgz
release_tag(read_version())   -> acadsharp-3.7.1-viprs.1
CLI flags                     -> --arch --output-dir --parallel --plan --platform --target --upload
```

So there are two vocabularies for one thing, and each matrix cell now carries both: `arch: amd64` for the driver, `cpu: x64` for the archive name and `verify_archive.sh`'s third argument. They are spelled out rather than derived, because the driver's `ARCH_ALIASES` maps `x64` to `amd64` and would have accepted `--arch x64` quietly. `CPU_FOR_ARCH` in the test module is the single place the two are tied together, and a cell whose `arch` and `cpu` disagree now fails rather than building one target and uploading another under a name the build never produced.

Three tests gained real teeth from this:

- `test_the_build_step_command_actually_plans_the_right_archive` takes the Build step's **real command line** out of the workflow, fills in one matrix cell, and runs it with `--plan`. That goes through the driver's argparse and its own resolution, so a rejected flag, a differently-resolved cell and a differently-named archive all fail. All five cells plan exactly the archive their Verify and Upload steps go on to name.
- `test_the_workflows_cells_are_what_the_driver_resolves` hands `resolve_jobs` the exact flags each cell passes and requires that one cell back.
- `test_every_cell_carries_both_vocabularies_and_they_agree` holds `arch` against `cpu` inside the workflow, with no driver needed.

What made this findable at all is that the driver-contract tests were written to skip on a branch without the driver and arm themselves when composed. They did exactly that. Running this file alone, or this branch alone, would never have said a word: "green on my branch" and "green on my files" are not the same claim as green, and this is the case that proves it. The `--platform`/`--cpu` guess is now a checked fact, and the `bin/<name>.tgz` output path is confirmed too (`--output-dir` defaults to `./bin`).

## What the adversarial review changed

Three findings, all the same shape: an assertion about intent standing in for one about mechanism.

The `version` dispatch input was honoured by the tag and ignored by every build. Both build jobs set `env: VERSION` and never read it, so a dispatched override would have published archives built from the committed `VERSION` under a different tag, with `LINKINFO.json` and the release page disagreeing and nothing checking. That is the path `acadsharp/README.md` steers a human onto, since the first release has to be cut by dispatch. Both Build steps now pass `--version "$VERSION"`.

The guard is general rather than specific: `test_every_environment_variable_a_step_declares_is_read_by_it` fails any step that declares an `env` key and never names it in its script, with `GH_TOKEN` the one exception since `gh` reads it without being named.

`release-notes` now reads `artifact_version` out of every archive and refuses any that is not the version being published, or that records none at all. The notes go up first so the page explains itself, then the offending assets come off, then the run fails.

`--prerelease` is dropped, so the workflow and `build_acadsharp.upload_release()` create the same kind of release for the same tag. Pinned on both sides.

The skip guards in `TestTheDriverSpeaksTheSameContract` are gone. They were right while the driver was unlanded and would have become permanent silent passes once #54 merged, sitting on the strongest cross-lane assertions in the file. Zero skips remain in this file.

## Merge order

**This PR has to merge with or after #54.** Its driver-contract tests reach `build_acadsharp.py` by plain attribute access now, so they are red on this branch alone and green composed. Three of them stay red until #54 adds `--version` to the driver.

## What is not done, and the commands to finish it

The last acceptance item on #49 is a throwaway pre-release cut on real runners with the assets smoke-tested on a machine with no .NET. **I did not cut one**, and it cannot be cut yet: it needs #48's archives and verifier to exist, and it needs a push to a real `release` branch. Everything else on the issue is done and green.

Once #48 has landed and this is on `main`, the sequence is:

```bash
# 1. Put the workflow on the release branch. It only fires from there.
git fetch origin
git switch -c sync-release origin/release
git merge --no-ff origin/main
gh pr create --base release --head sync-release --title "Sync release with main" --body "..."
# merge it

# 2. Add a throwaway upstream pin, since the driver refuses a version it has no digest for.
#    3.7.1-viprs.0 shares 3.7.1's upstream half, so SOURCE_SHA256 already covers it.

# 3. Cut it on real runners.
gh workflow run release-acadsharp.yml --repo libviprs/libviprs-dep \
  --ref release -f version=3.7.1-viprs.0
gh run watch "$(gh run list --repo libviprs/libviprs-dep \
  --workflow=release-acadsharp.yml --limit 1 --json databaseId --jq '.[0].databaseId')"

# 4. All five assets must be on the release, and the notes must name any that are not.
gh release view acadsharp-3.7.1-viprs.0 --repo libviprs/libviprs-dep \
  --json assets,body --jq '{assets: [.assets[].name], body: .body}'

# 5. Smoke each one where no .NET exists. debian:bookworm-slim has no runtime and no SDK.
for t in linux-x64 linux-arm64 musl-x64 musl-arm64; do
  gh release download acadsharp-3.7.1-viprs.0 --repo libviprs/libviprs-dep \
    --pattern "acadsharp-$t.tgz" --dir /tmp/smoke
done
docker run --rm -v /tmp/smoke:/s -v "$PWD/acadsharp/tests/smoke:/smoke" \
  --platform linux/amd64 debian:bookworm-slim sh -c '
    apt-get update -qq && apt-get install -y -qq gcc >/dev/null
    ! command -v dotnet   # the point of the exercise: no .NET anywhere
    tar -xzf /s/acadsharp-linux-x64.tgz -C /opt
    gcc /smoke/smoke.c -I/opt/acadsharp-linux-x64/include \
      -L/opt/acadsharp-linux-x64/lib -lacadsharp_native -o /tmp/smoke
    LD_LIBRARY_PATH=/opt/acadsharp-linux-x64/lib /tmp/smoke'
# repeat on --platform linux/arm64, and on alpine:3.20 for the two musl archives.
# mac-arm64 runs the same way on a Mac with no SDK installed.

# 6. Delete the throwaway and its tag.
gh release delete acadsharp-3.7.1-viprs.0 --repo libviprs/libviprs-dep --cleanup-tag --yes
```

`MANUAL.md`'s new "Cutting an acadsharp release" section carries the short form of this.

## Gate

All in containers. `ruff check .` and `ruff format --check .` clean, `tools/shellcheck-all.sh` clean, `pytest -v` **435 passed, 30 skipped** on this branch alone, up from 355 passed / 24 skipped on the base. Composed with `main` + #46 + #48 the same suite is **702 passed, 91 skipped**, ruff and shellcheck clean. The extra skips on this branch are the driver-contract tests, which have no driver to ask until they are composed. No new shell scripts, so shellcheck's coverage is unchanged.

This is a new workflow file, not `ci.yml`, so the libviprs-tests Hook Mirror contract does not read it and no pairing PR is needed. `ci.yml` is untouched.

## Why this is still a draft

It is branched off `feat/46-abi-freeze` because #52 had not merged when it started, so the diff above currently carries #52's two files as well. It needs a rebase onto `main` once #52 lands, and then it is ready.


spdrman:
Composition check, since `feat/46-abi-freeze` moved after this branched off it.

I merged the current tip of that branch (`90d18d6`) onto a scratch branch with this work and ran the full gate. `acadsharp/tests/test_release_workflow.py` stays at 76 passed / 5 skipped in the composed tree, so nothing here interacts with the ABI lane.

The composed tree is red, but every failure belongs to #52's own in-progress TDD commit ("pin the contract before writing the thing that has to satisfy it"): `test_abi_docs.py`, `test_abi_exports.py`, `test_abi_fingerprint.py`, `test_abi_layout.py`, `test_wire_protocol.py` and `test_conformance_consumers.py`. None of them is mine and none of them changes behaviour under this change.

One thing there is worth passing on rather than sitting on, since it is not visible from inside #52 alone: `test_acadsharp_targets.py::TestNothingNamesAMicrosoftTarget` goes red in the composed tree on `tests/test_abi_docs.py:103` and `tests/test_conformance_consumers.py:229`. That is the directory-wide acceptance guard on the spike, it walks all of `acadsharp/`, and it fires on a runtime identifier or toolchain name appearing anywhere under there. Two of #52's new files trip it. It has nothing to do with this PR beyond my having found it while composing, and it is #52's to resolve.

The scratch branch was deleted; nothing here is built on it. This branch still needs its rebase onto `main` once #52 merges.

spdrman:
Fixed the seam with #48. My side moved, and it was not a judgement call: `build_zstd.py` and `build_pdfium.py` both take `--platform`/`--arch`, and #48's driver follows them. `--cpu` is the archive-name token, not a CLI flag.

I checked the contract by running the driver in a composed worktree rather than by reading it, and reproduced all three failures first. The fix carries both words on every cell: `arch: amd64` for the driver, `cpu: x64` for the archive name and the verifier's third argument. Spelled out rather than derived, because `ARCH_ALIASES` maps `x64` to `amd64` and would have accepted `--arch x64` quietly, which is exactly how a difference like this stays hidden.

Three new guards, all mutation-tested in the composed tree. The strongest takes the Build step's real command line out of the workflow, fills in a cell and runs it with `--plan`: all five cells now have to plan exactly the archive their Verify and Upload steps go on to name. Reverting to `--cpu` turns 3 red, including *the driver refused acadsharp/build_acadsharp.py --platform linux --cpu x64*. Setting a cell to `arch: arm64, cpu: x64` turns 2 red. Deleting `arch` from a cell turns 3 red.

Then re-ran the other 21 mutations against the composed tree, since the fix touches the steps several of them exercise. Two had stale anchors once the cells gained `arch`, and the apply guard I added caught that rather than reporting a pass for a mutation that never applied. Refreshed both and they go red as before. 24 for 24.

Composed with `main` + #46 + #48: **702 passed, 91 skipped**, ruff and shellcheck clean.

On your other point, you are right and I was wrong to file it the way I did. I said the `test_acadsharp_targets.py` breakage was composition-only. I had only run the composed tree, saw it red there, and did not check whether #52 alone was red too, so "composition-only" was a claim I had not tested in either direction. It was red on its own branch. Noted, and it is the same lesson as the flag: the driver-contract tests here are written to skip without the driver and arm on composition, and they are the reason this was findable at all, but they say nothing at all until someone composes. Running my own file, or my own branch, was never going to find it.

Still based on `feat/46-abi-freeze` per your note, rebasing onto `main` once #52 lands.

spdrman:
Review findings addressed. All three were the same shape, an assertion about intent standing in for one about mechanism, and I have tried to fix the shape rather than the instances.

**The `version` input now reaches the build.** Both Build steps pass `--version "$VERSION"`. The general guard is `test_every_environment_variable_a_step_declares_is_read_by_it`: a step that declares an `env` key and never names it in its script fails, with `GH_TOKEN` the one exception since `gh` reads it without being named. That is exactly how `VERSION` came to reach the tag and not the build, and it now catches the whole class rather than this one case.

`test_an_overridden_version_reaches_the_build` takes **two** runs, because one proves nothing. A driver ignoring `--version` plans happily for any override, so the control is an override that must be **refused**: `9.9.9-viprs.7` has no `SOURCE_SHA256` entry, so a driver reading the flag rejects it and a driver ignoring it exits 0 against the committed version. `3.7.1-viprs.0` must then be accepted, so the first result is not just a driver that refuses everything. Neither run may fail with argparse's `unrecognized arguments`, which names the value it rejected and so looks exactly like the flag being honoured. My first draft asserted only that the override appeared in the output, and it passed against a driver with no `--version` at all, for precisely that reason.

**`release-notes` refuses a mismatched archive.** It reads `artifact_version` from each archive and refuses any that is not the version being published, or that records none at all, because cannot-be-checked is not checked. The notes go up **first** so the release page says what happened, then the offending assets come off with `gh release delete-asset`, then the run fails. The ordering is asserted: a run that dies mid-deletion must still leave a page that explains itself.

**`--prerelease` is gone, and this time it is actually in the file.** Worth owning: the earlier commit that claimed to drop it did not touch the workflow. I made the edit, left it uncommitted, then ran my mutation harness, whose per-iteration restore is `git checkout --` on exactly that path. It discarded the edit and I wrote the commit message without re-reading the file. The harness now refuses to run against a dirty tree on the paths it restores, and I watched it refuse. I chose dropping over promoting because `build_acadsharp.upload_release()` creates a plain release for the same tag, so marking it in CI would make the kind of release depend on who cut it. Pinned twice: once on the workflow text, and once as `test_both_paths_to_this_tag_create_the_same_kind_of_release`, which reads `inspect.getsource(ba.upload_release)` and requires the two paths to agree, so the next person to add it to either side is told about the other.

**The eight skip guards are gone.** `TestTheDriverSpeaksTheSameContract` reaches the driver by plain attribute access, so a rename is an `AttributeError`. **Zero skips remain in this file.**

**Mutations: 29 now, all re-run against the composed tree.** The five new ones:

| mutation | new red |
| --- | --- |
| remove `--version "$VERSION"` (the finding, restored) | `test_the_build_step_names_the_resolved_version` and `test_every_environment_variable_a_step_declares_is_read_by_it`: *these steps declare an environment variable and never read it* |
| reinstate `--prerelease` | `test_the_release_is_not_marked_a_pre_release` and *upload_release() and release-acadsharp.yml disagree about whether this release is a pre-release (driver: False, workflow: True)* |
| skip the `artifact_version` check | both refusal tests |
| leave a refused asset on the release | *a refused archive stays on the release page unless something removes it* |
| delete before the notes go up | *the notes must go up before the assets come off* |

## State, and what is blocking

Composed with `main` + #46 + #48: **710 passed, 91 skipped, 3 failed**, ruff and shellcheck clean. All three reds are the `--version` coupling, and `--version` is not on `origin/feat/48-packaging` yet:

```
build_acadsharp.py: error: unrecognized arguments: --version 3.7.1-viprs.1
```

They go green when #54 lands it. On this branch alone 10 fail, which is the driver-contract class with its guards removed and is the honest state rather than a shrug: those tests assert a cross-lane contract and now say so out loud. **This PR has to merge with or after #54, not before.**

Base updated to the current `feat/46-abi-freeze` including the ABI.md fingerprint fix.

spdrman:
Rebased onto `main` after #52 landed as a squash. Force-pushed with lease: `098458a` to `a4b6830`, nine commits sitting linearly on top of `c5e6b5a`, and the PR now shows exactly my four files.

I took the cherry-pick route, but not the list in the brief. Checked it against the branch first, and it was wrong in two ways:

- it omitted `5eee6b8` and `f9331ad`, which are the commits that **create** `.github/workflows/release-acadsharp.yml` and `acadsharp/tests/test_release_workflow.py`. Following it literally would have had `a7d4223` editing a workflow that did not exist, and the PR would have lost the workflow entirely;
- it placed `876793d` after `6e56c45`, but it is third chronologically, so replaying in that order would have conflicted against itself.

The real set is nine, in this order: `5eee6b8`, `f9331ad`, `876793d`, `a7d4223`, `de38733`, `6e56c45`, `6565e1f`, `a58f475`, `098458a`. All nine applied with no conflicts. I dropped the `-x` provenance lines too, since they pointed at SHAs that stop being reachable the moment the branch is force-pushed.

Rather than eyeball it, two checks that together prove the rebase both ways:

1. `git diff --stat origin/main HEAD` touches **only** my four files, so none of #52's work was replayed, duplicated or dropped.
2. Each of those four files is byte-identical to the pre-rebase branch (`git diff --quiet backup HEAD -- <file>` for each), so none of my work was lost.

## The pair works

`--version` has landed on `origin/feat/48-packaging`, so I composed `main` + #54 + #53 and ran the whole suite: **757 passed, 109 skipped, 0 failed**, ruff and shellcheck clean. My own file is **94 passed, 0 skipped**, and the three tests that were red now pass by name rather than quietly:

```
test_the_build_step_command_actually_plans_the_right_archive PASSED
test_an_overridden_version_reaches_the_build                 PASSED
test_the_build_step_passes_flags_the_driver_accepts          PASSED
```

That is the `--version` seam closed end to end, from the dispatch input through `resolve-version` to the flag the driver actually parses.

On this branch alone 10 still fail, the driver-contract class with its skip guards removed. Unchanged and intended: those tests reach `build_acadsharp.py` by plain attribute access, and the driver is #54's. **This PR still has to merge with or after #54.**

The end-to-end throwaway release remains the one open acceptance item.

PR #54, opened 2026-09-13
G1.4: Build, package and verify the ACadSharp archives across the five targets

### LINKED ISSUE #48: G1.4: Build, package and verify the ACadSharp native archives across the five targets, and freeze LINKINFO.json

spdrman:
Closes #48

`build_acadsharp.py` ran one `dotnet publish` and stopped. This turns it into the
whole pipeline: five cells, the documented archive layout, `LINKINFO.json` and
`BUILDINFO.json` written from facts measured on each target, and a verifier that
has been watched refusing every broken archive the issue lists.

Rebased onto `main` now that #52 is in, so the diff is the packaging and nothing
else. Every archive ships that header byte-identical and records its sha256;
nothing here hardcodes the digest, so the two header moves during the review cost
a rebuild rather than an edit.

## The five cells

| archive | .NET RID | Rust triple | built |
| --- | --- | --- | --- |
| `acadsharp-linux-x64.tgz` | `linux-x64` | `x86_64-unknown-linux-gnu` | for real, in a container |
| `acadsharp-linux-arm64.tgz` | `linux-arm64` | `aarch64-unknown-linux-gnu` | for real, in a container |
| `acadsharp-musl-x64.tgz` | `linux-musl-x64` | `x86_64-unknown-linux-musl` | not on this host, see below |
| `acadsharp-musl-arm64.tgz` | `linux-musl-arm64` | `aarch64-unknown-linux-musl` | for real, in a container |
| `acadsharp-mac-arm64.tgz` | `osx-arm64` | `aarch64-apple-darwin` | **synthesised** |

NativeAOT does not support cross-OS compilation, so the mac cell cannot be built
or executed in a Linux container and this lane runs everything in containers by
constraint. What is exercised instead is a hand-built Mach-O: the matrix entry,
the archive naming, `finish_archive`, both manifests and the verifier's Mach-O
reader all run against it, and its refusals are watched the same way the Linux
ones are. **The real mac build happens on `macos-15`**, which is what
`release.yml` already uses, in lane D's release workflow.

Naming follows this repo rather than the design document:
`<dep>-<platform>-<cpu>.tgz`, extracting to a same-named directory, tag
`acadsharp-<VERSION>`. Both sibling verifiers infer platform and cpu from that
shape and `pdfium/tests/test_naming.py` and `zstd/tests/test_zstd_naming.py` pin
it, so a third scheme would break this repo's own tests. The version lives in the
tag, the Rust triple lives inside `LINKINFO.json`.

## The static link, which is where the review landed

The four-expert review found a Critical here and it was right. The first version
of this shipped `link_args: ["-Wl,-u,_GLOBAL__sub_I_main.cpp"]` and told the
consumer to emit one `cargo:rustc-link-arg` per entry. **That flag never
arrives.** Cargo sends `rustc-link-search` and `rustc-link-lib` to the link line
of everything that depends on the emitting crate, and keeps `rustc-link-arg` for
that package's own targets. So the requirement reaches `acadsharp-sys`'s own
tests and examples, stays green in its CI, and is missing from every binary that
depends on it.

It is silent twice over. Without the forced initialiser the static link still
**succeeds**, and the binary aborts on the first call into the library, which is
exit 134 and nothing else.

So the initialiser ships as its own archive and the requirement is a library
rather than an argument:

```
cargo:rustc-link-search=native=<archive>/lib
cargo:rustc-link-lib=static:-bundle,+whole-archive=acadsharp_native_init
cargo:rustc-link-lib=static:-bundle=acadsharp_native
cargo:rustc-link-lib=<each static_system_libraries entry>
```

in that order, and all three modifiers are load-bearing.

`+whole-archive` because `libacadsharp_native_init.a` holds one object, the
runtime's static initialiser, which lives in `.init_array` and **defines no
symbol anyone references**: as an ordinary archive the linker leaves it out. Init
first, because it is the one with the dangling references; after the main archive
the link fails on `RhRegisterOSModule`.

`-bundle` on **both**, which the new end-to-end check found on its first run
against a real archive while the C probe beside it linked and ran. With the
default `+bundle` rustc packs a static native library into the `-sys` crate's
rlib, and that rlib lands ahead of the whole-archived init archive; the linker
reads left to right, has no reason to pull the runtime object defining
`RhRegisterOSModule` while it is at the rlib, and cannot go back once the
bootstrapper asks. Measured across all three combinations: bundled fails,
init-only-unbundled fails, both-unbundled works.

The binary comes out the same size as the `-u` route, so none of this costs
anything.

`static_link_args` survives the rename from `link_args` but now refuses anything
that forces a symbol, whatever symbol it names, because that is the shape that
cannot travel. It is empty on every cell.

## The manifest, which is the consumer contract

The two linking modes have separate fields now. They measure different things,
and `system_libraries` used to be the static link's list when `static_certified`
was true and the shared library's `NEEDED` list otherwise, with nothing in the
file saying which.

| field | when | what |
| --- | --- | --- |
| `shared_system_libraries` | always | the shared library's `NEEDED`, libc and the loader dropped |
| `static_library` | certified only | the merged archive |
| `static_init_library` | certified only | one object, the runtime's static initialiser |
| `static_system_libraries` | certified only | what the static link needed |
| `static_link_args` | certified only | extra flags that link needed, empty everywhere |

The four `static_*` fields are present together and absent together. Absent, not
empty: an empty string is a path and an empty list is a measurement of none, and
neither of those means "not measured".

## The check that would have caught it

The C smoke proved the archives are linkable from a hand-written `cc` line. It
proved nothing about the path MANUAL.md tells a consumer to take, and the
difference between those two is exactly where the flag went missing. So the gap
that froze a broken recipe into a manifest was the check itself.

`acadsharp/tests/link_consumer/` is a two-crate workspace, because one crate
cannot catch it: a `-sys` crate gets every directive its own build script emits,
including the ones that do not propagate. The binary is a separate crate that
depends on it and nothing else. `acadsharp/scripts/link_consumer_smoke.sh` points
that workspace at an unpacked archive, links it through the documented
directives and **runs** the binary, comparing the fingerprint it gets back
against the manifest; `verify_archive.sh` calls it whenever the host has cargo
and can build for the archive's target, and it runs on `ubuntu-latest`, which
ships cargo.

The test that matters most puts the broken recipe back:
`cargo:rustc-link-arg=-Wl,-u,...` plus a plain `static=` link, and the consumer
binary takes SIGABRT. If cargo ever starts propagating link arguments, that test
goes green and says so, which is the only honest way to hold a rule that depends
on somebody else's behaviour.

## Three more from the same review

**The probe is run, not only linked.** An archive whose initialiser was never
forced links perfectly and aborts at the first call, so a link-only check cannot
tell a working archive from a broken one. The verifier now runs it, prints the
exit status and compares the fingerprint, and the probe allocates a megabyte
before it asks the library anything, because an export that returns a constant
can be answered by a binary whose runtime never came up. The fixture reproduces
the same shape, so `test_an_inert_initialiser_is_caught` fails on exit 134 rather
than on anything a static reader could have seen.

**The merged archive had duplicate members.** `ar addlib` keeps each source
archive's member names, and two runtime archives ship objects called the same
thing: 247 members for 244 distinct names on linux-arm64. It links today only
because the linker takes the first definition, and it forecloses
`--whole-archive` on the merged file. The merge extracts and prefixes by source
archive instead, so nothing is dropped and no two members share a name, and the
verifier refuses an archive that has any.

**`--version`.** `release-acadsharp.yml` declares a `workflow_dispatch` input
that overrides `acadsharp/VERSION`, tags from it and writes the notes from it,
and the driver had no way to accept one, so a dispatch override published
archives built from the committed version under a different tag with the manifest
disagreeing. The flag settles it in the direction G1.5's acceptance needs: its
throwaway pre-release on `3.7.1-viprs.0` is not reachable otherwise without
committing a bump and reverting it. `artifact_version` reflects the resolved
version, and MANUAL.md no longer claims the opposite is deliberate.

## What each cell measured

Everything below came out of a container build on this machine, arm64 natively
and amd64 under Rosetta, against the header as merged in #52
(`765d22c58cf48b52`). Nothing here is hardcoded: the two header moves during the
review cost a rebuild, not an edit.

| cell | shared smoke | static | `shared_system_libraries` | `static_system_libraries` | archive |
| --- | --- | --- | --- | --- | --- |
| linux/x64 | pass | certified, 266 objects | `["m"]` | `["m"]` | 15.9 MB |
| linux/arm64 | pass | certified, 247 objects | `["m"]` | `["m"]` | 15.3 MB |
| musl/x64 | not built here | not built here | | | |
| musl/arm64 | pass | certified, 247 objects | `[]` | `["m"]` | 15.2 MB |
| mac/arm64 | synthesised | not attempted | measured on `macos-15` | | |

`static_link_args` is `[]` on every cell, which is the point of the initialiser
being a library.

Each archive was then verified again on a host that matches its target, because
the driver's own verification runs on the macOS host where neither probe can:

- linux/arm64, in a Debian arm64 container with cargo: the static probe links and
  **runs**, and **the documented cargo recipe links and runs against the real
  archive**.
- linux/x64, in the x64 builder image: the static probe links and runs. The
  merged archive carries `do_vxsort_avx512.cpp.o` and `do_vxsort_avx2.cpp.o`,
  the x64-only pair ADR 0001 warns is easy to miss, which is the 266 against 247.
- musl/arm64, in the Alpine arm64 builder image: the static probe links and runs.
  No cargo there, and the verifier says so rather than passing quietly.

`BUILDINFO.json` per platform: bookworm-slim with Debian clang 14.0.6 and GNU ld
2.40, alpine:3.20 with Alpine clang 17.0.6 and GNU ld 2.42, .NET SDK 10.0.401
everywhere, 16 AOT warnings everywhere, which is ADR 0001's count and all of it
upstream code.

## musl/x64 does not build on this host, and ADR 0001 predicted it

Three cells published, smoked and certified here. `musl/x64` did not, across four
attempts: ILC crashes inside its own multithreaded scanner, with
`IndexOutOfRangeException` in `RangeWorker.FindNewWork` once and
`NullReferenceException` from the same
`ILCompiler.ILScanner.CompileMultiThreaded` the other three times. Different
non-deterministic failures out of one `Parallel.For` is a memory-model symptom
rather than a source one, and this host is an Apple Silicon Mac running the
x86_64 container under Rosetta. `linux/x64`, also under Rosetta, gets through.

Publishing that cell with `-p:IlcSingleThreaded=true`, which is
`--parallelism:1`, succeeds in 117 seconds and the archive then smokes and
verifies like the others, so the target builds and the emulation does not.

ADR 0001's fifth decision is "Native runners per architecture. No emulation", and
this is that decision being right rather than cautious. Nothing is worked around
in the driver: capping ILC to one thread would fix one laptop and slow every CI
runner, and CI runs x64 natively. The cell is in the matrix, its naming, manifest
and verifier paths are covered by the tests, and `ubuntu-latest` builds it.

## What I left out, and why

**No job in `.github/workflows/ci.yml`.** A per-target smoke job there would
engage the `libviprs-tests` Hook Mirror contract, which reads this repo's
`ci.yml` at `main` and mirrors every job into the shared pre-commit hook, so it
would red that repo until a pairing PR merged behind it, and a .NET SDK is not
something a pre-commit hook should require. It belongs in lane D's separate
release workflow file, which that guard does not read.

**`--parallel` builds four, not five.** The acceptance bullet as written cannot
be satisfied by one host: the mac cell needs a macOS host and the Linux cells
need Docker. The default matrix is the four container-buildable cells, exactly as
`build_pdfium.py` and `build_zstd.py` exclude mac from theirs, and `MATRIX` still
holds all five so the matrix test pins the whole thing.

**The driver never emulates.** ADR 0001 measured that .NET does not support QEMU
and that the SDK ships no cross toolchain, so a foreign-architecture cell on a
Linux host needs a runner of that architecture. On an Apple Silicon host the
amd64 cells run under Rosetta, which is not QEMU; that is how the x64 cells here
were built.

## A note on `test_acadsharp_targets.py`

It was expected to break, and it does not. `TARGETS` grew from a tuple of runtime
identifiers into a dict keyed by the same identifiers, so `set(ba.TARGETS)` and
`for rid in ba.TARGETS` still see exactly what they saw, and its assertions hold
unchanged rather than being rewritten around the new matrix.


spdrman:
Five things I found while building this and folded in here rather than filing,
per the lane's rule. Two of them change what the issue and the ADR say.

**1. `link_args` for a merged static archive.** Covered in the body: the symbol
the NativeAOT sample says to force is gone in .NET 10, and the replacement is not
"nothing". `libbootstrapperdll.o` defines no global symbol, so an `ar` archive
can never pull it in, and the runtime never initialises. The spike did not see
this because it linked the pack's objects directly, where `.init_array` runs
regardless. The archive is what a consumer with no .NET actually gets, so the
merge globalises that initialiser and the link forces it.

**2. The NativeAOT runtime pack has no `sdk`/`framework` split.** ADR 0001
describes one. In 10.0.12 every runtime archive is in
`microsoft.netcore.app.runtime.nativeaot.<rid>/<ver>/runtimes/<rid>/native/`, and
`runtime.<rid>.microsoft.dotnet.ilcompiler` has no `sdk` directory at all. The
first static attempt went looking for `*ilcompiler*/sdk`, found nothing, and
downgraded to shared-only with a message that read like a missing package. The
merge now finds that directory by `libbootstrapperdll.o`, so a layout change is a
missing-file error rather than a silent downgrade.

**3. musl is built and certified, which ADR 0001 left open.** The ADR says musl
"is not in this issue's target list and I did not build it", and guesses the
shape of the answer is an Alpine container on the matching architecture. That is
what this does, and both musl cells publish, smoke and certify their static
archive. Alpine ships clang 17.0.6 and binutils 2.42 against bookworm's 14.0.6
and 2.40, and `BUILDINFO.json` records which produced which binary.

**4. ILC's own scanner is what fails under emulation, not the musl target.**
`musl/x64` crashed twice inside `ILCompiler.ILScanner.CompileMultiThreaded`, once
with `IndexOutOfRangeException` in `RangeWorker.FindNewWork` and once with a
`NullReferenceException`, which is two different non-deterministic failures from
one `Parallel.For`. Publishing the same cell with `-p:IlcSingleThreaded=true`
(that is `--parallelism:1`) succeeds in 117 seconds and the archive then smokes,
certifies and verifies like the other three. So the target is fine and the
emulation is not, which is ADR 0001's "native runners per architecture, no
emulation" being right rather than cautious. The driver does not set that flag:
it would fix one laptop and slow every CI runner, and CI runs x64 natively.

**5. A bug in my own verifier that every archive would have hit.** `slice` reads
a byte range with `tail -c +N | head -c M`, `head` closes the pipe as soon as it
has its bytes, `tail` dies of SIGPIPE, and `pipefail` reports 141 for a read that
worked. Every archive failed at the first symbol table with no FAIL line to say
why, which is the worst kind of red: a refusal that names no defect. Fixed by
turning pipefail off for that one pipeline in a subshell. The same shape was
about to misread a musl host as glibc in the static link probe.

One thing I did not do: the Rust half of the conformance smoke. The C smoke is
generated from the header and resolves every entry point against the unpacked
archive, shared and static, and checks the live fingerprint. The Rust consumer
lives in the ABI and adapter lanes and the FFI itself belongs to `acadsharp-rs`,
so a second copy here would be a third place for the same contract to drift.

spdrman:
`--version` is pushed and on the head of this branch now, so lane D is unblocked: `python3 acadsharp/build_acadsharp.py --version 3.7.1-viprs.0 --plan` resolves to shim revision 0, and a malformed or unpinned override exits non-zero with the reason rather than building something it cannot name. The rest of the review response is still landing on top of it.

spdrman:
The four-expert review was right about the Critical and it cost the recipe twice,
not once. Both are fixed, and the second one is only visible because of the check
the review asked for, so that is worth spelling out.

**What the review found.** `cargo:rustc-link-arg` does not reach a dependent's
link line, so `link_args: ["-Wl,-u,_GLOBAL__sub_I_main.cpp"]` was a requirement
that arrived nowhere. The initialiser now ships as
`lib/libacadsharp_native_init.a`, `static_init_library` joins the frozen fields,
and the recipe is a library rather than an argument. `static_link_args` now
refuses anything that forces a symbol, whatever symbol it names.

**What the check then found.** The first run of the new end-to-end smoke against
a real archive failed on `undefined reference to RhRegisterOSModule`, while the C
probe beside it linked and ran. With the default `+bundle`, rustc packs a static
native library into the `-sys` crate's rlib, and that rlib lands on the link line
ahead of the whole-archived init archive: the linker reads left to right, has no
reason to pull the runtime object defining `RhRegisterOSModule` while it is at
the rlib, and cannot go back once the bootstrapper asks. So the recipe is
`static:-bundle,+whole-archive=` and `static:-bundle=`, both unbundled, init
first. Measured across all three combinations: bundled fails,
init-only-unbundled fails, both-unbundled works.

That is the whole argument for the check existing. `--whole-archive` first was
the right shape and still did not work through cargo, and nothing that reads
bytes or compiles a `cc` line could have said so.

**And a third, from running rather than linking.** Moving off `ar addlib` to fix
the duplicate members changed the merged archive's member order, because the new
merge fed `ar` whatever order `find` handed it. The archive linked cleanly and
the binary segfaulted on the way out, reproducibly: the linker emits
`.init_array` in the order it pulls members, so the archive's own order decides
what runs when. Members now go in each source archive's own order, managed
archive first, which is what ILC linked and what addlib produced. A link-only
probe would have shipped that.

**Four things in the tests were measuring nothing, and two of them were mine.**

- The regression test that puts the broken recipe back guarded on
  `"rustc-link-arg" in source` after a `str.replace`. build.rs says those words
  in the comment explaining why it never emits one, so a replace that matched
  nothing sailed through and the test exercised the *working* recipe. The guard
  is now that the substitution changed the file.
- The fixture's initialiser referenced only symbols the entry points already drag
  in, so every link ordering worked and the fixture could not tell the three
  modifier combinations apart. Its init object now calls `viprs_test_register`,
  which lives alone in the main archive with no other caller, which is the shape
  the real bootstrapper has.
- musl's `ldd` exits 1 for `--version`, and the verifier piped it straight into a
  condition, so under `pipefail` no host was ever musl. A musl archive was never
  link-tested on the only kind of machine that could do it, and the script
  reported that as "this host cannot build for musl/arm64", which reads like a
  deliberate skip. Fixed, and the real musl/arm64 archive now links and runs on
  Alpine.
- The `ar` symbol index was read by splitting the whole member on NUL, which
  glues the offset table's trailing bytes onto the first name. A one-symbol
  archive came back as `d_GLOBAL__sub_I_fixture`. The index header is parsed
  properly now, GNU and BSD layouts both.

**Rebased onto `main` now that #52 is in**, and the archives rebuilt against the
header as merged, `765d22c58cf48b52`. Nothing in this lane hardcodes the digest,
so the two header moves during the review cost a rebuild rather than an edit,
which is the property the lane was built for.

**On `--version`.** Added, defaulting to `read_version()`, and
`artifact_version` follows the resolved value. MANUAL.md no longer claims the
absence was deliberate. It refuses a malformed or unpinned override rather than
building something it cannot name: `--version 3.7.1` and
`--version 9.9.9-viprs.1` both exit non-zero with the reason.


One self-inflicted note for the record, since it cost a cell. I rewrote
`verify_archive.sh` while a build was running, and the driver invoked it mid-write:
`line 711: ---...---: command not found`, exit 127, reported as a rejected archive.
The archive was fine and re-verifies clean. Do not edit a script a running job is
about to execute.

PR #55, opened 2026-09-13
G1.3: Flatten DWG into the ABI's record stream, with the limits enforced and no curve tessellated

### LINKED ISSUE #47: G1.3: Implement the ACadSharp adapter and primitive flattening, and enforce the limits the ABI declares

spdrman:
Closes #47

The ACadSharp adapter: the flattening, the corpus that exercises it, the limits the ABI declares, and the measurements the issue asks for rather than the prose.

## The review round: two Criticals, both fixed and both mutation-tested

The safety reviewer found two ways a file takes the consumer's process down. They share one root cause, which is worth more than either fix: **every bound this decoder had counted output.** `max_entities` counted records emitted, `max_output_bytes` counted bytes, `max_block_depth` counted one kind of nesting. Nothing counted what the walk *did*, and a document can make a walk do an unbounded amount of work while producing none of those three. Both Criticals live in that gap and the fix for both is the same shape, counting visits rather than results. The next limit anyone adds to this ABI should be checked against that question before it is written down, and `tests/benchmarks/DECISION.md` now says so.

**C1: a DIMENSION's block can hold another DIMENSION.** That was walked by calling `Map` again, so the nesting a file chose became recursion on the CLR stack, and nothing consulted `max_block_depth` on the way. `MapHatch` had the same shape for its edge entities and its pattern lines. A `StackOverflowException` cannot be caught, so the export's catch-all was never in the picture: the runtime calls `FailFast`, which is a SIGABRT in the consumer's address space from an untrusted file, and the header's own comment says that is the one thing `max_block_depth` exists to prevent.

`Walk` already owned an explicit stack. Dimensions and hatches go on it now. A `Pending` item carries either an entity or a record that is already made, so a composite hands back its finished records and its further entities in one sequence, in order, with none of it going back through `Map`. `max_block_depth` therefore applies to a dimension's picture and a hatch's boundary as well as to an insertion, which is what closes the way round the bound.

**C2: `max_entities` counted records, and an INSERT emits none.** A chain of block records each holding two insertions of the next expands exponentially, emits nothing, and keeps its nesting inside `max_block_depth` the whole way. The counter the walk already kept was compared against nothing and read by nobody. It is compared now, which is what the header's own wording ("counted across the whole decode, block expansion included") describes.

And the caller could not get out of it. The cancel flag was read once at the entry to `decode_next_batch`, and on a document that yields no record that call does not return, so a caller who sets the flag is waiting on a decode that will never look. Checking in `BatchWriter.Peek()` does not help either, because the block is *inside* one `MoveNext()`. The walk polls the flag itself, every 1024 items.

**M2: `max_polyline_points` did not apply to splines**, and it was applied by the encoder after the list already existed. It moves into the flattener, where the count is known before anything is allocated, and it covers a spline's control points and knots as well as a polyline's and a polygon's vertices.

**H4: the malformed corpus was four inputs that all died at the same line.** Every derivative was refused inside `DwgReader` with the same `EndOfStreamException`, so the flattener, the encoder and the batch writer had never seen a byte of hostile input, which is exactly where both Criticals were. There are hostile-but-well-formed fixtures now, and a test that fails if the corpus collapses back to one failure site.

### Every mutation I ran for this round

| Mutation | What went red |
| --- | --- |
| Take the comparison back out of `CountEntity`, leaving the counter uncompared | The fan-out walked 32,767 entities against a bound of 100 and returned OK. At depth 24 the scenario recorder itself stopped terminating: the run that takes 2.4 s with the bound no longer finished in 560 s. |
| Put the dimension walk back inside `Map`, recursing at the same depth | `g13_dimension_deep.dwg` printed a stack overflow with `Flattener+<Map>d__27.MoveNext()` repeating through `BatchWriter.Peek()`, and the process never came back. With the fix it returns `LIMIT_EXCEEDED` naming `DIMENSION ... at expansion depth 65 and max_block_depth is 64`, exit 0, zero live handles. |
| Remove the spline's point count | The 20,000-point spline passed a `max_polyline_points` of 4,096 silently. With the guard it refuses and names the record type and the number. |

The earlier round's mutations (change a fixture, tessellate a curve, skip a close, drop the notifications, grow the recorded output) were rerun and still go red.

### Assertions, not outcomes

A decode that refused for the wrong reason looks identical to one that refused for the right one, so each new test names the bound in the message and sits beside a control on the same shape that decodes: three thousand nested dimensions against four, a fan-out at depth 24 against one at 8, a 20,000-point spline against the same file with the bound raised. For C1 the assertion is that the process came back at all, which is why the fixture is three thousand levels rather than a hundred: the recursive walk died around a thousand, so a shallower file would have passed against the broken build.

### Under NativeAOT, through the C ABI, with no .NET present

```
g13_dimension_deep       open 0, view_count 2, decode_begin 0, decode_next_batch -> 9   (LIMIT_EXCEEDED, not an abort)
g13_many_inserts         batches 38, bytes 2481264, done 1, live handles after close: 0
real_AC1032              batches 1, bytes 35964, done 1, live handles after close: 0
```

2,481,264 bytes, identical to the JIT harness, so trimming changes nothing about what the flattener produces and the bound holds on a different stack.

### Three shapes ACadSharp cannot build, recorded rather than left to be rediscovered

A **cycle**, of dimensions or of block records: a `CadObject` gets exactly one owner and `Add` refuses a second, and `new Insert(record)` deep-clones a record that belongs to a document, so the second edge recurses until the stack goes. A file from another writer can carry one, and the walk refuses it the same way it refuses the deep chain, by depth and by entity count, neither of which cares whether the graph closes. The **fan-out as a file**: `DwgWriter` does not come back from writing one, so the generator builds it in memory and the harness drives the flattener over it directly, which is what the hole was about anyway. A **string past the 64 KiB default**: ACadSharp's MText round trip silently shortens a long value (8,192 characters survive, 40,000 come back as 7,232, 200,000 as 3,392), so the bound is exercised with a limit the caller sets. All three are in `PROVENANCE.md`.

### What moved outside my lane, and why

`IDocumentSource.EnumerateView` takes a `Func<bool> canceled`. That is one line in the interface, one in `SyntheticSource` (which ignores it and says why), and one in `DecodeSession` to hand its flag over. It is the only seam that works: the flag has to reach the loop that can block, and nothing above that loop gets control while it does. `Volatile.Read` on the flag itself is already in `main` from the ABI lane's round.

`docs/ABI.md`'s limits table now says what `max_entities`, `max_polyline_points` and `max_block_depth` actually bound, with a paragraph on the two wordings whose obvious reading leaves a hole.

The only change to a committed expectation is bit 0 of the geometry prologue: records lifted out of a dimension's block or a hatch's boundary now carry it. They do come out of an expansion, and they did not say so before, because the old walk gave a composite's contents the same depth as the drawing they sit in. No record changed order or value.

## The rule this lane exists to hold

Arc, circle, ellipse and spline cross the boundary carrying the parameters that define them. `libviprs` owns zoom and precision aware tessellation and is the only layer that knows either, so anything decided here would be wrong at every zoom but one. The mutation is run below, not argued: tessellate the arc arm and four tests go red.

Two places where the wire made me think about it properly:

**A polyline bulge.** Wire version 1's `Polyline` is a point count and a run of triples, and there is nowhere to put a bulge. Dropping it turns an arc into a chord, which is worse than tessellation because a chord is not even a good approximation. Adding a field is a wire version bump and belongs to #46. So a bulged polyline goes out as what it is: the straight runs as `Polyline` records and every bulged span as an `Arc` with its own centre, radius and angles, every record keeping the source entity's handle so a consumer can regroup them. A polyline with no bulge is still exactly one record. The arithmetic is closed form from the two endpoints and the one number, and there is a test that checks the sweep against `4·atan(bulge)`.

**A hatch boundary.** A loop of straight edges is one `Polygon`. A loop carrying an arc, an ellipse or a spline goes out as its own edge records with a `HATCH_LOOP_NOT_POLYGON` warning, because a polygon of points is a straightened curve by another name.

## Flattening, per entity kind

| Entity | Records | Fixture |
| --- | --- | --- |
| `LINE` | `Line` | `g13_line.dwg` |
| `LWPOLYLINE`, `POLYLINE` 2D and 3D | `Polyline`, plus an `Arc` per bulged span | `g13_polyline.dwg` |
| `ARC` | `Arc` | `g13_arc.dwg` |
| `CIRCLE` | `Circle` | `g13_circle.dwg` |
| `ELLIPSE` | `Ellipse` | `g13_ellipse.dwg` |
| `SPLINE` | `Spline` (degree, knots, control points, weights) | `g13_spline.dwg` |
| `TEXT`, `MTEXT`, block attributes | `Text` | `g13_text.dwg` |
| `INSERT` | the block's primitives, transformed, depth bounded, flagged as coming from a block | `g13_insert.dwg` |
| `DIMENSION` | `Line` and `Text` from its block, or a warning when it has none | `g13_dimension.dwg` |
| `HATCH` | `Polygon` per straight loop, edge records otherwise, warning when there is no boundary | `g13_hatch.dwg` |
| anything else | `Warning(UNSUPPORTED_ENTITY)` naming the type and carrying the handle | `g13_unsupported.dwg` |

`real_AC1032.dwg`, the drawing AutoCAD produced, is the one fixture that reaches every record type at once: 361 records over all nine types, including 31 reader notifications. There is a test that fails if any type stops appearing in it.

## Block expansion never recurses on the CLR stack

`max_block_depth` is a number the caller sets, a caller may set it to four billion, and a bounded refusal that arrives as a stack overflow is not a bounded refusal. The walk is an explicit stack of enumerators.

## Every mutation I ran, and what it printed

**Change a fixture, leave the expectation.** Replaced `g13_arc.dwg` with the circle fixture's bytes. `test_the_digest_matches[g13_arc.dwg]` went red with both digests, and `fixturegen decode --check` named the record:

```
record 4 differs
  expected: 00004 Arc handle=49 flags=0 c=[(3.250000,4.500000,0.000000)] r=7.125000 a0=0.250000 ...
  actual:   00004 Circle handle=49 flags=0 c=[(-5.500000,6.250000,0.000000)] r=9.875000 ...
```

**Tessellate a curve in the shim.** Made the `Arc` arm emit a 33-point polyline, rebuilt, re-recorded. `g13_arc.dwg` became `{'Warning': 4, 'Polyline': 1}` and four tests went red: `test_exactly_one_record_of_that_type[g13_arc.dwg-Arc]`, `test_no_polyline_anywhere_in_it[g13_arc.dwg-Arc]`, `test_the_arc_carries_a_centre_a_radius_and_two_angles`, and `test_a_loop_with_an_arc_keeps_the_arc`. Restored and re-recorded; green.

**Raise `max_block_depth` past the fixture's depth.** This one is committed rather than run and thrown away, because it is the control that proves the bound is the limit and not a crash. `g13_deep_blocks.dwg` nests six deep: at five it is `LIMIT_EXCEEDED` naming `max_block_depth` and `nesting depth 6`; at seven the same file decodes cleanly. Both scenarios assert the same fixture digest.

**Skip a close in one error path.** Took the `Handles.Remove` out of `DocumentHandle.Dispose`. `viprs_acad__test_live_handles()` went from 0 to 1 and **35 tests** went red across the stream, limits and malformed suites. Restored; green.

**Drop the reader notifications.** `real_AC1032.dwg` went from 361 records to 330 and four warning tests went red, including `test_none_was_dropped`, which compares what reached the stream against what the reader raised. Restored; green.

**Grow the recorded output past the amplification ceiling.** This one found a real hole first. The ceiling was `recorded_output_bytes * 1.10`, which moves with the number it guards, so an implementation that doubled its output would rerecord a doubled ceiling and pass. I only noticed because the mutation did nothing. The ceilings are written out now, and with them pinned, setting `output_bytes` to 3,000,000 reds both `test_the_output_has_not_grown` and `test_it_quotes_the_output_bytes` (the check that `DECISION.md` still carries the figure it was decided on).

## The amplification numbers, and the block-definition decision

`tests/benchmarks/amplification.json` and `tests/benchmarks/DECISION.md`. `g13_many_inserts.dwg` is one three-entity block inserted ten thousand times.

| | |
| --- | --- |
| Input | 179,150 bytes |
| Records out | 30,004 |
| Bytes out | 2,481,264 (13.85x the input) |
| Batches at 64 KiB | 38 |
| Peak RSS | 94,036 KB |
| Managed bytes retained across the whole decode | 1 KB |

**Decision: keep expansion for v1, do not add block-definition records before release.** The last row is why. Thirty thousand records cross and the decoder ends holding one kilobyte more than it started with, because the stream is an iterator the batch writer pulls from and is never materialised. Expansion costs output bytes, not memory. A definition-plus-instance scheme would emit 3 records once and 10,000 instance records of 124 bytes each, about 1.24 MB against 2.48 MB: half, on this fixture, which is not what a new wire version and a second code path in every consumer should buy. The saving is a function of block size, not instance count, and this fixture is hostile on the cheap axis. The trigger to revisit is written down: a real corpus where a view's expanded output exceeds the default `max_output_bytes`, or where amplification passes 50x input, which means blocks of roughly twenty entities being instanced heavily.

## RSS: path versus memory, and streaming

**Streaming.** Retained memory is 1 KB at 132, 516, 2,052 and 30,004 records. Peak RSS during a decode moves by a few hundred kilobytes to a few megabytes and does not track the record count: 2,052 records grew RSS by 2,700 KB and 30,004 records grew it by 344 KB. That ordering is the test. The issue's literal budget (peak growth under four batch sizes) is **not** met and I do not think it can be on a managed runtime, because a collector's allocation churn is in that number and the flattener's own garbage is real even though none of it is retained. I have said so rather than tuned the GC until the number looked right; the retention figure is the one that answers the question the budget was asking, and it is two orders of magnitude inside it.

**Path versus memory.** On the largest committed fixture the peak RSS difference across five pairs was −84, −12, 76, 96 and 392 KB against a 175 KB file. That is noise, and the capture records it as noise rather than picking the run that agreed with me. What carries the claim is the same pair on the same drawing with 32 MiB appended, which the decode does not read (both runs produce exactly the 2,481,264 record bytes the unpadded fixture produces, asserted, not assumed):

| | |
| --- | --- |
| Input | 32,942 KB |
| Peak RSS difference | 32,688 KB (minimum of three runs: 32,524 KB) |
| Allocation difference across the open | 33,723,536 bytes against a 33,733,582 byte file |

The ~250 KB shortfall against the full file size is the path route's stream buffering, which the memory route does not need because it wraps the caller's array. `caller_copy_bytes` is 0 for every path open and the file size for every memory one, which is the same claim without any collector in it.

## Everything else the issue asks for

Every limit has a scenario that hits it **and a control on the same fixture that passes**, because a bound that has only ever been seen refusing might be refusing everything. `max_input_bytes` gets a third: a mode 000 copy made inside the container, where Linux permissions are real and the process is not root, so refusing on the size is a refusal that never opened the file and the control run shows that opening it fails with `UnauthorizedAccessException`.

The version gate has a control too. `g13_ac1009.dwg` is six bytes of signature and then zeros and is `UNSUPPORTED_FORMAT`; the identical file with an `AC1032` signature is `CORRUPT_INPUT`. If the gate ran after the parse, both would be indistinguishable.

Malformed inputs are derived at test time (truncation at 25, 50 and 90 percent, and 64 bytes flipped under seed 4713), never committed, and the capture records the digests so the refusal that was recorded is a refusal of the bytes the test is holding. All four are `CORRUPT_INPUT`, every process exits 0, and the live-handle count is 0 after each. The seed does not put a flip in the six-byte signature and the capture says so, so a future seed change that starts hitting it is visible rather than confusing.

XREF: the fixture decodes to a single `UNRESOLVED_BLOCK` warning naming the path, and the scenario runs it again with `--network none` and the repository mounted read-only. Both runs produce the same bytes, which is what rules out a filesystem reach that happened to succeed.

## It works under NativeAOT

Not asserted, run. I published the composed tree for `linux-arm64`, then `dlopen`ed the result in an `ubuntu:24.04` container with no .NET in it (checked: no `dotnet` on `PATH`, no `/usr/share/dotnet`) and drove it through the C ABI from a C program:

```
open_path_utf8 -> 0
view_count -> 2
decode_begin -> 0
first batch magic: VACB wire_version 1
batches 38, bytes 2481264, done 1
live handles after close: 0
--- the version gate, through the same library ---
open_path_utf8 -> 2
```

2,481,264 bytes, identical to what the JIT harness recorded, so trimming has not changed what the flattener produces. The shim now exports all eleven header entry points, which is what #54's archive verifier was refusing every archive for.

## Things I folded in rather than filing

**`SourceFactory.OpenPath` read the head of the file before applying `max_input_bytes`.** The header says that bound is applied "before the input is read", and the length is a stat, so deciding it needs no idea which source would have handled the bytes. Reading first made the refusal cost an open and turned an unreadable file into an IO failure where the answer was "too big". The check moved to the top. The mode 000 scenario is the test.

**An unreadable input came back `INTERNAL_ERROR`**, which tells a caller the library has a bug when the truth is they handed it something it cannot read. `ReadHead` failures are `CORRUPT_INPUT` now.

**Lane A's export allowlist accounted for one test-only export and two issues had specified one each.** #46 needs an export that throws, to prove the catch-all is real; #47 needs one that reports how many handles are alive, to prove a failure path closed what it opened. `TEST_EXPORT` became `TEST_EXPORTS`, and `viprs_acad__test_live_handles` joined `INFALLIBLE` rather than being exempted from the result-code contract, because it has no pointer, no length and nothing to fail at. The assertion that the throwing export actually throws stays specific to that one.

**`supports_block_expansion` moved from 0 to 1**, because this build can now do it.

**The scenario capture is not in `tests/fixtures/captures/`.** That directory is G1.1's JIT-versus-AOT parity set and `test_acadsharp_recorded_parity.py` asserts every JSON in it carries a system-variable count. A scenario capture is a different artefact and sits with the expectations it belongs to.

**Every file ACadSharp's own `DwgWriter` produces raises four reader notifications** (a `TableStyle.CellStyle` referencing a `TextStyle` by an empty handle and an empty name). The issue assumed the writer would emit none. It is an upstream round-trip gap, it is in every expectation in this PR, and `PROVENANCE.md` records it so the next person reading a dump does not have to work it out.

## What I left out, and why

**The streaming peak-RSS budget**, as described above. Measured, reported, and not met; the retention number is there instead and it is the sharper claim.

**A large committed fixture.** ACadSharp's `DwgWriter` costs minutes to produce a multi-megabyte drawing (I gave it three goes at 20k to 600k entities and killed each one past ten minutes), so the path-versus-memory measurement uses a padded input built at measurement time and thrown away. `PROVENANCE.md` says so.

**`DIMENSION_WITHOUT_BLOCK` has no fixture.** `DwgWriter` generates a block for every dimension it writes, so the corpus generator cannot produce one without. The arm stays because a real drawing can carry it, the test says out loud that this is the one code with no fixture behind it, and the check that every *other* code has one is what keeps that honest.

**No `ci.yml` job.** Running the corpus under the AOT binary would need a .NET SDK in the shared pre-commit hook through libviprs-tests' Hook Mirror, and ADR 0001 already decided that is the packaging issue's call. The recorded artefacts are what covers this on every pytest run instead.

## Gate

Composed `main` + #47 + #49 on a scratch branch: **1135 passed, 115 skipped, no failures.** On this branch alone: 1135 passed, 114 skipped. `ruff check` and `ruff format --check` clean, shellcheck clean over every tracked script. Everything ran in containers, one at a time.


spdrman:
Files I changed that are not this lane's, so the reviewers of those lanes see them in one place rather than in a diff.

**`acadsharp/tests/test_abi_exports.py`** (#46's). `TEST_EXPORT` was a single string and two issues had each specified a test-only export without knowing about the other's: #46 needs one that throws, to prove the catch-all is real, and #47 needs one that reports live handles, to prove a failure path closed what it opened. It is `TEST_EXPORTS` now, and `viprs_acad__test_live_handles` went into `INFALLIBLE` rather than being exempted case by case, because it takes no pointer and no length and has nothing to fail at. `test_it_actually_throws` stays specific to the throwing one: a live-handle counter would satisfy the letter of that assertion and prove nothing. This is a contract gap between two issues, not a guard relaxed to get green.

**`acadsharp/native/Exports.cs`** (#46's). Added `viprs_acad__test_live_handles`, inside the existing `#if VIPRS_ACAD_TEST_EXPORTS` block. Issue #47 asks for it by name and the malformed corpus reads it after every refusal.

**`acadsharp/native/Abi.cs`** (#46's). Two changes. `Handles.LiveCount`, which is what the export above reads. And `SupportsBlockExpansion` from 0 to 1, which the comment beside it said to do when the adapter landed.

**`acadsharp/native/Sources/SourceFactory.cs`** (#46's). The one branch it was left room for, plus two fixes the composition turned up:

- `max_input_bytes` was applied after the head of the file was read. The header says that bound is applied before the input is read, and the length is a stat, so deciding it needs no idea which source would have handled the bytes. The mode 000 scenario in `test_limits_abi.py` is the test that would have caught it.
- An unreadable input came back `INTERNAL_ERROR`, which tells a caller the library has a bug when the truth is they handed it something it cannot read. `ReadHead` failures are `CORRUPT_INPUT` now.

**`acadsharp/native/Wire/DecodeSession.cs`** is untouched. I mutated it to prove the live-handle counter works (35 tests went red) and restored it.

One thing worth a second pair of eyes: wire version 1's `Polyline` has nowhere to carry a bulge, so this PR decomposes a bulged polyline into `Polyline` runs and an exact `Arc` per bulged span rather than dropping the bulge or bumping the wire. If #46 would rather carry bulges in the record, that is a v2 conversation and this is the shape that does not need one.

spdrman:
On the bulge question you flagged for a second pair of eyes: I put it through Fable and filed the answer as #56.

**The verdict is that the decomposition should go, and the wire should carry bulges as bulges (wire v2).** Your own header comment already says it should: `Flattener.cs` lines 17 to 18 promise "a polyline bulge crosses as a bulge", and the file then converts it to a centre and two angles.

Nothing here is a criticism of the call you made. You were right that dropping the bulge is wrong and right that the decomposition needs no bump, and you flagged it rather than quietly choosing. The review found two things that are only visible from outside the change.

**Block instances are indistinguishable, and that rules the decomposition out on its own.** Under expansion every instance emits the block entity's own handle. A block holding one closed slot, inserted twice, produces two back-to-back groups under one handle with no delimiter. Grouping by handle merges the instances into one path; grouping by contiguity cannot find the seam. Your `test_the_arc_shares_the_polyline_handle` pins exactly the handle-sharing this ambiguity rides on. No flag bit inside v1 fixes it.

**Mirrored insertions bend the arcs the wrong way.** `ArcFromBulge` takes the centre from the transformed endpoints and the untransformed bulge. A reflection flips which side of the chord the centre sits on and the bulge's sign does not follow, so every arc in a mirrored block bends away from where the drawing has it. `Basis.Mirrored` is measured and only used to fail `IsSimilarity`, and that warning's text ("a shape the transform does not preserve") is false for a mirror, which preserves shape exactly.

**A correction to something I told you earlier.** I said a denormal bulge overflows and pushes non-finite doubles onto the wire. At `1e-300` nothing overflows: the radius is about `2.5e299` chords and finite. What crosses is worse, because the two angles come out **exactly equal** and `WIRE.md` never says what an `Arc` with `start_angle == end_angle` means. Non-finite only appears below `2.2e-308`. And the real problem starts fourteen orders of magnitude earlier: a consumer reconstructing the arc's start point drifts from your own vertex by about `1e-17/|b|` chord lengths, so `2.6e-5` at `b = 1e-12` and a full two chord lengths at `1e-17`, where the sweep is already exactly zero. Every arc-to-polyline conversion in every CAD tool leaves `1e-15`-ish residue on "straight" vertices, so that band is not exotic.

**Nothing is being asked of this PR.** #56 is its own piece of work with a full record layout, numerical bounds and a fifteen-point verification plan. The reason it is worth doing soon rather than later is timing: `gh release list` shows zero `acadsharp-*` releases and `acadsharp-rs`'s `src/lib.rs` is still a 1,318-byte stub, so a wire bump today touches nothing outside this repository. That stops being true the first time we cut a release.

Carry on with C1 and C2; those are what the batch is waiting on.

spdrman:
Review round landed. Both Criticals are fixed, mutation-tested and green; the detail is in the updated body. This comment is only the things I changed outside my lane, so the reviewers of those files see them in one place.

**`acadsharp/native/Sources/IDocumentSource.cs`** gains a parameter: `EnumerateView(int index, Func<bool> canceled)`. This is the half of C2 that had nowhere else to live. Checking the flag in `BatchWriter.Peek()` does not work for the case that matters, because the fan-out blocks *inside* a single `MoveNext()` and never comes back to `Peek()`. The flag has to reach the loop that can block. One line here, one in `SyntheticSource` (which ignores it, with a comment saying why: it yields a primitive per item, so the export reads the flag between every one of them), and one in `DecodeSession` to hand `IsCanceled` over. The `Volatile.Read` on the flag itself is already in `main` from your round.

**`acadsharp/docs/ABI.md`** limits table. Three rows got more accurate and a paragraph went in under them:

- `max_entities` counts entities walked as well as records emitted, and either passing it is a refusal. The old wording was already right ("counted across the whole decode, block expansion included"), the code just was not doing it.
- `max_polyline_points` says "points in one record" rather than "vertices in one `Polyline` record", and names a spline's control points and knots. The narrow wording is how the spline arm got missed.
- `max_block_depth` says every kind of expansion, not only insertions, because a `DIMENSION` carries a block and a `HATCH` boundary expands into entities that expand again.

**`acadsharp/native/Abi.cs`**: I dropped my duplicate `Handles.LiveCount` in favour of yours (the rebase brought both). No other change.

One thing worth a second pair of eyes, since it is a behaviour change and not only a fix: `max_block_depth` now bounds dimension and hatch expansion too. A drawing with dimensions nested past 64 that used to decode will now be refused. I think that is right (unbounded is what the Critical was) and the header's rationale reads the same way for all three, but it is a wider bound than the field's name suggests and it is the kind of thing worth disagreeing with now rather than after release.

The only expectation that moved is bit 0 of the geometry prologue on records lifted out of a dimension's block or a hatch's boundary. No record changed order or value, and the diff is `flags=0` to `flags=1` on exactly those lines.

Issue #59, opened 2026-09-13
Outstanding findings from the EPIC G adversarial review, verified against main


spdrman:
EPIC G (#50) went through a four-expert adversarial review: a principal architect, a systems engineer, a safety and reliability engineer, and an API design expert, each reviewing all four PRs independently. They raised roughly sixty findings. The Criticals and most Highs were fixed before merge; this issue is what was not.

Everything below I re-verified against `main` at `30d0fcd` rather than copying from a review comment, because several findings were partly fixed in passing and I did not want to file work that is already done. Where I could not settle something cheaply I say so.

Three already have homes and are **not** repeated here: #56 (wire v2 bulges, which also carries the non-finite-double and `ABI.md`/`WIRE.md` result-code disagreement), and #49 (the throwaway release).

## High: the library ships two doors, and only one has locks

`viprs_acad_describe` and `viprs_acad_entity_count` are still exported, unguarded, from every published archive. Verified: both are in `Exports.cs` outside the `#if VIPRS_ACAD_TEST_EXPORTS` block, and `test_abi_exports.py` still allow-lists them as `SPIKE_EXPORTS`.

They break four rules the header states. They report through negative `int` codes instead of `uint32_t`. They take paths via `Marshal.PtrToStringUTF8`, which is the implicit null termination the header forbids. `Describe` writes an **exception message** into the caller's buffer, against `ABI.md`'s "there is no error string anywhere on this boundary, in either direction". And `viprs_acad_entity_count` calls `DwgReader.Read(path)` directly: no version gate, no limits struct, no cancel flag, on untrusted input.

Three reviewers raised this independently (architect, safety, API). The comment in `Exports.cs` says they stay "until the adapter issue retires the smoke programs with them", and the adapter merged as #55 without retiring them. Their only consumer is `tests/smoke/`, which nothing executes: not CI, not the driver, not pytest.

Worth noting the test that should catch this currently **prevents** the fix: `test_the_spike_exports_are_still_the_only_two` asserts `present == list(SPIKE_EXPORTS)`, so removing them turns it red.

Fix: delete both, delete or relocate `tests/smoke/`, and invert that test to assert `SPIKE_EXPORTS` is empty. If they cannot go yet, put them behind the test configuration so no release build ships them.

## High: nothing automatically runs the evidence

Verified: `ci.yml` has zero references to the conformance consumers or `run.sh`. `release-acadsharp.yml` checks exported symbols and never runs behaviour.

[...]

PR #62, opened 2026-09-13
Run the evidence, and bind the captures to the code that made them

spdrman:
Nothing in this repository ran the conformance consumers. `ci.yml` has zero references to either `run.sh`, and `release-acadsharp.yml` checks that the exported symbols are present and never calls one, so the C consumer's checks and the generated Rust consumer's ran only when a person remembered to run them. The suites that do run on every push read committed captures, and nothing said those captures were a run of the code in the tree, so a change to `Flattener.cs`, `RecordEncoder.cs` or `Primitive.cs` left the whole adapter suite green against a recording of the old behaviour.

The job earned its keep on its first run: it found that every archive this repository has ever published answers `viprs_acad_capabilities_v1` with an empty string. That is the last section below.

## The captures are bound to the code now

`tests/expectations/MANIFEST.json` grows a `shim` block: one sha256 per source the fixture generator compiles, plus a rollup over the sorted paths and the bytes behind them. `tests/test_shim_digest.py` recomputes both and goes red naming the file that moved and `regenerate.py` as the fix.

The source list is read out of the generator's csproj rather than restated in the test. The set that matters is the set that actually compiled into the run, and reading it from there means a new file under `native/Adapter/`, `native/Sources/` or `native/Wire/` is covered by the same glob the generator uses, the day it lands. It also turns out to be slightly wider than the three directories the issue names: the generator compiles `native/Abi.cs` and `native/Probe.cs` too, and both are in the digest. An `<Compile Include>` shape the reader does not understand is an error, not a skip.

A digest over an empty set passes every check, so `TestTheGuardIsNotHollow` asserts the three files the review named are in the resolved set.

## The consumers run in CI

`.github/workflows/acadsharp-conformance.yml`, a file of its own rather than a job in `ci.yml`. Both guards that make a `ci.yml` job expensive read `ci.yml` and nothing else: libviprs-tests' `Hook Mirror (every repo in the org)` and `zstd/tests/test_ci_coverage.py::test_no_step_hardcodes_a_dependency_directory`. So no pairing PR is needed in libviprs-tests, and `test_acadsharp_ci_decision.py` still passes unchanged. ADR 0001 already said a separate workflow file was the cheaper way in once there was an artifact to verify, so its C# coverage section records the revisit rather than being overturned.

On a push that touches `acadsharp/` the job builds the linux/arm64 archive, verifies it before anything trusts it, runs both consumers against the unpacked archive, then publishes the `AbiTest` configuration in the builder image the archive build already made and runs both consumers again against that. The second pass is there because `AbiTest` is the only build carrying the exports that make the exception and handle-leak cases runnable; publishing it in the image that already holds the SDK, clang, the pinned source and a warm cache costs one publish and no downloads. The image name is asked of the driver through a new `builder_image_tag()`, which `_build_docker` uses as well.

## Both runners can be aimed at an archive

They defaulted to the build tree's publish directory and hardcoded `viprs_acadsharp.so`, while the packaged archive ships `lib/libacadsharp_native.so`, so the one artefact a conformance run most wants to be pointed at was the one it could not be pointed at. `tests/conformance/find_shim.sh` knows the three layouts, both runners ask it, the path comes back absolute, and a lib directory outside the checkout is mounted into the container.

## What the job found

Every published archive reports an empty backing version. `viprs_acad_capabilities_v1` answers the one text question on the frozen surface with nothing, which fails two checks in the caller-buffer convention: the sizing call reports a required length of 0, and a one-byte buffer is not too small for a zero-byte string so it returns OK instead of `LIMIT_EXCEEDED`.

It is not a Release-versus-AbiTest divergence and it is not trimming. Same commit, same container, same build context, same publish command, one flag different: `-c AbiTest` gives 139 checks 0 failures and `-c Release` gives 127 checks 0 failures, and the 12-check gap is exactly the test-export cases with the Release run missing nothing else. The variable is the build *context*: `make_dockerfile` stages `native/` and `include/` and never `VERSION`, msbuild's `ReadLinesFromFile` returns nothing for a file that is not there rather than failing, and the generated constant came out as `public const string Text = "";`. I read that file out of the builder image that produced the archive rather than inferring it. `_build_mac_native` shares `_write_build_context`, so the mac archive shipped the same way.

So the build context stages the file, the Dockerfile copies it, and the in-container smoke now asks the library what it is and refuses an empty answer before the archive is packed. It includes the shipped header rather than declaring the frozen struct. `backing_version` is a fact, so the refusal names the value. The guards read the csproj's input list rather than repeating it.

The string the library now reports is `3.7.1-viprs.1`, the artifact version, while the header promises the ACadSharp version and `LINKINFO.json` records `3.7.1`. That disagreement is one of the Mediums on #59, it was invisible while the answer was empty, and I have not decided it.

**`Exports.cs` declared `viprs_acad__test_live_handles` twice**, so `-c AbiTest` fails with `CS0111` and has since the adapter merged, which means neither conformance consumer could be built in the mode that exercises the exception and handle-leak cases. `test_abi_exports.py::_bodies` is keyed by entry point, so the duplicate replaced the first silently and every check there kept passing against whichever body came last. `TestEveryEntryPointIsDeclaredOnce` counts the attributes instead. #63 found the same defect independently.

**The test-export probe ran `nm` on the host** with the failure sent to `/dev/null`, so on a Mac a run against a library that had the exports skipped every case needing them and printed a pass. Both probes now run in the container that links the library, a probe that cannot run is an error, and each run prints which mode it chose.

## What I ran

Everything in containers. Full gate green: `ruff check .`, `ruff format --check .`, `tools/shellcheck-all.sh`, and pytest at **1470 passed / 6 skipped** against **1385 passed / 6 skipped** on `origin/main` at `b324e45`, same 6 skips.

Both consumers, all four ways the CI job runs them, on this tree:

| run | checks |
| --- | --- |
| C against the unpacked archive | 127, 0 failures |
| Rust against the unpacked archive | 106, 0 failures |
| C against the `AbiTest` publish | 139, 0 failures |
| Rust against the `AbiTest` publish | 118, 0 failures |

Every guard added here was watched failing, including the archive build itself refusing to ship a version-less library. The mutations and what they printed are in the comments.

Refs #59


spdrman:
## Every guard, watched failing

A guard nobody has seen fail is not a guard. Each mutation below was applied to a clean tree, run, and restored from a snapshot copy rather than with `git checkout --`, and the tree was confirmed clean afterwards.

**The shim digest.** Before the `shim` block existed, `test_shim_digest.py` was red on 4 of its checks, which is the order the tests landed in.

| mutation | what went red |
| --- | --- |
| appended one comment line to `native/Adapter/Flattener.cs` | `test_each_source_is_byte_for_byte_what_recorded_the_captures[native/Adapter/Flattener.cs]` and the rollup. "`native/Adapter/Flattener.cs` has changed since the captures under tests/expectations were recorded, so every test that reads one is measuring the old behaviour. Rerun acadsharp/tests/fixtures/gen/regenerate.py and read the diff it makes." |
| added `native/Wire/Mutation.cs` | "the shim gained ['native/Wire/Mutation.cs'] and lost [] since the captures were recorded, so they are a run of a different program" |
| renamed `native/Wire/Primitive.cs` away | four red, including the hollow-guard check: "native/Wire/Primitive.cs is not in the set the fixture generator compiles, so the captures are no longer bound to it" |
| changed the generator's `native/Wire/**/*.cs` include to `native/Wire/*.cs` | six red: "'../../../native/Wire/*.cs' is a `<Compile Include>` this reader does not understand, so the shim digest would quietly stop covering it" |

**The duplicate entry point.** Putting the second `viprs_acad__test_live_handles` back, exactly as `main` carries it, turns two red and nothing else: "['viprs_acad__test_live_handles'] is declared more than once in Exports.cs. The project does not compile in any configuration that defines it, and the checks in this file keep passing because they read a map that only holds the last one." and "Exports.cs defines ['TestLiveHandles'] twice, so it will not compile".

**The workflow.**

| mutation | what went red |
| --- | --- |
| renamed the workflow away | six in `test_conformance_workflow.py` plus "nothing runs the conformance consumers again. The ci.yml decision above is only defensible while a separate workflow covers the shim" |
| set `VIPRS_CONFORMANCE_IMAGE` back to the runners' local default | "viprs-rust:arm64 is the runners' local default and no registry serves it, so every consumer step would fail on the pull" |
| dropped the `paths:` gate | "the push trigger is {...} so this builds an archive on pushes that cannot have changed it", and the self-trigger check |
| deleted the C run against the test configuration | "acadsharp/tests/conformance/c/run.sh runs 1 times. It is meant to run against the unpacked archive and against the test configuration" |
| spelled the builder image name out instead of asking the driver | "the workflow spells the builder image name out, which goes stale on the first version bump" |
| made `_build_docker` assemble the tag itself again | "the docker build assembles the tag itself again, so the name the workflow asks for and the name the build writes can drift" |

**The runners.** The `find_shim.sh` behaviour cases were written against fabricated layouts before either runner was wired to it, and the two wiring checks were red until they were: "the c runner resolves the library itself, so the two runners can disagree about which layouts they accept".

## The two things folded in, and why

I did not open issues for these, per the brief. Both are in the blast radius of this one.

**`Exports.cs` declares `viprs_acad__test_live_handles` twice**, so `-c AbiTest` fails with `CS0111` and has done since the adapter merged. That is not a tidy-up: `AbiTest` is the only configuration carrying the test exports, so neither conformance consumer could be built in the mode that exercises the exception and handle-leak cases, and this PR's second CI pass would have been red on its first run. I found it by trying to publish the configuration rather than by reading the file. The second declaration also landed in the middle of the comment belonging to `TestThrow`, which is why it reads as one comment about two different things on `main`.

Why nothing caught it: `test_abi_exports.py::_bodies` returns a dict keyed by entry point, so the duplicate replaced the first silently and every check in that file went on passing against whichever body came last. Had the two been the other way round, `test_every_fallible_export_catches_everything` would have gone red, because the one that survived is the one with the `try`/`catch`. The new class counts the attributes rather than the map.

**The test-export probe ran on the host.** Both runners asked `nm -D --defined-only "$library" 2>/dev/null` outside the container. A macOS `nm` does not take those flags, the failure went to `/dev/null`, and the answer came back "absent", so a run on a Mac against a library that *had* the test exports skipped every case that needs them and printed a pass. That is the same shape as the finding this issue is about, in the tool meant to check it. Both probes now run in the container that is about to link the library, `nm` failing is an error rather than an answer, and each run prints which mode it chose.

## What is not in here

The issue also suggests a CI step running `regenerate.py --check`. `regenerate.py` has no such mode, and giving it one means rerunning the whole corpus through the .NET container, which is a much larger job than the staleness link it would prove. The digest does that link for nothing, so I left the flag alone rather than half-building it.

The conformance job runs the linux/arm64 cell only. The other four cells are the release workflow's business and adding them here quadruples the cost of a push that touches a comment.

spdrman:
## Composed with the other #59 lanes before merging, not after

Three branches for this issue have commits on them right now, and two of them touch files this one touches, so I composed all three on a scratch branch and ran the full gate on the result rather than trusting three separate green runs.

| branch | head | overlap with this PR |
| --- | --- | --- |
| `fix/59-spike-exports-off-the-release-surface` | `dc7aef0` | `acadsharp/native/Exports.cs`, `acadsharp/tests/test_abi_exports.py` |
| `docs/59-publish-the-contracts` | `207d214` | `acadsharp/README.md`, `acadsharp/build_acadsharp.py` |
| `fix/59-mediums`, `perf/59-encoder-and-allocations` | still at `origin/main` | nothing yet |

Both merged with no conflict. On the composed tree: `ruff check .` and `ruff format --check .` clean, `tools/shellcheck-all.sh` exit 0, pytest **1487 passed, 6 skipped**, and the `AbiTest` publish plus both consumers **139 checks 0 failures** and **118 checks 0 failures**. Nothing about composing them changed an answer.

Two things the merge order matters for, though.

**The spike-exports PR needs this one to build.** It moves `viprs_acad_describe` and `viprs_acad_entity_count` inside the `#if VIPRS_ACAD_TEST_EXPORTS` block, which is the right fix, and it leaves the duplicate `viprs_acad__test_live_handles` exactly where `main` has it. That duplicate is the `CS0111` that stops `-c AbiTest` compiling, so on that branch alone the conformance consumers still cannot be built, and this PR's CI job would be red the moment it landed on top. Composed, they are fine. Either order works as long as both land.

**The two empty branches are the ones to watch.** `perf/59-encoder-and-allocations` is aimed at `RecordEncoder.cs` and `Primitive.cs`, and `fix/59-mediums` at several of the same files. Every one of those is in the shim digest this PR adds, so a change to any of them turns `test_shim_digest.py` red with the file named and `regenerate.py` as the instruction. That is the guard doing its job and not a surprise to be worked around: a performance change to the encoder that leaves the captures alone is a change measured against a recording of the old encoder, which is the exact thing this issue is about. Those PRs will need to regenerate the corpus, which needs .NET and docker, and the digest moves with it in the same commit.

spdrman:
`fix/59-mediums` picked up commits after the composition above, and at `eaf986c` it does overlap. Recording the shape while it is still cheap to act on, because all three conflicts are additive against additive in the same few lines rather than two lanes disagreeing about anything.

- `acadsharp/tests/conformance/c/run.sh` and `.../rust/run.sh`: that branch adds an `acadsharp_version` argument passed into the container, this one removes the `test_exports` argument because the probe moved inside. Resolution is both: keep the in-container `nm`, take their new argument, renumber.
- `acadsharp/README.md`: two new sections added at the same anchor. Keep both.
- `acadsharp/native/Exports.cs` and `acadsharp/tests/test_conformance_consumers.py` merged clean.

It also edits `native/Adapter/WarningCodes.cs`, `native/Sources/AcadSharpSource.cs`, `native/Sources/SourceFactory.cs` and `native/Wire/DecodeSession.cs`, which are four of the fifteen sources in the shim digest, so `test_shim_digest.py` will be red on that branch once this one lands and the corpus needs regenerating with it. That is the point of the digest rather than a problem with it: three of those four are on the decode path the committed captures record.

spdrman:
## The job's first red run, chased to the bottom

It is not a Release-versus-AbiTest divergence, and it is not trimming. It is the build context: the archive build never staged `acadsharp/VERSION`, so the library was compiled with an empty version string and answered the one text question on the frozen surface with nothing.

### What the two failures actually are

Both are the same fact. `viprs_acad_capabilities_v1` reports the backing version through the caller-buffer convention, the string was empty, and an empty string is zero bytes long:

- `required > 0` fails, because `required` is 0.
- `cap = 1` returning `LIMIT_EXCEEDED` fails, because 1 byte is not too small for 0 bytes, so the call correctly returns OK.

`Utf8Out.Write` is behaving exactly as `ABI.md` documents. The consumer's expectations are right too. The bug is upstream of both.

### Held the build path fixed, varied only the configuration

This is the experiment asked for. Same commit, same container, same build context, same publish command, one flag different:

| publish | backing version | result |
| --- | --- | --- |
| `-c AbiTest` | `3.7.1-viprs.1` | **139 checks, 0 failures** |
| `-c Release` | `3.7.1-viprs.1` | **127 checks, 0 failures** |

Zero failures in both. The 12-check gap is the test-export cases and nothing else. Diffing the two runs' check lists, the 12 present only in the AbiTest run are the five managed-exception cases, the "still works after five exceptions" case and the six live-handle cases, and **the Release run is missing no check the AbiTest run has**. So the configurations do not disagree about anything the consumer asks.

### Held the configuration fixed, varied only VERSION

Same `-c Release` archive cell, twice, changing nothing but whether `VERSION` reached the build context:

| archive build | outcome |
| --- | --- |
| without `VERSION` staged (what `main` does) | the build refuses: *"the library reports an empty backing version, so `viprs_acad_capabilities_v1` cannot answer what it is"* |
| with `VERSION` staged | archive builds, verifier passes, C consumer **127 checks, 0 failures**, Rust **106 checks, 0 failures** |

That is the variable.

### The mechanism, read rather than inferred

`Viprs.ACadSharp.Native.csproj` reads `$(MSBuildProjectDirectory)/../VERSION` and writes `AcadSharpVersion.g.cs`. `make_dockerfile` copies `native/` and `include/` into `/work` and never copies `VERSION`, and msbuild's `ReadLinesFromFile` returns nothing for a file that is not there rather than failing. Reading the generated file straight out of the cached builder image that produced the archive:

```
// Generated from acadsharp/VERSION by the GenerateAbiFingerprint target.
internal static class AcadSharpVersion
{
	public const string Text = "";
}
```

and out of the same-path Release publish with `VERSION` present:

```
	public const string Text = "3.7.1-viprs.1";
```

That settles trimming on mechanism as well as on measurement: the string was decided by msbuild, before ILC ever ran, and a C# `const string` is inlined at every use site at compile time so there is no field for the trimmer to remove.

`_build_mac_native` goes through the same `_write_build_context`, so the mac archive shipped the same way.

### On the 147 against 127

Those two numbers come from two different programs. Lane 3's `conformance.c` is 1298 lines with 122 `check(` sites; the one on this branch is 1212 lines with 113. The gap between 147 and 127 is 12 test-export cases plus about 8 assertions that lane added, not a configuration difference. The comparison that isolates the configuration is the 139 against 127 above, on one tree.

### What changed here

`_write_build_context` stages the file and the Dockerfile copies it, which fixes the container path and the mac path together. Then the in-container smoke asks the library what it is and refuses an empty answer before the archive is packed, because a question like this belongs where the artefact is made rather than three steps later. It includes the shipped header rather than declaring the frozen struct, for the reason the conformance consumers do. `backing_version` is a fact now, so the refusal names the value.

Guards read the input list out of the csproj rather than repeating it, so a third file added to what the build reads is covered the day it lands.

### One thing I did not decide

The string the library now reports is `3.7.1-viprs.1`, the artifact version, while the header promises the ACadSharp version and `LINKINFO.json` records `3.7.1`. That disagreement is one of the Mediums on #59. It was invisible while the answer was empty and it is visible now. Making the mechanism work is not the same as choosing which of the two strings is right, and I have not chosen.

### Mutations

- Removed `shutil.copy2(VERSION_FILE, ...)`: *"the csproj reads VERSION during the build and the build context does not carry it, so msbuild reads nothing and says nothing about it"*, plus the staged-version check.
- Removed `COPY VERSION`: *"the Dockerfile never copies VERSION, so the build reads VERSION from a path that does not exist in the image"*.
- Neutered the smoke's `needed == 0` refusal: *"the smoke reads the backing version and does not refuse an empty one, which is the state every published archive was in"*.
- Removed `-I"$WORK/include"`: *"the smoke includes the header, so the compile line has to say where it is"*.
- And the real one, end to end: the control archive build above, which refused to ship and exited 1.

Full gate on the merged tree: ruff clean, shellcheck exit 0, pytest **1470 passed / 6 skipped** against **1385 passed / 6 skipped** on `origin/main` at `b324e45`, same 6 skips.

spdrman:
## Composed with #63 at `7ef5304`, and the two need each other

I merged #63 onto this branch in a scratch worktree, resolved it, and ran the whole thing rather than describing it.

### `Exports.cs` does not conflict

Both lanes deleted the same copy of the duplicate `TestLiveHandles`, so git collapses two identical deletions and exactly one survives, the one with the `try`/`catch`. I checked the merged file rather than assuming: one attribute, one method. This was the place an invisible break would have hidden, if we had each deleted a different copy.

### Three conflicts, all additive against additive

| file | shape | resolution |
| --- | --- | --- |
| `conformance/c/run.sh` | #63 adds an `acadsharp_version` argument; this PR removes the `test_exports` argument because the probe moved into the container | keep the in-container `nm`, renumber `acadsharp_version` to `$4` |
| `conformance/rust/run.sh` | same, as two `-e` lines | drop `-e VIPRS_ACAD_TEST_EXPORTS`, keep `-e VIPRS_ACAD_EXPECTED_ACADSHARP_VERSION` |
| `acadsharp/README.md` | two new sections at one anchor | keep both |

### The two version changes are complementary, and neither works alone

This is the part worth the merge attention. #63 decides *which* string `viprs_acad_capabilities_v1` reports, taking the upstream half of `VERSION` in the csproj. This PR makes `VERSION` reach the build at all. Each is inert or wrong without the other:

- **#63 alone**: the archive build fails outright. Its new `<Error>` fires on the empty string, which I ran: `error : /tmp/acadsharp/native/../VERSION reads '', which is not <upstream>-viprs.<revision>`. So #63 landing first turns the release build red rather than shipping something wrong, which is the better failure, but it is still red.
- **This PR alone**: the library reports `3.7.1-viprs.1`, the artifact version, which is the Medium I deliberately did not decide and which #63's consumer check would reject.
- **Composed**: the archive reports `3.7.1`, matching `LINKINFO.json` and what the header promises. `BACKING_VERSION=3.7.1` in the staging log, `backing version="3.7.1"` in the consumer output.

So the Medium I left open is answered by #63, and the answer is only reachable with both.

### The composed tree, measured

| check | result |
| --- | --- |
| archive build | ok, `verify_archive.sh` all invariants hold |
| C consumer against the unpacked archive | **135 checks, 0 failures** |
| Rust consumer against the unpacked archive | **114 checks, 0 failures** |
| pytest | 1529 passed, 6 skipped, **5 failed** |

The five failures are the ones I said to expect, and they are the digest doing its job: #63 edits `WarningCodes.cs`, `AcadSharpSource.cs`, `SourceFactory.cs` and `DecodeSession.cs`, all four in the shim digest, so `test_shim_digest.py` names each one and the rollup. Nothing else in the composed suite broke. Three of those four are on the decode path the committed captures record, so #63 needs a `regenerate.py` run in the same commit as the source change, and the digest moves with it. That is the thing the digest exists to force.

I have not touched `open_path_utf8`, so #63's `INVALID_ARGUMENT` refusal for an embedded NUL is untouched by anything here.

spdrman:
`consumers` is green, and it is green at the same width it was red at.

| run | before | after |
| --- | --- | --- |
| C against the unpacked archive | 127 checks, **2 failures** | 127 checks, **0 failures** |
| Rust against the unpacked archive | 106 checks | 106 checks, 0 failures |
| C against the test configuration | 139 checks | 139 checks, 0 failures |
| Rust against the test configuration | 118 checks | 118 checks, 0 failures |

Same 127. Nothing was narrowed, no check was skipped or relaxed; the archive now has a version to report, so the two checks that were asking about it get an answer. The staging log carries `BACKING_VERSION=3.7.1-viprs.1` and the consumer prints `backing version="3.7.1-viprs.1"` where it printed `""`.

The whole job, archive build included, is 2m21s on `ubuntu-24.04-arm`.

Issue #56, opened 2026-09-13
Carry polyline bulges on the wire as bulges (wire version 2), and retire the polyline-to-arc decomposition


spdrman:
Part of EPIC G (#50), arising from the adversarial review of #55.

Wire version 1's `Polyline` record is a point count, a closed flag and packed XYZ triples. A DWG polyline vertex also carries a **bulge**, `tan(θ/4)` of the arc to the next vertex, negative when the arc runs clockwise, and the record has no slot for it. #55 works around that by decomposing every bulged polyline into `Polyline` runs and one `Arc` per bulged span, all under the source handle, and its author asked for a second opinion on whether that is the shape we want.

It is not, and `Flattener.cs` already says so. Its header comment, lines 17 to 18, promises that "a polyline bulge crosses as a bulge". The file then converts it to a centre and two angles.

**Recommendation: bump to wire version 2, give `Polyline` and `Polygon` a per-vertex bulge array and a normal, and delete the decomposition.** The shim passes the bulge through untouched and the bulge-to-arc conversion moves to the consumer, which is the only layer that knows the tolerance. That is the curves-stay-curves rule applied to the one curve parameter DWG stores this way.

## Why this cannot wait

The cost of changing the record is one digit today and a second protocol version forever after. Verified just now:

- `gh release list` shows **zero** `acadsharp-*` releases. Nothing carrying wire v1 has shipped.
- `acadsharp-rs`'s `src/lib.rs` is a 1,318-byte stub and its batch decoder is still an open issue (H2.2).

This is the last moment a wire change touches nothing outside this repository.

## The defects, verified

Line numbers from the `feat/47-adapter` copy of `acadsharp/native/Adapter/Flattener.cs`.

**1. `Flush` hardcodes `closed: false`.** Only the no-bulge fast path carries `closed`. For a closed bulged polyline the traversed geometry closes, but no record says so, and a consumer cannot recover it except by comparing first and last points, which also matches an open polyline that happens to return home.

[...]

PR #63, opened 2026-09-13
Fix the Medium findings from #59 that live outside the header

spdrman:
Seven of the eight Mediums from #59 I was handed, all of them outside
`include/viprs_acadsharp.h`.

**The header is untouched, so the ABI fingerprint has not moved.** The DWG
read-range macros were the eighth item and they are deliberately not here:
#56 is already editing that file to bump `VIPRS_ACAD_WIRE_VERSION`, which
moves the fingerprint, which `LINKINFO.json` records and
`scripts/verify_archive.sh` checks. Two lanes recomputing the same digest buys
a conflict whose resolution is to recompute it again, so the macros ride along
with #56.

**Merge this after #62, not before.** The two halves of the version fix need
each other and neither works alone. #62 makes `acadsharp/VERSION` reach the
Docker build context at all, which it never did, so every archive ever produced
reported an empty backing version. This decides which string the call reports
once it has one. On its own, this turns the release build red with
`VERSION reads '', which is not <upstream>-viprs.<revision>`; #62 on its own
reports `3.7.1-viprs.1`, which the consumer check added here correctly rejects.
Composed, a real archive answers `3.7.1`, matching `LINKINFO.json` and the
header. Rebased onto `5053b0f` and measured; the run is in the comments.

One thing worth flagging for merge sequencing: this does touch the MSBuild
target that generates `AbiFingerprint.g.cs`, but only the block below it that
generates `AcadSharpVersion.g.cs`. The `GetFileHash` over the header and every
property derived from it are byte for byte what they were.

## What is in it

**`Warning.code` is specified.** `docs/WIRE.md` gave the field's offset and
width and named no value, while the enumeration lived in `WarningCodes.cs`
where no second implementation would ever read it. A consumer was told to
switch on a number the specification does not define and forbidden from
reading the string that does, so the only honest thing it could do with a
warning was count it. The seven codes are now a table in `WIRE.md` with a
meaning each, plus reserved ranges (1 to 999 VIPRS-defined, 1000 and up
backend-specific and ignorable) and the forward rule: a consumer skips a code
it does not know rather than refusing the stream, which is why adding one is
not a wire version bump. `test_warning_codes.py` holds the table and the shim
to each other in both directions.

The comment justifying the numbering said the codes start at 100 "because
SyntheticSource already uses 1". That has not been true for a while:
the synthetic probe is 1100, which lands in the backend range on its own, so
the split fell out rather than being imposed. The stale sentence is gone and a
guard refuses it coming back.

**`DocumentHandle.Untrack`.** `decode_begin` appended to `_decodes` and
`decode_close` removed the handle from the global table, disposed the session
and left the entry, so the list only emptied when the document did. A consumer
re-decoding one view in a loop grew it by a handle per iteration, and
`viprs_acad__test_live_handles` could not see it: that counts the handle
table, which is the one thing `decode_close` did clean up. `Untrack` takes the
same `_decodes` lock that `Track` and `Dispose` take.

**The verifier checks the version fields.** `abi_version` and `wire_version`
were in the required-field list and were never compared against the header
shipped beside them, so a manifest claiming `abi_version: 7` next to a
version-1 header passed every check. It is four lines in the embedded Python
block that already opens the header for hashing, and the header is read, never
written. Three new cases in `test_verify_archive.py`: each field mutated, a
control that the good archive states the header's own numbers, and a header
with the define stripped out, so the check fails loudly rather than skipping
itself into a pass.

**One `max_input_bytes`.** It was enforced in `SourceFactory.FileLength`,
`AcadSharpSource.OpenPath` and `Exports.OpenPathUtf8`, and the three disagreed
on a failed stat: `CORRUPT_INPUT` in the first, `INVALID_ARGUMENT` in the
other two. Which code a caller saw depended on which layer noticed first,
which depends on a race. The two copies are gone and `SourceFactory` owns the
bound, the message and the mapping, where the ordering comment already argued
it belonged. `Exports.OpenMemory` still refuses before it duplicates the
caller's buffer, because a bound applied after the copy has already paid for
the allocation it exists to prevent, but it asks `SourceFactory.CheckInputBytes`
rather than deciding for itself.

Both conformance consumers gained the test the issue asked for: the same
16-byte input, `max_input_bytes` of 8, through `open_memory` and
`open_path_utf8`, asserting both report `LIMIT_EXCEEDED`, that the two codes
are equal, and that neither refusal handed back a handle. Each is paired with
the loosened control, because a bound only ever seen refusing might be
refusing everything.

**`viprs_acad_capabilities_v1` reports the ACadSharp version.** The MSBuild
target read `acadsharp/VERSION` verbatim, so the call wrote `3.7.1-viprs.1`
into a parameter the header documents as "the pinned ACadSharp version" while
`LINKINFO.json`'s `acadsharp_version` carried the upstream half only and
`test_version.py` asserted `"viprs" not in` it. Two frozen surfaces, one
question, two answers. The target now splits on the shim suffix and refuses a
`VERSION` it cannot split. Both conformance consumers assert the exact string
against a value their `run.sh` reads out of `acadsharp/VERSION`, and neither
types it.

**The README's worked example runs.** `--static` is not a flag and never was:
every Linux cell publishes the static archive as part of the same run. The
opening paragraph still called the directory a feasibility spike, and the
Layout block predated `include/`, `docs/`, `scripts/`, the conformance
consumers and the link consumer. `test_doc_examples.py` pulls every
`build_acadsharp.py` invocation out of `acadsharp/README.md` and `MANUAL.md`
and runs its flags and values through the driver's own `--help` and choices,
which is `test_the_build_step_passes_flags_the_driver_accepts` pointed at the
documents instead of the workflow.

**`VIPRSSYN` on both routes.** `ABI.md` documented it for `open_memory` while
`SourceFactory.OpenPath` has always sniffed it too, so a second implementation
written from the specification would have got it wrong in the direction
hardest to notice: it would work on every real drawing. `ABI.md` now covers
both routes and says what the path route sees. The header mention the issue
also asked for is not here, for the fingerprint reason above.

## One thing found on the way

`viprs_acad__test_live_handles` was declared twice inside the same
`#if VIPRS_ACAD_TEST_EXPORTS` block. Two members with one signature is a C#
build error, so the `AbiTest` configuration did not compile, and `AbiTest` is
the configuration both conformance consumers run against. Nothing caught it
because every check in `test_abi_exports.py` keys exports by name into a
dictionary, where the second declaration silently replaces the first. The
duplicate is deleted and `test_shim_source_rules.py` refuses another.

Refs #59


spdrman:
## What I ran, and what I watched fail

### The gate

Everything in containers, arm64, against a clean `origin/main` worktree for the
baseline so the numbers compare.

| | `origin/main` | this branch |
| --- | --- | --- |
| `pytest -q` | 1380 passed, 9 skipped | 1444 passed, 9 skipped |
| `ruff check .` | pass | pass |
| `ruff format --check .` | pass | pass |
| `tools/shellcheck-all.sh` | pass | pass |

The 64 new tests are the guards below plus the parametrised cases each one
fans out into. Nothing was lost.

### The shim actually builds, and both consumers actually run

I published the `AbiTest` configuration for `linux-arm64` against the pinned
ACadSharp 3.7.1 checkout in the .NET 10.0.401 container and ran both
conformance consumers against it. This is not something pytest can do
(ADR 0001), so it was worth doing by hand once.

```
--- capabilities
      backing version="3.7.1" abi=1 wire=1 dwg=1014..1032
ok    capabilities writes the pinned upstream version the header promises
ok    and not the artifact version, which names the archive and not the reader
--- max_input_bytes through both opens
      too large: open_memory=9 open_path=9
ok    an input past max_input_bytes is LIMIT_EXCEEDED through open_memory
ok    and through open_path_utf8
ok    the two open calls report the same code for the same too-large input
ok    and neither refusal handed back a handle
ok    the same input under a bound above its size opens through both calls
```

C consumer: **146 checks, 0 failures**. Generated consumer: **125 checks, 0
failures**. `backing version="3.7.1"` is item 6 proven at run time rather than
argued from the MSBuild source.

That the publish succeeded at all is the proof for the duplicate export: on
`main` the `AbiTest` configuration does not compile, so neither consumer could
ever have run against it.

### Every guard, watched failing

Each mutation was applied to the tree, the guard run, the failure read, and the
file restored from a snapshot taken beforehand.

| # | Mutation | What went red |
| --- | --- | --- |
| 1 | Delete the `100 UNSUPPORTED_ENTITY` row from WIRE.md | `test_every_code_the_shim_emits_is_documented`, plus the table-size control |
| 2 | Rename `HATCH_PATTERN_ONLY` to `HATCH_ONLY` in WIRE.md | `test_every_documented_code_has_the_name_the_shim_uses` |
| 3 | Add a `500 FUTURE_THING` row nothing emits | `test_the_document_defines_no_code_nothing_emits` |
| 4 | `SyntheticSource.ProbeWarningCode` 1100 to 1 | `test_the_synthetic_probe_code_is_in_the_backend_range` |
| 5 | Delete the skip-what-you-do-not-know sentence | `test_the_forward_rule_is_written_down` |
| 6 | Put the backend-range 1100 in the VIPRS table | `test_every_viprs_defined_code_is_below_the_backend_range` and two others |
| 7 | Drop `session.Document.Untrack(handle)` from `decode_close` | `test_decode_close_untracks_what_it_closes` |
| 8 | `Untrack` without `lock (_decodes)` | `test_untrack_removes_under_the_same_lock_as_track` |
| 9 | Delete `Untrack` entirely | `test_the_document_can_untrack_a_decode` and the lock check |
| 10 | A second `MaxInputBytes` comparison in `Exports.cs` | `test_only_the_source_factory_compares_against_it` |
| 11 | A second `new FileInfo(` in `AcadSharpSource.cs` | `test_only_one_place_stats_a_path` |
| 12 | Move `CheckInputBytes` after `Marshal.Copy` | `test_the_memory_route_is_bounded_before_the_copy` |
| 13 | Drop the bound from `SourceFactory.OpenMemory` | `test_both_open_routes_reach_it[OpenMemory]` |
| 14 | `if (length > bounds.MaxInputBytes)` to `if (false)` | `test_the_source_factory_still_enforces_it`, the positive control |
| 15 | `SourceFactory.OpenPath` stops sniffing `VIPRSSYN` | `test_the_path_route_sniffs_it` |
| 16 | ABI.md back to naming `open_memory` only | `test_both_open_calls_are_named_as_selecting_it` |
| 17 | `--static` back in the README | `test_every_flag_is_one_the_driver_accepts` |
| 18 | `--target linux-x86` in the README | `test_every_value_is_one_the_driver_offers` |
| 19 | Every invocation removed from the README | `test_the_documents_carry_invocations`, the positive control |
| 20 | csproj generates `Text` from the artifact version again | `test_the_generator_takes_the_upstream_half` |
| 21 | C `run.sh` stops passing the version in | `test_each_consumer_reads_the_version_out_of_the_file[c]` |
| 22 | The crate types `"3.7.1"` instead of the constant | `test_the_generated_consumer_compares_the_exact_string` and the typed-literal check |
| 23 | A second `viprs_acad__test_live_handles` declaration | `test_no_entry_point_is_declared_twice` |
| 24 | Revert the version check in `verify_archive.sh` | all three new verifier cases, with the good-archive control still green |

Two of them cost a rebuild, because a text guard cannot see either.

**25.** Removing the `max_input_bytes` bound from both memory-route call sites
and republishing the shim turns the new cross-open check into
`too large: open_memory=0 open_path=9`, and the C consumer reports
`146 checks, 3 failures`. That is the check doing the thing it was added for:
the two open calls disagreeing about the same input.

**26.** Putting the duplicate `viprs_acad__test_live_handles` declaration back
and building the `AbiTest` configuration:

```
error CS0111: Type 'Exports' already defines a member called 'TestLiveHandles'
with the same parameter types
```

So that was not a tidy-up. `AbiTest` is the configuration both conformance
consumers run against, and it did not build.

### What is not covered, and why

Nothing runs the conformance consumers automatically, so the run above is a
human running them. That is the separate High finding in #59 about CI and I
have not touched it; what is in this PR is that
`test_conformance_consumers.py` now fails if either consumer stops carrying the
cross-open case, which is the most a suite with no .NET can hold.

spdrman:
## One more mapping, found by removing the stat

Taking `File.Exists` out of `open_path_utf8` quietly took a mapping with it,
and I only noticed when I went looking for what else that call had been
deciding. A path is a pointer and a length on this boundary, so a caller can
hand over one with a NUL inside it. `File.Exists` answered false for that and
the export returned `INVALID_ARGUMENT`. Without it the path reaches
`SourceFactory`, the platform throws on the stat, and the caller gets
`CORRUPT_INPUT`: the input is fine, the argument was not, and that is exactly
the kind of silent remapping this PR exists to remove.

`open_path_utf8` now refuses an embedded NUL explicitly, which is an argument
check rather than a second copy of the bound, and both conformance consumers
assert it.

**Mutation 27.** Delete the NUL check, republish the shim, run the C consumer:

```
FAIL  a path with a NUL inside it is INVALID_ARGUMENT, not a parse failure
147 checks, 1 failures
```

Final state on the rebuilt shim: C consumer **147 checks, 0 failures**,
generated consumer **126 checks, 0 failures**. Gate is 1444 passed, 9 skipped
against a 1380 passed, 9 skipped baseline on `origin/main`, with `ruff check`,
`ruff format --check` and `tools/shellcheck-all.sh` all clean.

spdrman:
## Rebased onto `5053b0f`, and the composed archive reports `3.7.1`

### The conflicts

Three, all additive against additive, all in the two `run.sh` files and the
README. `Exports.cs` auto-merged: #62 and I deleted the same duplicate
declaration, one survives, and my `Untrack` call, the `max_input_bytes` removal
and the NUL refusal all landed around it untouched.

- **`conformance/c/run.sh`** — #62 moved the test-export probe inside the
  container, so it stopped being a positional argument. The upstream version I
  pass in is `$4` now rather than `$5`.
- **`conformance/rust/run.sh`** — same shape: `VIPRS_ACAD_TEST_EXPORTS` is set
  inside now, so only `VIPRS_ACAD_EXPECTED_ACADSHARP_VERSION` is passed through
  `-e`.
- **`acadsharp/README.md`** — both sides kept. The `--static` note sits after
  the build block it explains and #62's Conformance section follows it; the two
  Layout lists merged into one, keeping #62's `docs/` framing and my
  `include/`, `scripts/`, `tests/conformance/` and `tests/link_consumer/` rows.
  The paragraph I had added under Layout now points at the Conformance section
  instead of restating half of it.

### The five failures were correct, and what they surfaced

Straight after the rebase: 5 failed, 1552 passed, 9 skipped, and the five were
`test_shim_digest.py` naming exactly the four sources this PR edits plus the
rollup. The digest doing its job.

I reran `regenerate.py --skip-corpus --only expectations`. **The only lines
that moved are the ones my four edits explain:**

```
-  "recorded": "2026-09-12",
+  "recorded": "2026-09-13",
   "shim": {
-    "sha256": "dbd61743…",
+    "sha256": "3aa1e4d9…",
-      "native/Adapter/WarningCodes.cs":   "8594434b…",  +  "6644f9b7…"
-      "native/Sources/AcadSharpSource.cs":"10804cb9…",  +  "3df1067c…"
-      "native/Sources/SourceFactory.cs":  "f804f5ce…",  +  "b7c860ab…"
-      "native/Wire/DecodeSession.cs":     "cdc91674…",  +  "450109551…"
```

Not one record dump moved. Not one fixture digest, record count or warning
count, across all 24 dumped fixtures. That is what I would have argued from
reading the diff, since the four edits are comments, a dead second copy of a
bound, and a method nothing on the decode path calls, but arguing it and
measuring it are different things.

I also ran the scenario stage, as a check rather than as a commit, and left
`g13_scenarios.json` alone. Across all 34 scenarios every `open_code`,
`decode_code`, `open_detail`, `decode_detail`, `live_handles`, record count and
output byte count is byte for byte what it was. The only fields that moved are
peak RSS and managed allocation, which are run-to-run noise. So the
`max_input_bytes` refactor refuses with the same code and the same sentence it
always did, including `limits/max_input_below_size` and both unreadable-copy
controls, and committing 229 lines of RSS churn would have buried that rather
than shown it.

### Both halves are in

I built the real archive on this branch and ran both consumers against the
unpacked `acadsharp-linux-arm64.tgz`, which is the route CI now takes:

```
library: …/unpacked/acadsharp-linux-arm64/lib/libacadsharp_native.so
upstream: 3.7.1
test exports: absent, so the cases that need them are skipped
      backing version="3.7.1" abi=1 wire=1 dwg=1014..1032
ok    capabilities writes the pinned upstream version the header promises
ok    and not the artifact version, which names the archive and not the reader
      too large: open_memory=9 open_path=9
ok    the two open calls report the same code for the same too-large input
```

C **135 checks, 0 failures**. Generated consumer **114 checks, 0 failures**.
The same numbers lane 1 got. `LINKINFO.json` in that archive reads
`artifact_version: 3.7.1-viprs.1`, `acadsharp_version: 3.7.1`, and the call now
answers `3.7.1`, so the run-time surface and the link-time manifest agree for
the first time.

One thing worth keeping: the `<Error>` I put in the MSBuild target is what
turns the empty-`VERSION` failure into a build error naming the file, rather
than a library that links fine and answers with nothing. #62's smoke catches it
at the archive stage; this catches it at compile time. That is the half of the
pair that only exists because both changes are in.

### Gate on the rebased branch

| | `origin/main` (`5053b0f`) | this branch |
| --- | --- | --- |
| `pytest -q` | 1493 passed, 9 skipped | 1557 passed, 9 skipped |
| `ruff check .` | pass | pass |
| `ruff format --check .` | pass | pass |
| `tools/shellcheck-all.sh` | pass | pass |
| archive build + `verify_archive.sh` | — | all invariants hold |
| conformance, against the archive | — | C 135/0, Rust 114/0 |

PR #64, opened 2026-09-13
Carry polyline bulges on the wire as bulges (wire version 2), and retire the polyline-to-arc decomposition

### LINKED ISSUE #56: Carry polyline bulges on the wire as bulges (wire version 2), and retire the polyline-to-arc decomposition

spdrman:
Closes #56

A polyline's bulges cross the boundary as bulges. Wire version 2 gives records 4 and 9 a normal and a per-vertex bulge array, and the polyline-to-arc decomposition is gone.

`Flattener.cs` has promised this since G1.3. Its header comment, lines 17 to 18, says "a polyline bulge crosses as a bulge", and the file then converted it to a centre and two angles. It is true again.

## Why the record had to change rather than the decomposition get fixed

Three things the decomposition lost, and only the first was ever argued about.

**Closed-ness.** A run of `Polyline` records has nowhere to say the path closes, and `Flush` hardcoded `closed: false` anyway. Comparing the first and last point cannot tell a closed polyline from an open one that happens to return home. `g13_slot.dwg` went out as two runs and two arcs, both runs saying `closed=0`, for a shape that is closed.

**Instance identity, and this one rules the decomposition out on its own.** Under block expansion every instance emits the block entity's own handle, so three insertions of one slot arrived as twelve records under handle `4C` with no delimiter anywhere. Grouping by handle merges three slots into one path; grouping by contiguity cannot find the seams. It was sitting in two committed artefacts already: `real_AC1018.txt` and `real_AC1032.txt` both carry handle `782`, a polyline in a block inserted twice, as two back-to-back groups under one handle. It is now exactly two records.

**Accuracy.** A centre and two angles reconstructed from a small bulge drift from the vertices either side of them by roughly `1e-17/|b|` chord lengths: at `1e-12` an endpoint is already `2.6e-5` out, and at `1e-17` the two angles come out exactly equal, which `WIRE.md` never defined a meaning for. Bulges of about `1e-15` are what every arc-to-polyline conversion in every CAD tool leaves on vertices it considers straight. The bulge now crosses as the number the file holds and the consumer converts in the well-conditioned direction, whenever it wants to, at its own tolerance.

## The commits

| | |
| --- | --- |
| C1 | `test(wire): pin the seven wire-version declarations to each other` |
| C2 | `tooling: harness --raw, regenerate --write-only and --only corpus` |
| C3 | `corpus: slot, slot block, mirrored bulge and non-finite fixtures` |
| C4 | `wire v2: version bump, foreign version is ABI_MISMATCH, Polyline and Polygon gain a normal and a bulge array` |
| C5 | `flattener: one Polyline per polyline, bulges cross as read, closed survives, a mirror negates` |
| C6 | `flattener: a hatch loop of lines and circular arcs is one Polygon` |
| C7 | `flattener: non-finite geometry crosses as a warning, and the encoder backstops it` |

C3 records every new fixture under the decomposition that still existed, so the "before" is in the history rather than in a paragraph. C4 and C5 are deliberately separate: C4's diff is the format, C5's is the semantics, and a reviewer reading one mixed diff cannot tell an added field from a collapsed arc.

Every commit is green on its own: 1568, 1568, 1620, 1649, 1685, 1686, 1694 passed against `main`'s 1560, 6 skipped throughout.

## What the record looks like

| off | size | field | rule |
| --- | --- | --- | --- |
| 16 | 4 | `point_count` | `n`, at most `max_polyline_points` |
| 20 | 4 | `closed` | 0 or 1 |
| 24 | 4 | `bulge_count` | `bc`, either 0 or equal to `point_count` |
| 28 | 4 | `reserved1` | 0 |
| 32 | 24 | `f64 nx, ny, nz` | the entity's normal |
| 56 | 24n | vertices | |
| 56+24n | 8bc | bulges | absent when `bc` is 0 |

`length = 64 + 24n + 8bc`. Record 9 takes the identical payload with `closed` always 1, so one reader serves both and a hatch loop that turns out to carry a curved edge needs no second shape.

`bulge_count` is 0 or `point_count` and nothing between, which is record 8's `weight_count` rule: anything else leaves a consumer working out which spans an array covers, which is a length it inferred rather than one the record gave it. Both counts are checked and so is the length, because the second does not follow from the first: a producer that writes the array and leaves the count at zero emits a record whose four framing rules all pass and whose trailing numbers nobody reads.

Those rules live in a `polyline_shape` helper beside each of the three parsers rather than inside the framing loop. A framing parser walks record headers knowing nothing about what a payload means, which is exactly what lets it skip a type it has never heard of, and a layout rule smuggled in there makes the parser wrong for every record type the day one of them changes.

## Counts

Every re-recorded expectation, before and after:

| fixture | records | output bytes | what moved |
| --- | --- | --- | --- |
| `real_AC1032.dwg` | 361 → 341 | 35,964 → 35,716 | Arc 23 → 3, Polyline 25 → 25 |
| `real_AC1018.dwg` | 354 → 334 | 34,784 → 34,536 | Arc 23 → 3, Polyline 25 → 25 |
| `g13_hatch.dwg` | 13 → 10 | 1,860 → 1,836 | Polygon 1 → 3, Arc 1 → 0, Line 3 → 0, Warning 7 → 6 |
| `g13_polyline.dwg` | 8 → 7 | 1,204 → 1,252 | Arc 1 → 0 |
| `g13_wide_polyline.dwg` | 5 → 5 | 99,180 → 99,212 | the normal, and no bulge array |
| `g13_slot.dwg` | new, 5 | 1,012 | one Polyline, `closed=1`, `bulges=[0,1,0,1]` |
| `g13_slot_block.dwg` | new, 7 | 1,396 | three Polylines, one per instance |
| `g13_mirrored_bulge.dwg` | new, 5 | 980 | one Polyline, `bulges=[0,-0.5,0]`, no warning |
| `g13_nan_bulge.dwg` | new, 6 | 1,160 | two NON_FINITE_GEOMETRY warnings, no geometry |

Eighteen fixtures are unchanged. `g13_hatch.dwg` is the only existing fixture whose bytes moved, and it moved because it gained a clockwise arc loop.

The wide polyline grows by exactly 32 bytes, the normal, and no more: the bulge array is left out when every span is straight. Emitting it unconditionally makes that record 131,980 bytes instead of 99,212, which is 32,768 bytes of zeroes for a polyline with no curve in it.

## Evidence

**Both consumers, against a NativeAOT shim published from this branch.** The C one: 170 checks, 0 failures. The generated one: 149 checks, 0 failures. Both report `abi=1 wire=2 dwg=1014..1032`, and both pass "every field of every record is at the offset docs/WIRE.md gives it" against a synthetic document whose polylines and polygons now carry a bulge per vertex, because the empty array is the easy case and a probe set that only saw it would leave the trailing array unread.

**The archive.** `metadata/LINKINFO.json` says `wire_version 2`, `abi_version 1`, `abi_fingerprint a7805641d1a3b64f`. Nothing is hand-pinned: the fingerprint is the first eight bytes of the header's sha256, `a7805641d1a3b64f9d3bac055c29f61eea5b13a228d2a3f46fd67a4c7e451a4b`, and the packaged `libacadsharp_native.so` returns `a7805641d1a3b64f` when asked. `VIPRS_ACAD_ABI_VERSION` stays `1u`, because no declaration in the header changed shape.

**Amplification, re-measured by hand rather than re-recorded.** `g13_many_inserts.dwg` has no polyline in it, so the benchmark is not a thing this change moves and rerunning it would only churn RSS numbers. Decoded at the tip: 2,481,264 bytes, 38 batches, `live_handles 0`, and `g13_insert.dwg` still 972 bytes. Both are exactly the committed values.

**Fifteen mutations, each run once and each red.** In the comment below, with the output.

## Two deviations from #56's text that want a reviewer's eye

**A mirrored polyline no longer raises `NON_UNIFORM_BLOCK_SCALE`.** #56's degenerate-case table has "mirrored insertion: every bulge negated" and "non-uniform insertion: emitted as read, plus `NON_UNIFORM_BLOCK_SCALE` as today", and does not say the mirror stops warning. I think it has to. The warning's message says the transform "does not preserve" the shape, which is false for a reflection: both scale magnitudes are 1 and, with the bulge negated, the record is exact. So `Basis` gains `IsUniform`, which allows a mirror, and `IsSimilarity` becomes `IsUniform` plus the handedness check. Records 4 and 9 use the first; `Arc`, `Circle` and `Ellipse` keep the second, because their angles are counter-clockwise and nothing in them follows a reflection yet. `g13_nonuniform.dwg` still warns, which is the control.

**The finiteness guarantee is scoped to geometry records, types 3 to 10.** #56 asks for "a producer guarantee that no record carries a non-finite `f64`", without qualification. `ViewBegin`'s extents are outside it on purpose: they are a bounding box the source reports rather than a shape anybody draws, and a view holding nothing has no finite one, so promising a number there would mean inventing one. That is a real hole rather than a tidy line, and it is item 4 of the follow-up.

## Folded in

Two things found on the way, both in the diff rather than in a new issue, and described in the comment below: `PROVENANCE.md`'s corpus table had lost its header row, and `max_polyline_points` needed saying out loud now that the record carries a second array of the same length.

## Follow-up

#65 carries what #56 scopes out: OCS to WCS across five record kinds, `Arc` under a mirrored insertion, `NON_UNIFORM_BLOCK_SCALE`'s wording, and `ViewBegin`'s extents.


spdrman:
## Fifteen mutations, each red

Every guard on this branch was mutated once against the tree as it stands, the red read, and the mutation restored from a snapshot taken before it was applied rather than through `git checkout --`. `git status --porcelain` was empty before and after each one and named exactly the file being mutated in between, which the harness checks and refuses on.

A mutation that did not go red would be a finding rather than a tick. None of them was.

### 1. The bulge crosses as read

`Array.Clear(b, 0, b.Length)` after `BulgeArray` fills it.

```
### g13_polyline
  "check": "record 4 differs
    expected: 00004 Polyline handle=49 ... bulges=[0.000000,0.500000,0.000000] normal=[(0.000000,0.000000,1.000000)]
    actual:   00004 Polyline handle=49 ... bulges=[0.000000,0.000000,0.000000] normal=[(0.000000,0.000000,1.000000)]"
### g13_slot
  "check": "record 4 differs
    expected: 00004 Polyline handle=49 ... bulges=[0.000000,1.000000,0.000000,1.000000]
    actual:   00004 Polyline handle=49 ... bulges=[0.000000,0.000000,0.000000,0.000000]"
```

### 2. Nothing is decomposed

An `Arc` re-emitted beside the polyline when a bulge is non-zero.

```
  "output_bytes": 1348,      (1252 clean)
  "check": "record 5 differs
    expected: 00005 Polyline handle=4A flags=0 n=3 closed=0 pts=[(20.000000,0.000000,0.000000);...]
    actual:   00005 Arc handle=49 flags=0 c=[(6.250000,5.000000,0.000000)] r=6.250000 a0=-0.927295 a1=0.927295"
```

### 3. Closed-ness survives

`closed: false` passed instead of `lw.IsClosed`.

```
  "check": "record 4 differs
    expected: 00004 Polyline handle=49 flags=0 n=4 closed=1 pts=[...] bulges=[0.000000,1.000000,0.000000,1.000000]
    actual:   00004 Polyline handle=49 flags=0 n=4 closed=0 pts=[...] bulges=[0.000000,1.000000,0.000000,1.000000]"
```

### 4. One entity is one record under block expansion

Wire version 1's shape in miniature: one record per span rather than one per entity.

```
### g13_slot_block
  "output_bytes": 2260,      (1396 clean)
  "check": "record 4 differs
    expected: 00004 Polyline handle=4C flags=1 n=4 closed=1 pts=[(0,0,0);(20,0,0);(20,10,0);(0,10,0)] bulges=[0.000000,1.000000,0.000000,1.000000]
    actual:   00004 Polyline handle=4C flags=1 n=2 closed=0 pts=[(0.000000,0.000000,0.000000);(20.000000,0.000000,0.000000)] bulges=[]"
```

### 5. The array is absent when every bulge is zero

`double[] b = new double[n];` unconditionally.

```
### g13_wide_polyline
  "batches": 3,              (1 clean)
  "output_bytes": 131980,    (99212 clean)
  "check": "record 4 differs ... bulges=[] -> bulges=[0.000000, x4096]"
```

32,768 bytes of zeroes, and it crosses a batch boundary twice on the way.

### 6. A reflection negates the bulge

`b[i] = v;` regardless of `mirrored`.

```
### g13_mirrored_bulge
  "check": "record 4 differs
    expected: 00004 Polyline handle=4C flags=1 n=3 closed=0 pts=[(0,0,0);(-10,0,0);(-10,10,0)] bulges=[0.000000,-0.500000,0.000000]
    actual:   00004 Polyline handle=4C flags=1 n=3 closed=0 pts=[(0,0,0);(-10,0,0);(-10,10,0)] bulges=[0.000000,0.500000,0.000000]"
```

Through the midpoint formula that is the arc at `(-7.5, 5)` rather than `(-12.5, 5)`: the wrong side of the chord, which is where every arc in a mirrored block sat before this branch.

### 7A. The walk's finiteness guard, and the encoder behind it

`Finite(...)` deleted at both `Walk` yield sites. The encoder backstop is unreachable by construction, so the only honest way to know it works is to make it reachable and watch it fire.

```
### g13_nan_bulge
  "decode_code": "INTERNAL_ERROR",
  "batches": 0,
  "output_bytes": 0,
  "live_handles": 0,
  "check": "record 4 differs
    expected: 00004 Warning handle=49 flags=0 code=NON_FINITE_GEOMETRY message=\"Polyline value 13 is NaN, ...\"
    actual:   00004 Polyline handle=49 flags=0 n=3 closed=0 pts=[...] bulges=[0.000000,NaN,0.000000]"
```

Exit 0 and `live_handles 0`: the failure comes back as a result code and nothing is left open.

### 7B. And with the backstop disabled too, the bytes reach the wire

7A plus `if (false)` on `GeometryDoubles`'s check. Two files, both snapshotted and both restored, and `git status` named exactly those two in between.

```
  "decode_code": "OK",
  "raw_bytes": 1116,
--- non_finite_exponents_in_output = 4 over 1116 bytes (0 with the guards in place)
```

That number is the scan the `finiteness/nothing_non_finite_reaches_the_wire` scenario records, run over `decode --raw`'s output. It reports 0 on the shipped code and 2 on a hand-built control holding one NaN and one infinity, so it is not a scan that had stopped scanning.

### 8. The seven wire-version declarations move together

`WireFormat.Version = 3` and nothing else.

```
E  AssertionError: the wire version is not the same everywhere:
   {'the published header': 2, "the shim's ABI constants": 2, "the shim's wire format": 3,
    'the C consumer': 2, "the crate's parser": 2, "the crate's layout table": 2,
    'the reference parser': 2}
```

### 9. Both consumers read every new field at the documented offset

The normal's `nz` swapped with `bulge[0]` in `SyntheticSource`, which is slot 2 against slot 3 + 3n.

```
=== C consumer ===
      first mismatch: vertex record: got 415.250000, wanted 402.250000
FAIL  every field of every record is at the offset docs/WIRE.md gives it
170 checks, 1 failures
=== generated consumer ===
      first mismatch: vertex record: got 415.25, wanted 402.25
FAIL  every field of every record is at the offset docs/WIRE.md gives it
149 checks, 1 failures
```

`415.25` is `probe(4, 15)`, the first bulge of a four-vertex polyline; `402.25` is `probe(4, 2)`, the normal's `nz`. Each consumer names the slot on its own.

### 10. The `bulge_count` rule, in all three parsers

Deleted separately in each.

```
--- the crate's parser (wire.rs)
FAIL  a Polyline bulge_count that is neither 0 nor point_count
FAIL  a Polygon bulge_count that is neither 0 nor point_count
149 checks, 2 failures

--- the C parser (vacb.c)
FAIL  a Polyline bulge_count that is neither 0 nor point_count
FAIL  a Polygon bulge_count that is neither 0 nor point_count
170 checks, 2 failures

--- the reference parser (test_wire_protocol.py)
FAILED TestThePolylineAndPolygonPayloadRules::test_a_bulge_count_that_is_neither_zero_nor_the_point_count[4]
FAILED TestThePolylineAndPolygonPayloadRules::test_a_bulge_count_that_is_neither_zero_nor_the_point_count[9]
E  assert (0, (4, 3)) == (3, None)
```

### 11. `WIRE.md` says it, one sentence at a time

Five deletions, five different tests.

```
the sign convention sentence  -> FAILED test_the_bulge_sign_convention_is_documented
the midpoint formula line     -> FAILED test_the_midpoint_formula_is_documented
the finiteness guarantee      -> FAILED test_the_finiteness_guarantee_is_documented
the bulge_count refusal row   -> FAILED test_each_payload_refusal_has_a_row[`bulge_count` ...]
the length formula            -> FAILED test_the_polyline_payload_is_documented
```

The first is the one worth looking at. The `Arc` paragraph two records down still says "counter-clockwise", so a test that searched the whole document would have stayed green with the `Polyline` section saying nothing at all about which way a positive bulge turns. It reds because the test slices record 4's section out first.

### 12. The sign convention is pinned to numbers, not to prose

`cross(d, N)` becomes `cross(N, d)` in the test's own copy of the formula.

```
E  AssertionError: the arc midpoint came out at (7.5, 5.0, 0.0), ...
E  assert 1.25 == 6.25 ± 1.0e-09
E  assert (-7.5, 5.0, 0.0) == approx((-12.5, 5.0, 0.0))
E  assert (25.0, 5.0, 0.0) == approx((35.0, 5.0, 0.0))
4 failed, 57 passed
```

`1.25` rather than `6.25` from the centre wire version 1 recorded for the same arc, and the two hatch loops move with it.

### 13. A hatch loop of lines and arcs is one polygon

`TryPolygon`'s arc arm made to return false again.

```
### g13_hatch
  "output_bytes": 2332,      (1836 clean)
  "check": "record 5 differs
    expected: 00005 Polygon handle=4A flags=0 n=4 closed=1 pts=[...] bulges=[0.000000,1.000000,0.000000,0.000000]
    actual:   00005 Warning handle=4A flags=0 code=HATCH_LOOP_NOT_POLYGON message=\"boundary loop 1 carries a curve a closed polygon cannot express...\""
```

spdrman:

Three things turned up on the way. Two are in the diff; the third fixed itself.

**`PROVENANCE.md`'s corpus table had lost its header row.** The G1.3 section's rows began straight after the paragraph with no `| File | sha256 | What it holds |` line and no separator, so none of the twenty-five rows rendered as a table anywhere that reads markdown. I noticed it adding four rows of my own. The header is back, and `g13_hatch.dwg`'s row is rewritten because the fixture gained a clockwise arc loop and its digest moved with it.

**`max_polyline_points` needed saying out loud.** The bound is on vertices, and wire version 2 gives the record a second array of the same length, so there is now an obvious wrong way to apply it that halves the bound a caller thinks it set. Two scenarios pin it: `g13_slot.dwg` is four vertices and four bulges, it decodes at `--max-polyline 4`, and it is refused at 3. Without the second one, "the bulges do not count" would also be what a bound that had stopped applying looks like.

**The AbiTest configuration did not compile, and `main` fixed it first.** `Exports.cs` declared `TestLiveHandles` twice inside `#if VIPRS_ACAD_TEST_EXPORTS`, once as an expression body and once as a block, so the publish failed with CS0111 and neither conformance consumer could be built against a shim at all. The comment above the first one was the tell: two paragraphs run together with no blank line, the first about `TestThrow` and the second about the handle count, with the duplicate wedged between them. I fixed it here because it was in the way, and the rebase onto the current `main` dropped my commit as empty: #62 or #63 had reached the same line. Nothing to review, and worth knowing it was found twice independently.

## What a reviewer might want to look at first

`docs/WIRE.md`'s record 4 section and `Flattener.OnePolyline` are the whole change. Everything else follows from those two.

The place I would look hardest is `ArcEdge` in `Flattener.cs`. A boundary arc's direction is a flag rather than a sign on the sweep, upstream reads a clockwise edge as the arc from `2*pi - end` to `2*pi - start`, and the corpus had only counter-clockwise loops, so a converter that ignored the flag was right on every fixture there was. `g13_hatch.dwg` gains a clockwise loop whose arc bulges inward for exactly that reason: getting it wrong puts the arc outside the rectangle and the bulge comes out `+1` instead of `-1`. The conversion goes through ACadSharp's own `ToEntity`, which is the same object the not-a-polygon path emits as an `Arc` record, so the two paths cannot disagree about what an edge means.

PR #68, opened 2026-09-13
Retain the module section so --gc-sections cannot drop it

### LINKED ISSUE #67: linux/x64 cannot link the documented consumer recipe: __stop___modules is undefined under rust-lld
The first real dispatch of `release-acadsharp.yml` published four of five targets. `linux/x64` failed, and it failed in the documented consumer recipe rather than in the build.

Run: https://github.com/libviprs/libviprs-dep/actions/runs/34766314854 (dispatched on `main` `87861e7` with `version=3.7.1-viprs.0`).

```
rust-lld: error: undefined symbol: __start___modules
rust-lld: error: undefined symbol: __stop___modules
>>> referenced by main.cpp:211 (nativeaot/Bootstrap/main.cpp:211)
>>> bootstrapperdll.o:(InitializeRuntime()) in archive lib/libacadsharp_native_init.a
collect2: error: ld returned 1 exit status
error: could not compile `consumer` (bin "consumer")
verify_archive.sh: one or more invariants failed
```

The verifier refused the archive, so nothing broken was published. `linux/arm64`, `musl/x64`, `musl/arm64` and `mac/arm64` all built, verified and uploaded.

## What the run proved, before the failure

Worth recording because it is the first time any of it has been true:

- **`musl/x64` built for the first time.** It had only ever crashed ILC in its own `Parallel.For` under Rosetta, four attempts, and ADR 0001 called that emulation rather than the code. On a native x64 runner it just works, which settles that.
- **`mac/arm64` built for the first time.** Every prior run used a synthesised Mach-O because NativeAOT has no cross-OS compilation.
- **`--version` reached the build.** The published `LINKINFO.json` reads `artifact_version: 3.7.1-viprs.0`, the dispatch override, not the committed `3.7.1-viprs.1`.
- **`VERSION` reached the build context**, so `acadsharp_version` is `3.7.1` rather than the empty string every archive before #62 shipped.
- **The notes named the gap**: "Targets that did not publish: `linux/x64` did not publish (job `build-linux`, failure). This release is incomplete."
- Both published arm64 archives link and run on a machine with **no .NET**, control included:
  ```
  control: no dotnet on PATH
  abi_version = 1   abi_fingerprint = a7805641d1a3b64f
  backing version = 3.7.1   wire_version = 2   dwg range = 1014..1032
  ```

## Root cause

Measured, not inferred. From the published `linux-arm64` archive, which has the identical two-archive shape:

```
lib/libacadsharp_native_init.a   -> 1 member: bootstrapperdll.o
                                    does NOT contain a __modules section
lib/libacadsharp_native.a        -> viprs_acadsharp__viprs_acadsharp.o
                                    DOES contain the __modules section
```

And the failing consumer's link line, from the job log:

```
"-Wl,--whole-archive" "-lacadsharp_native_init" "-Wl,--no-whole-archive" "-lacadsharp_native"
"-fuse-ld=lld"  "-B<sysroot>/lib/rustlib/x86_64-unknown-linux-gnu/bin/gcc-ld"  "-Wl,--gc-sections"
```

So the chain is:

1. `--whole-archive` forces `bootstrapperdll.o` in. Its `InitializeRuntime()` references `__start___modules` and `__stop___modules`.
2. Those are **linker-synthesised encapsulation symbols**. No object defines them. A linker creates them only when a section literally named `__modules` is present in the output.
3. The section that provides `__modules` lives in `libacadsharp_native.a`, which is linked **lazily** (`-lacadsharp_native`, no whole-archive).
4. Archive member selection is driven by **defined** symbols. Since nothing defines `__start___modules`, the reference cannot pull `viprs_acadsharp__viprs_acadsharp.o` in on that basis.
5. If no other live reference happens to pull that member first, the section never enters the output, the symbols are never synthesised, and the link fails.

**Why only x64.** Rust uses `rust-lld` by default for `x86_64-unknown-linux-gnu`, which the link line shows (`-fuse-ld=lld`, the `gcc-ld` shim directory). For `aarch64-unknown-linux-gnu` it still goes through `cc` to GNU ld. I am marking this half **inferred**: a successful link prints no command line, so the arm64 job's log does not show its linker, and I have not measured it. What is measured is that the same archive shape links on arm64 and musl and not on x64 under lld.

This is also **ordering-sensitive and therefore luck-shaped**. It is the same class as the `.init_array` ordering bug PR #54 found, where the archive linked cleanly and segfaulted at exit.

## Why nothing caught it earlier

The two-crate cargo consumer smoke that found this was added in #54, precisely because `cargo:rustc-link-arg` does not propagate and the recipe needed proving end to end. It has only ever run on arm64, on this machine and in the `consumers` CI job. **This release run is the first time the documented recipe has been exercised on x64 at all.**

So: a check added for one bug found a second one, on its first run on a platform nobody had tried it on.

## Options

I have not picked one. Each has a real cost and the ABI's consumers are about to be frozen against whichever we choose.

**A. Whole-archive the main archive too.** `cargo:rustc-link-lib=static:-bundle,+whole-archive=acadsharp_native`. Guarantees `__modules` is in the output. Propagates correctly, since it is a `rustc-link-lib` modifier. Costs binary size: it pulls all ~250 members rather than what the program reaches, and `--gc-sections` will not recover all of it.

**B. Put `bootstrapperdll.o` back in the main archive and whole-archive that.** One archive, one directive, no ordering question. Same size cost as A, and it undoes the split that #54 introduced, so the reasoning for the split needs re-reading before anyone does this.

**C. Make the init archive carry the `__modules` section too.** Smallest link-time footprint, but it means synthesising or duplicating a section the runtime owns, and a duplicate `__modules` across two archives may merge in ways that are worse than the disease.

**D. Order the directives so the main archive precedes the whole-archived init.** Cheap to try and I suspect it does not work: the init archive must come first or the link fails on `RhRegisterOSModule`, which is how the current order was arrived at in #54. Worth one experiment to confirm rather than assume.

My weak preference is A, because it is the smallest change to a recipe that is already documented in `LINKINFO.md` and `MANUAL.md`, and because size is a cost we can measure rather than a correctness risk we cannot. But someone should measure the size delta before committing to it.

- Reproduce first, in a container, before changing anything: build the linux/x64 archive and run `link_consumer_smoke.sh` against it. It must fail with `__stop___modules`. A fix that was never watched failing is not a fix.
- After the fix, the same smoke links **and runs** the binary, not just links it. PR #54's own finding was that a link-only check cannot tell a working archive from one that aborts on first call.
- Run it on both x64 and arm64. The bug is that one platform was never covered, so a fix proven on one platform repeats the mistake.
- Measure the binary size before and after, and put both numbers in the PR. Option A's whole cost is size.
- Mutation: revert to `-lacadsharp_native` without the modifier and watch x64 go red again.

## Acceptance

- [ ] `linux/x64` builds, verifies and publishes.
- [ ] The cargo consumer smoke links and runs on x64 and arm64, both glibc and musl.
- [ ] `LINKINFO.md` and `MANUAL.md` carry the corrected recipe with the reason, the way they already carry the `-bundle` reason.
- [ ] The size cost of whichever option is chosen is measured and recorded.
- [ ] The failure was reproduced once before the fix and the red output is in the PR.
- [ ] A re-dispatch of `release-acadsharp.yml` publishes five of five.

spdrman:
The release's linux/x64 cell could not link the recipe we document, and
the other four targets could, so four archives shipped and one did not.

## What was actually wrong

My analysis in the issue had the linker half right and the mechanism
wrong. I said the archive member carrying `__modules` was never selected
out of the archive. It is always selected: `viprs_acad_abi_version` is
defined in the same object, so the link pulls it in regardless. The
section is there and then gets collected.

ILC puts the runtime's module headers in a section called `__modules`,
and the bootstrapper walks it through `__start___modules` and
`__stop___modules`, the symbols a linker synthesises around any section
whose name is a C identifier. Nothing relocates against the section, so
those two symbols are its only references. lld has defaulted to
`-z start-stop-gc` since version 13, which says a reference through an
encapsulation symbol is not a reason to keep a section alive, and rustc
asks for `--gc-sections`. So lld collects `__modules` and then has
nothing for `__start___modules` to point at. lld says so itself in the
failure, and I had not read far enough down its message:

```
rust-lld: error: undefined symbol: __start___modules
>>> the encapsulation symbol needs to be retained under --gc-sections
>>> properly; consider -z nostart-stop-gc
```

GNU ld keeps the section. rustc uses lld on `x86_64-unknown-linux-gnu`
and `cc` with GNU ld on aarch64, which is the whole of the split.

## The fix

Staging sets `SHF_GNU_RETAIN` on the section, so the archive carries its
own requirement. The alternative is asking consumers to pass
`-z nostart-stop-gc`, and a consumer cannot: `cargo:rustc-link-arg` does
not travel from a dependency's build script to the binary that links it,
which is the same limitation that put the initialiser in its own archive
in the first place. The documented recipe is unchanged. One byte of a
51 MB archive moves.

`objcopy` has no spelling for the flag (binutils 2.40 accepts alloc,
load, noload, readonly, debug, code, data, rom, exclude, share, contents,
merge and strings, and nothing else), so `scripts/retain_sections.py`
sets it directly. Staging refuses to ship a static archive if the section
is absent or the flag cannot be set, the same way it already refuses when
the merge would drop an object.

## Measured

On the NAS, which is native x86_64, not emulated. rustc 1.98.1, the same
version the failing job ran, using rust-lld through the `cc` shim. The
archive is the real one this driver builds for linux/x64, and the link is
the documented cargo recipe through the two-crate workspace.

| | result |
| --- | --- |
| the archive as it shipped | `rust-lld: error: undefined symbol: __start___modules` |
| `__modules` sh_flags `0x3` -> `0x200003`, nothing else changed | links, runs, prints `ABI_FINGERPRINT=a7805641d1a3b64f` |

Archive member order is unchanged, which matters here because the merge
order decides `.init_array` order and getting it wrong segfaults on exit.

## Why no test caught it

The stand-in archive the verifier suite builds had no section whose only
references are encapsulation symbols, so the consumer link had nothing to
collect. The fixture now carries that shape, which means
`test_build_link_consumer.py` would have gone red on x64 CI before any
release ran. Taking the one `retain_sections.py` call out of the fixture
builder is the mutation: red on x64, still green on arm64, exactly how
the release behaved.

`test_retain_sections.py` covers the flag-setter against ELF objects
synthesised in Python, so it needs no toolchain, plus the wiring: the
script reaches the build context, the images install a python to run it,
staging calls it before `ar`, and a failure degrades to shared-only
rather than shipping an archive that links on one linker and not the
other.

## Folded in, found on the way

Two things I found while measuring, both recorded here rather than as new
issues:

- **musl/x64 and musl/arm64 have never had the documented recipe run
  against them.** `verify_archive.sh` only runs the consumer link when the
  host can build for that target, and the release log says
  `static_certified: not link-tested (this host cannot build for
  musl/x64)`. So the claim in the issue that the recipe works on musl was
  never measured. Both musl archives ship a static half nothing has
  linked. They get the same flag from this change, which is the right
  default, but the gap in the verifier is real and is not closed here.
- **mac/arm64 ships no static archive at all**: its facts record
  `static_ok 0`, so `static_certified` is false and the manifest says so
  honestly. Worth knowing when reading the per-target table.

Closes #67


spdrman:
## Evidence

Everything on native hardware, nothing emulated. x64 results are from the NAS (`HIGARA`, native x86_64); arm64 from this Mac. Every container is pinned with `--platform`, because the default on this Mac has been `linux/amd64` and an unpinned run was silently Rosetta.

### 1. The failure, reproduced outside CI

The real linux/x64 archive, built by this driver, linked through the documented cargo recipe in `rust:1-bookworm` (rustc 1.98.1, the same version the failing job ran):

```
rust-lld: error: undefined symbol: __start___modules
>>> referenced by main.cpp:211 (nativeaot/Bootstrap/main.cpp:211)
>>> bootstrapperdll.o:(InitializeRuntime()) in archive .../libacadsharp_native_init.a
```

### 2. The fix, on that same archive

One member extracted, one flag set, member put back:

```
viprs_acadsharp__viprs_acadsharp.o: __modules sh_flags 0x3 -> 0x200003
MEMBER_ORDER=unchanged
```

Re-linked, nothing else touched:

```
ABI_VERSION=1
ABI_FINGERPRINT=a7805641d1a3b64f
LINK_RESULT=pass
```

Member order matters here because the merge order decides `.init_array` order, and getting that wrong links clean and segfaults on the way out.

### 3. The fix through the real build path

A full rebuild on the NAS with the committed driver, then the same link against what it staged:

```
/tmp/merge/viprs_acadsharp__viprs_acadsharp.o: __modules sh_flags 0x3 -> 0x200003
retained_module_sections	1
static_ok	1
```
```
ABI_FINGERPRINT=a7805641d1a3b64f
LINK_RESULT=pass
```

### 4. arm64 is unaffected

A full linux/arm64 build on this Mac, then the same recipe in an arm64 `rust:1-bookworm`. GNU ld accepts `SHF_GNU_RETAIN` and ignores it, because it was keeping the section anyway:

```
host: aarch64-unknown-linux-gnu
ABI_FINGERPRINT=a7805641d1a3b64f
LINK_RESULT=pass
```

### 5. The mutation

The test only counts if it can fail. On the NAS, in the same container, the suite with and then without the single `retain_sections.py` call in the fixture builder:

```
### with the retain call
15 passed in 3.03s

### retain call removed
FAILED test_the_smoke_script_links_and_runs_the_binary
FAILED test_the_fingerprint_comes_back_through_the_link
2 failed, 13 passed
```

Both failures carry the #67 error. On arm64 the same mutation stays green, which is the split that let this reach a release.

CI ran the same tests on its x64 runner and they pass on this branch, so the guard is live where it matters.

## Two things I got wrong

- **The mechanism in the issue.** I said the archive member carrying `__modules` was never selected. It is always selected, because `viprs_acad_abi_version` is defined in the same object. The section is in the link and then gets collected. All four options I proposed in the issue were built on that, and Fable measured all four dead: A and B give `multiple definition of minipal_xoshiro128pp_init`, C links and exits 134, D gives an undefined `RhRegisterOSModule`.
- **A control I claimed and never had.** I wrote that musl/x64 passed the recipe on the same native runner, and used the glibc-versus-musl difference to reason about the cause. The release log says `static_certified: not link-tested (this host cannot build for musl/x64)`. That comparison never ran.

spdrman:
## A second bug, found by the review panel, unrelated to #67

The four-expert panel measured something I had not: **both published musl archives fail the documented cargo recipe**, for a reason that has nothing to do with `__modules`. It is a duplicate-symbol collision.

Measured on native hardware of each architecture, rustc 1.98.1, using this repo's own `link_consumer` workspace against the exact published assets:

| archive | host | documented cargo recipe |
| --- | --- | --- |
| `acadsharp-musl-x64.tgz` | native x86_64 musl | fails, `multiple definition of __unw_get_reg` and about a dozen more |
| `acadsharp-musl-arm64.tgz` | native arm64 musl | fails the same way |
| `acadsharp-linux-x64` post-fix | native x86_64 glibc | passes, `ABI_FINGERPRINT=a7805641d1a3b64f` |

The third row is the positive control, so the harness can pass and the musl failures are real.

The collision is between `libRuntime.WorkstationGC__libunwind.cpp.o` inside the merged archive and the `self-contained/libunwind.a` that rustc links for musl targets and does not link for glibc targets. Note the linker there is GNU ld from the Alpine gcc, not lld, so #67's bug was never going to bite musl in the first place.

**The C recipe is fine.** `MANUAL.md`'s plain `cc` form links and runs against the published musl archive, and the in-container static smoke that sets `static_certified` did run and did pass. So `static_certified: true` is literally accurate: what was never measured on musl is the *cargo* recipe, because `verify_archive.sh` only runs it when the host can build for the target and neither musl cell's host could. I said in the PR body that musl ships "a static half nothing has linked", and that was wrong in a way worth correcting: something did link it, just not the thing `LINKINFO.md` calls the measured working recipe.

### Two candidate fixes, both measured, both dead

**Drop the runtime's own libunwind from the merge on musl.** Dead. The runtime references the C++ internals of its bundled copy, not just the `__unw_*` C API, so removing the member breaks the C recipe that currently works:

```
undefined reference to `libunwind::LocalAddressSpace::sThisAddressSpace'
  referenced from UnixNativeCodeManager and RhRegisterOSModule
```

**Let the consumer turn off rustc's self-contained unwinder.** Dead. On native x86_64 musl, rustc 1.98.1:

```
as documented                     FAIL  multiple definition of __unw_get_reg
-C link-self-contained=no         FAIL  breaks the build script's own link
-C link-self-contained=-unwind    FAIL
```

What is left is renaming our copy's symbols across the whole archive with `objcopy --redefine-sym`, which is mechanical but touches every unwind path in someone else's runtime and needs the full conformance suite run on musl before I would believe it. That is a larger piece of work than this PR.

### What this PR does about it

The verifier's new `__modules` check reads the archive rather than linking it, so it covers musl and every other target a host cannot link for. That closes the class of gap. It does not fix this particular collision, and I am not going to pretend otherwise.

The decision on what musl should ship in the meantime goes in a separate comment once the rest of the panel reports.

spdrman:
## Four-expert adversarial review: outcome

Four reviewers, one each on linkers and ELF, the build driver and its shell, test quality, and release and documentation risk. Two blockers, and the panel disagreed with itself on one point in a way that turned out to be the most useful thing in the review.

### The disagreement worth having

The linker reviewer and the build reviewer reached opposite conclusions about whether `SHF_GNU_RETAIN` needs `EI_OSABI` set to `ELFOSABI_GNU`. One measured lld 13 through 22 honouring the flag on an `ELFOSABI_NONE` object and concluded the byte was unnecessary. The other measured binutils reading it as retain only on a GNU-ABI object.

Both were right about what they measured, and neither claim was safe to inherit, so I ran it myself:

```
plain    osabi=0  WA      readelf does not see a retain flag
flagged  osabi=0  WAo     an unnamed OS-specific bit
gnuabi   osabi=3  WAR     retain
```

So the flag alone fixes lld, which is the bug in #67, and binutils does not recognise it until the byte is set. I could not reproduce a plain GNU ld actually collecting the section either way, so what the byte buys is recognition rather than a fix for a failure I have seen. It costs nothing and it is what GAS stamps for any section it assembles with `R`, so it goes in, described as recognition rather than as a second fix.

The lld-only measurement would have shipped without it. That is what the fourth reviewer was for.

### Blockers, both fixed

**The docs #67 asked for did not exist.** Zero markdown files mentioned `--gc-sections`, `__modules` or `start-stop-gc`, and the archive ships four of these to people who will hit the error with nowhere to look it up. `LINKINFO.md`, which ships inside the archive, now carries the mechanism, the exact error text, that a consumer needs no flag, the workaround for an older archive, and which linkers this covers, gold included (it never implemented the flag, and I am calling that untested rather than known-good).

**Both musl archives fail the documented cargo recipe**, for an unrelated duplicate-symbol collision. Its own comment above. Recorded in `LINKINFO.md` rather than papered over.

### Everything else the panel turned up, all folded in

- **Two more sections are exposed the same way.** The bootstrapper references six encapsulation symbols, not two: `__managedcode` and `__unbox` survive only because ILC emits each as one monolithic section, so any live symbol keeps the whole thing (59471 relocations reach one, 1259 the other, zero reach `__modules`). That is a layout accident. All three now get the flag, measured to produce an identical binary size and a byte-identical `.init_array`.
- **The exact-count check was wrong.** `__modules` is an encapsulation array and N contributors is its designed shape, so `-ne 1` would redden a release for an archive that links perfectly. The floor is what matters.
- **A one-token `|| true` on the retain call re-shipped the exact broken archive and kept every test green.** So did pointing the loop at an empty directory. The wiring test asserted a substring within 400 characters and could not tell a variable that stops the build from one nothing reads. It asserts the branch now, and both mutations are verified red.
- **The synthetic ELF had zero `sh_offset`, `sh_size` and `sh_addralign`**, so "nothing else changed" could not see a setter that clobbered them. Realistic values now, and the test counts differing bytes: exactly two.
- **No decoy section**, so a setter matching by prefix or substring passed everything. `.rela__modules` sits beside `__modules` in the real object; both decoys are in the fixture now.
- **The inert-initialiser fixture had quietly stopped being one.** gcc folds `0 * (count)`, so that object had no undefined `__start___modules` at all and was a fixture for a different bug. The count goes through a volatile.
- **A retain failure degraded to a green shared-only release.** Fixed by refusing a static target that does not certify.
- `/work/retain_sections.py` hardcoded where everything else uses `$WORK/`, which is the variable the mac host build overrides.
- `struct.error` now caught, so a truncated object gets a diagnosis rather than a traceback. The `--check` usage line no longer prints the blank line under `Usage:`.
- `--target $GITHUB_SHA` on `gh release create`, because without it a dispatch from any ref builds that ref and tags the default branch: every archive right, the tag pointing at something that never produced them.
- `LINKINFO.md` still said `wire_version` was 1 in the field table and both example manifests. The header has said 2 since wire v2.

### Corrections to my own account

The issue title says linux/x64. It is lld versus GNU ld: on arm64 with `-fuse-ld=lld` the unpatched object fails identically. And `static_ok 0` on mac is the configuration, not a finding: `STATIC_TARGETS` lists four Linux RIDs and the mac cell has never attempted a static build.

PR #69, opened 2026-09-13
ABI v2: the rename, a distinct buffer code, and a terminal latch in the decode

### LINKED ISSUE #59: Outstanding findings from the EPIC G adversarial review, verified against main

spdrman:
ABI v2, and the three things that had to move together because each one alone costs a fingerprint and a rebuild of every archive.

This is **PR A1** from the triage on #59, and the last of the wave, so it carries the keyword.

Closes #59

---

## Rebased onto `4506c28`, and the thing that was reddening conformance is fixed

I was last in the sequence. `lane/59a1-abiv2` now sits on `4506c28` with #72 (driver), #73 (musl), #71 (geometry) and #70 (encoder) under it. Two conflicts, both small: `AcadSharpSource.cs`, where #71's `_viewsWithoutExtents` and my `volatile bool _closed` are adjacent declarations and both belong, and the two capture files, which I regenerated rather than hand-merged.

**#72 solved the archive smoke properly, and better than the two-line patch I proposed.** `archive_smoke_source()` no longer carries the capabilities symbol as a literal at all: `capabilities_entry_point()` takes it off the header, matched on "capabilities" so either spelling of the rename resolves, and refuses zero or two matches rather than guessing. The same literal was in the verifier's probe, where after the rename it would have failed to *compile* rather than to resolve, and that reaches the compiler as `-DVIPRS_CAPS_CALL` read from the shipped entry-point list. The `dlsym`-names-a-declared-symbol guard I asked for is in, with a test that generates the smoke from a renamed list.

Against my rebased header, with nothing changed in the driver:

```
capabilities_entry_point() -> viprs_acad_get_capabilities_v1
dlsym targets: ['viprs_acad_abi_version', 'viprs_acad_abi_fingerprint',
                'viprs_acad_get_capabilities_v1']
```

and the generated smoke run against a shim published from this branch:

```
ABI_FINGERPRINT=0502ac0f61611530
BACKING_VERSION=3.7.1
DWG_VERSION_MIN=1014
DWG_VERSION_MAX=1032
archive smoke exit=0
```

**#70 landed the truncation behaviour my `ABI.md` clause describes**, so that clause now documents code that is in the tree rather than code that is coming.

## 1. The names

`viprs_acad_capabilities_v1` named a struct and a function at once. In C a typedef name and a function name are the same kind of identifier, so the header could not typedef that struct, and a rule with one exception is a rule somebody gets wrong, so it could not typedef any of them. Every C and C++ consumer has been writing `struct viprs_acad_limits_v1` for as long as the file has existed.

| was | is |
| --- | --- |
| `viprs_acad_capabilities_v1()` | `viprs_acad_get_capabilities_v1()` |
| `viprs_acad_view_info_v1()` | `viprs_acad_get_view_info_v1()` |
| `struct viprs_view_info_v1` | `struct viprs_acad_view_info_v1` |
| `viprs_cad_handle` | `viprs_acad_handle` |
| `viprs_decode_handle` | `viprs_acad_decode_handle` |

**No typedefs are added even so.** The generator in `tests/conformance/rust/build.rs` reads every `typedef struct X X;` as an opaque handle, because that is what the pattern means everywhere else in the header, so typedefing a struct with a body emits `pub struct viprs_acad_limits_v1` twice and the crate stops compiling. The tags-only rule keeps its assertion and the stated reason is rewritten, since the old one no longer holds.

`test_no_struct_shares_a_name_with_a_call` got stronger rather than being deleted: it used to allow a tag to share a name with a call as long as nobody typedef'd it, and now it refuses the sharing outright. A new one refuses any type on the boundary without the `viprs_acad_` prefix.

Red, before the header moved: 26 failures across `test_abi_header.py`, `test_abi_layout.py` and `test_build_manifests.py`. Green: the shim's own export table, which is the thing a consumer resolves against.

```
T viprs_acad_get_capabilities_v1@@V1.0
T viprs_acad_get_view_info_v1@@V1.0
```

and no `viprs_acad_capabilities_v1` at all. The C consumer builds with `-Wall -Wextra -Werror`, so a half-done rename could not have got this far.

## 2. `VIPRS_ACAD_BUFFER_TOO_SMALL`, 10

`decode_next_batch` reported a breached bound and a caller's buffer being too small with one number, told apart by whether `*written` came back larger than the cap that went in, which no document ever stated. A consumer holding one number for both either retried a decode that was over and got nothing, or gave up on a buffer it could have grown.

`VIPRS_ACAD_LIMIT_EXCEEDED` now means a bound in `viprs_acad_limits_v1` and nothing else, and it is terminal for a decode. `VIPRS_ACAD_BUFFER_TOO_SMALL` means nothing was written, the size is in `*required` or `*written`, and the call may be retried. The capabilities and view_info sizing calls move onto it too, through `Utf8Out.Write`.

Eight "a buffer too small is LIMIT_EXCEEDED" checks flip across the two consumers, and each one gains a second assertion that the code is specifically **not** the limits code, so the pair cannot be satisfied by whichever number happens to come back.

The mutation the triage named, run: leave `Utf8Out.Write` on `LimitExceeded`, rebuild, run both consumers.

```
=== C consumer under the mutation ===
FAIL  a buffer too small is BUFFER_TOO_SMALL, not LIMIT_EXCEEDED
FAIL  and specifically not the code a bound in the limits struct reports
178 checks, 2 failures
=== Rust consumer under the mutation ===
FAIL  a buffer too small is BUFFER_TOO_SMALL, not LIMIT_EXCEEDED
FAIL  and specifically not the code a bound in the limits struct reports
157 checks, 2 failures
```

Restored: 178 checks 0 failures, 157 checks 0 failures.

## 3. The terminal latch, which is the silent truncation

`DecodeSession.Stream()` is an iterator, and a C# iterator whose body threw is finished: every later `MoveNext` returns false. `BatchWriter` reads that as "no more records", frames an empty batch carrying the last-batch flag, reports `done` 1 and returns `OK`. So the grow-and-retry the header documents after a `max_entities` breach handed the caller a well-framed, complete-looking stream with every record after the bound missing, and no code to look at.

Nothing could see it, because every scenario in the corpus stopped at the first non-OK result. So the harness now calls `NextBatch` once more after whatever happened, and records `code_after_refusal`, `written_after_refusal` and `done_after_refusal`.

**Red, re-run on the rebased tree.** The count is the same as it was before the rebase, with #70's encoder and #71's flattener underneath: six of the eight limit refusals. The recorder in, the latch out:

| scenario | decode_code | after | written | done |
| --- | --- | --- | ---: | ---: |
| `limits/max_entities_1` | LIMIT_EXCEEDED | **OK** | **12** | **1** |
| `limits/max_polyline_2048` | LIMIT_EXCEEDED | **OK** | **12** | **1** |
| `limits/bulged_polyline_past_the_bound` | LIMIT_EXCEEDED | **OK** | **12** | **1** |
| `limits/max_block_depth_5` | LIMIT_EXCEEDED | **OK** | **12** | **1** |
| `limits/spline_points_4096` | LIMIT_EXCEEDED | **OK** | **12** | **1** |
| `criticals/dimension_chain_deep` | LIMIT_EXCEEDED | **OK** | **12** | **1** |
| `limits/max_string_4096` | LIMIT_EXCEEDED | LIMIT_EXCEEDED | 0 | 0 |
| `limits/max_output_2048` | LIMIT_EXCEEDED | LIMIT_EXCEEDED | 0 | 0 |

Six of the eight. Twelve bytes is a batch header, the flag in it is the last-batch flag, and `OK` with `done` 1 is a caller being told the drawing is complete. The two that already repeated are the two that do not come out of the iterator: `max_output_bytes` is compared on every call, and the string bound throws from the encoder rather than from the walk, so the producer behind them was still alive. That is the point: whether a caller got a truncated drawing or a refusal depended on which bound it hit.

`cancel/after_one_batch` sits outside that table and repeats `CANCELED` either way, because the harness leaves its own flag up. The sharp version of that case, a caller that clears the flag and asks again, is in both consumers instead, where the corpus cannot reach it.

**Green.** Same scenarios, latch in:

| scenario | decode_code | after | written | done |
| --- | --- | --- | ---: | ---: |
| all eight above | LIMIT_EXCEEDED | LIMIT_EXCEEDED | 0 | 0 |
| `cancel/after_one_batch` | CANCELED | CANCELED | 0 | 0 |
| control, a clean decode | OK | OK | 12 | 1 |

The control matters as much as the rest. A finished stream still answers a further call with an empty last batch, so a caller's loop still terminates instead of failing, which is a rule the header states and the latch must not break.

`DecodeSession` keeps a `_terminal` code, set from the `catch (AbiException)`, from the catch-all that becomes `INTERNAL_ERROR`, from a `LIMIT_EXCEEDED` the writer returns rather than throws, and when cancel fires. Once set, every later call returns it, writes nothing and reports `done` 0. `BUFFER_TOO_SMALL` is deliberately excluded, because latching it would make the documented retry impossible, and so is the `INVALID_ARGUMENT` a null pointer earns, which never reaches the decode.

Both consumers get a case that drains a decode under `max_entities` 1 and then asks eight more times, and the cancel case now clears its own flag and asks again, which is the shape the corpus cannot reach:

```
--- cancellation
ok    a cancel flag set before the first batch is CANCELED, not a batch
ok    and a later call is still CANCELED
ok    writing nothing and reporting done 0, so a loop cannot read it as the end
--- a refusal is terminal
ok    max_entities at one refuses with LIMIT_EXCEEDED before the stream ends
ok    and does not claim the stream finished
ok    and every later call returns the same code
ok    and writes nothing, so there is no batch to parse
ok    and never reports done, which is what stopped a truncated stream looking complete
```

## The `max_string_bytes` exception, handed over from the encoder lane

That lane (#70) made an over-long `Warning` message shorten and mark rather than refuse, and it did not touch `ABI.md` because this wave gives that file to me. So the sentence after the limits table gains its exception, along with the half that keeps it narrow: a `Text` record and a view's name are still refused under the same bound, because a consumer cannot tell a label the file carries from one that was cut. The paragraph keeps "never a silently truncated stream" and says why that still holds, since a message that says it was shortened is the opposite of a silent one.

Three checks pin it, including that one, because a document that states only the exception reads as if every string is now shortened.

**This describes what #70 lands rather than what this branch does on its own**, so the two want merging in that order, or `ABI.md` is briefly ahead of the shim. Nothing in my own corpus contradicts it either way: `limits/max_string_4096` refuses a `Text` record on `g13_long_text.dwg`, and that stays true under #70.

## Folded in, all small

- `_closed` in `AcadSharpSource` is `volatile`. The close-during-decode race is reachable and benign, every path is managed so the worst outcome was `INTERNAL_ERROR` and never a fault, and volatile turns most of the window into the documented code instead.
- `ABI.md` said the cancel read "is a plain load" and the code has always done a `Volatile.Read`. The document now says volatile is part of the contract, and why: an implementation allowed to hoist it out of a loop turns a cancel into something that arrives eventually or not at all.
- `ABI.md` now says the fingerprint is over the published header **file**, byte for byte, so a comment-only edit moves it and that is intended. Canonicalising first would need the same whitespace-and-comment algorithm implemented independently in five places, and drift between any two of them is a false `ABI_MISMATCH` on a pair that works.
- The generated consumer builds without a warning. Two of the three were mine; the third poisoned an out parameter nothing asserted on.

## Re-recording

Regenerated against the rebased tree rather than hand-merged, because three PRs moved these two files under me and a rollup assembled by hand would be a recording of no build in particular.

`MANIFEST.json` carries the digests of the sources this touches. `g13_scenarios.json` gains the three new fields on all 27 decoding scenarios, two more than before the rebase because #70 added the long-warning pair, and **nothing else moves at all**: not one code, not one timing. The expectations stage reran and rewrote none of the 32 dumps. That is the control I wanted across a rebase this wide, since the latch changes what happens after a refusal and not the refusal itself, and not one record stream either.

Not re-recorded: `tests/benchmarks/amplification.json`. It carries no shim digest, nothing pins it to one, the harness's extra call happens after the allocation and RSS measurements are taken, and re-recording timings on a host that has been running containers all afternoon would replace real numbers with worse ones.

## The prose sweep

Twelve sites across eight files still named the call by its old name, all present-tense statements about the entry point rather than about the struct, so after this merge every one of them would name a symbol that does not exist. Two are messages somebody reads when something has already gone wrong: the driver's refusal of an archive whose library cannot say what it is, and the csproj's refusal of a malformed `VERSION`. One is not prose at all: `verify_archive.sh` falls back to a literal when the shipped header declares no capabilities call, and since `fail` records the refusal without exiting, that line really runs.

Left alone deliberately: every `struct viprs_acad_capabilities_v1`, which kept its name, and the comments that explain the rename by naming what it renamed. Those are history, and reading them as staleness would delete the reason.

## What was run

Everything in a container, arm64 explicitly, after the rebase.

The first version of this list said "1718 passed, 122 skipped" and was worth very little. My image had no C compiler, so every test that builds a stand-in archive skipped, and CI failed seven of them. A skip and a pass are the same character in a summary line. With gcc in the image: **14 skips instead of 149**, and the comment thread has what was hiding.

- Whole-repo `pytest`: **2140 passed, 10 skipped, 0 failures**, on Python 3.9 and 3.12.
- `ruff check`, `ruff format --check`, `tools/shellcheck-all.sh`: clean.
- C conformance consumer, `gcc -std=c11 -Wall -Wextra -Werror`, against a fresh AbiTest publish: **178 checks, 0 failures**, reporting `abi=2 wire=2`.
- Generated Rust consumer, `cargo test` plus the binary: 4 layout tests, **157 checks, 0 failures**, no warnings.
- The driver's own archive smoke, generated from this header and run against that shim: **exit 0**.
- The shim's export table carries `viprs_acad_get_capabilities_v1` and `viprs_acad_get_view_info_v1`, and no `viprs_acad_capabilities_v1`.

And the `acadsharp conformance` workflow, which was the one that could not pass before the rebase, is green. It builds a real archive, verifies it, unpacks it the way a consumer would, and runs both consumers twice, once against the archive and once against the test configuration:

```
Run the C consumer against the unpacked archive      backing version="3.7.1" abi=2 wire=2 dwg=1014..1032
                                                     --- a refusal is terminal
                                                     166 checks, 0 failures
Run the generated Rust consumer ...                  145 checks, 0 failures
Run the C consumer against the test configuration    178 checks, 0 failures
Run the generated Rust consumer ...                  157 checks, 0 failures
```

Two more things I checked by compiling rather than by reading, since the bug this rename fixes was "compiles as C++, does not compile as C": the header passes `gcc -std=c11 -Wall -Wextra -Werror -fsyntax-only` and `g++ -std=c++17 -Wall -Wextra -Werror -fsyntax-only` against a probe that names all three renamed types. I did not add that as a test, because the pytest suite has no toolchain and putting it in a conformance runner is somebody else's file.

The recorded header digest and fingerprint move in the same commit as the header, which is the rule `test_abi_fingerprint.py` states.


spdrman:
## Found on the way: `archive_smoke_source()` hardcodes the symbol it resolves

**File:** `acadsharp/build_acadsharp.py`, the generated C string in `archive_smoke_source()`, around line 929.
**Owner:** not this lane, which is why it is a comment rather than a commit.
**Effect:** `acadsharp conformance` cannot go green on this branch, and no archive can be built at all.

The function is half generated. It takes the entry-point list from the header, so all eleven names follow the rename on their own, and then it hardcodes the capabilities symbol and its full C signature as a string. I generated the smoke from this branch's header, compiled it against the shim this branch builds, and ran it:

```
entry points the generator took from the header:
   viprs_acad_abi_version, viprs_acad_abi_fingerprint, viprs_acad_get_capabilities_v1, ...
smoke: viprs_acad_capabilities_v1 did not resolve
ABI_VERSION=2
ABI_FINGERPRINT=0502ac0f61611530
smoke exit=7
```

Exit 7 becomes `fact shared_smoke_ok 0`, which `build_acadsharp.py` turns into a `RuntimeError`. That is the first step of the conformance workflow, and here it is on a real run of this branch:

```
[linux/arm64] ---- shared smoke ----
[linux/arm64] smoke: viprs_acad_capabilities_v1 did not resolve
[linux/arm64] shared_smoke_ok	0
  FAILED  linux/arm64  (the ABI smoke failed against the staged library: the library
  reports an empty backing version, so viprs_acad_capabilities_v1 cannot answer what
  it is. ... it is not shippable.)
```

### The exact change

```diff
--- a/acadsharp/build_acadsharp.py
+++ b/acadsharp/build_acadsharp.py
@@ around line 929, inside archive_smoke_source()
 \tuint32_t (*caps_of)(struct viprs_acad_capabilities_v1 *, uint8_t *, uint64_t,
 \t\tuint64_t *) = (uint32_t (*)(struct viprs_acad_capabilities_v1 *, uint8_t *,
-\t\tuint64_t, uint64_t *))dlsym(h, "viprs_acad_capabilities_v1");
+\t\tuint64_t, uint64_t *))dlsym(h, "viprs_acad_get_capabilities_v1");
 \tif (!caps_of) {
-\t\tfprintf(stderr, "smoke: viprs_acad_capabilities_v1 did not resolve\\n");
+\t\tfprintf(stderr, "smoke: viprs_acad_get_capabilities_v1 did not resolve\\n");
 \t\treturn 7;
 \t}
```

`struct viprs_acad_capabilities_v1` on the two lines above is correct as it stands and must not be touched: only the call was renamed, the struct keeps its name. With just that string changed, the same smoke against the same library:

```
ABI_VERSION=2
ABI_FINGERPRINT=0502ac0f61611530
BACKING_VERSION=3.7.1
smoke exit=0
```

### Why nothing caught it

`test_build_manifests.py::TestTheShippedLibraryHasToSayWhatItIs::test_the_smoke_asks_for_the_backing_version` asserts `"viprs_acad_capabilities_v1" in source`, and that still passes, because the generated source names the **struct** as well as the call. It was never checking the symbol the smoke actually resolves, which is why a rename could reach CI through it.

The check that would catch it is "every string this smoke hands to `dlsym` is a name the header declares". **I have not added it**, deliberately: it would be red for exactly the same reason, and a red test in a branch I am handing over as done is worse than a comment. Whoever makes the driver change should add it in the same PR, because without it the next rename walks into this again.

### While it is open: prose still naming the old call

All comments, none load-bearing, none of them in files this lane owns:

- `acadsharp/build_acadsharp.py` 846, 1669, 1725
- `acadsharp/native/Viprs.ACadSharp.Native.csproj` 107, 121, 137
- `acadsharp/tests/test_version.py` 130
- `acadsharp/tests/test_release_workflow.py` 904
- `.github/workflows/release-acadsharp.yml` 187

I fixed the two that are mine (`tests/conformance/rust/build.rs`, `tests/conformance/c/run.sh`) and left these. A comment naming a symbol that no longer exists sends the next reader looking for it, so they are worth one sweep rather than nine separate ones.

spdrman:
## The `archive_smoke_source()` finding above is closed by #72, and closed better

Updating my own comment rather than leaving it standing as an open ask.

#72 did not take the two-line patch I proposed, and it was right not to. The literal is gone entirely: `capabilities_entry_point()` reads the name off the shipped header, matched on `capabilities` so either spelling of the rename resolves, and refuses zero matches or two rather than guessing between them. It also found the copy I had missed, in `verify_archive.sh`'s probe, where after the rename the literal would have failed to **compile** rather than to resolve, which is a worse failure to debug because it says nothing about a symbol. That one reaches the compiler as `-DVIPRS_CAPS_CALL` taken from the same list.

And the guard I said should exist but would not add, because it would have been red here, is in: `test_every_symbol_it_resolves_is_declared_by_the_header`, with a test that generates the smoke from a renamed entry-point list so the guard has been seen to fire.

Verified against my rebased header with nothing changed in the driver:

```
capabilities_entry_point() -> viprs_acad_get_capabilities_v1
dlsym targets: ['viprs_acad_abi_version', 'viprs_acad_abi_fingerprint',
                'viprs_acad_get_capabilities_v1']
```

and the generated smoke, compiled and run against a shim published from this branch:

```
ABI_FINGERPRINT=0502ac0f61611530
BACKING_VERSION=3.7.1
DWG_VERSION_MIN=1014
DWG_VERSION_MAX=1032
archive smoke exit=0
```

### The prose half, now done

Of the nine sites I listed, #72 swept none of the ones that mattered and added a few of its own, so I have done the sweep here: twelve sites across eight files, every one a present-tense statement about the entry point rather than about the struct. Two of them are messages a person reads when something has already gone wrong, and one is not prose at all. `verify_archive.sh` falls back to a literal when the shipped header declares no capabilities call, and because `fail` records the refusal without exiting, that line really runs, so the name it carried could never have resolved again.

What I left: every `struct viprs_acad_capabilities_v1`, which kept its name through all of this, and the handful of comments that explain the rename by naming what it renamed. Reading those as staleness would delete the reason.

spdrman:
## A green local run with 122 skips in it, and what was hiding behind them

Recording this because the way I found it is worth more than the fix.

After the rebase I ran the suite in my container and got `1718 passed, 122 skipped`, called it green, and pushed. CI then failed `test (3.12)` with seven tests in `test_verify_archive.py` and `test_privatise_unwind.py` that had skipped locally, because my image had no C compiler and those tests build a stand-in archive and run `verify_archive.sh` against it. A skip and a pass look identical in a summary line, and 122 of them looked like nothing at all.

I rebuilt the image with gcc. Same tree, same command: **14 skips instead of 149, and 2140 tests that actually execute**. Everything below was run in that image.

### What was hiding

`test_verify_archive.py` builds a stub library and special-cases the capabilities call by name in four places, because that one needs a real body rather than `return 0u`: since #72 the verifier asks the library what AC10xx range it reads and holds `LINKINFO.json` to the answer, so a stub returning zero would make that check pass over anything.

All four were literals. After the rename none of them matched, so the generic arm of the loop emitted

```c
uint32_t viprs_acad_get_capabilities_v1(void) { return 0u; }
```

under the header's new name, the probe resolved *that*, and the library answered a read range of 0 to 0:

```
FAIL: the library does not read the AC10xx range LINKINFO.json states.
    probe: LINKINFO.json dwg_version_min 1014 and dwg_version_max 1032, library answers 0 and 0
    ABI_VERSION=1 ABI_FINGERPRINT=0502ac0f61611530
    DWG_VERSION_MIN=0 DWG_VERSION_MAX=0
```

Seven tests red, and the message blames a manifest that was correct. This is the same shape as the driver problem #72 fixed, one layer further out: the driver and `verify_archive.sh` both take the name off the header now, and the fixture that exercises the verifier did not.

So it asks `capabilities_entry_point()`, which is the single source of truth the other two already read.

### And a second one in the same fixture

`CAPABILITIES_SOURCE` answered `caps->abi_version = 1u` to a header declaring `2u`. That is exactly the drift the verifier's version check exists to catch, reported against a fixture rather than against a build, and it would have fired the moment the first line was fixed. It reads `VIPRS_ACAD_ABI_VERSION` and `VIPRS_ACAD_WIRE_VERSION` from the header it already includes, so there are no digits left in it to go stale.

### Where this leaves the count

Four hardcoded spellings of this one call have now been found in three separate rounds: the archive smoke and the verifier probe (#72), and the verifier's own fixture (here). Each was invisible to the layer above it. What closes the class rather than the instances is #72's `test_every_symbol_it_resolves_is_declared_by_the_header`, and there is no equivalent for a fixture, because a fixture is allowed to lie about the ABI on purpose. The honest version of this note is that I do not think the class is closed, and the next rename should grep for the literal before trusting a green run.

PR #70, opened 2026-09-13
The encoder: a warning that does not fit, the ceilings a uint32 count has, and what DocumentEnd counts

spdrman:
Refs #59

The **PR C** section of the triage: the encoder and `WIRE.md`. Three items, none of
them touching the header, so nothing here moves the ABI fingerprint and no archive
has to be rebuilt.

## 1. An over-long string killed the whole decode

`max_string_bytes` is a bound on one string, and breaching it used to end the
decode wherever the string came from. For a drawing's own text that is right. For a
`Warning` it is a bad trade: that message is a sentence this library wrote, and it
usually carries a name the file chose the length of. The reader's `Unlisted object
with DXF name ...` notification is the file's own class name, so a drawing can
decide how long our sentence about it is and take every record after it down.

`RecordEncoder.Utf8WithLength` now takes a `truncate` argument, true for
`TypeWarning` and nothing else. The cut lands on a UTF-8 character boundary and
ends with the twelve bytes ` [truncated]` when those twelve fit inside the bound.
`Text` and the `ViewBegin` name still refuse, because a consumer cannot tell a
label the file carries from one we shortened.

`AcadSharpSource.OnNotification` stores through the same helper. Notifications
accumulate during the open, before a record has been asked for and before any
bound has been consulted, so a 100 KB class name would otherwise sit in that list
for the life of the document.

### Red

```
FAILED TestTheBoundNoLongerEndsTheDecode::test_the_decode_completes
  an 8192-byte reference path still ends the decode at max_string_bytes 4096.
  The message is this library's own sentence about the drawing, and refusing it
  costs every record after it. a string in this record is longer than
  max_string_bytes allows
```

and the run behind it, recorded before anything under `native/` was touched:

```
warnings/long_warning_truncated: open=OK decode=LIMIT_EXCEEDED
warnings/long_warning_truncated: run=0 markers=0 dangling=0
warnings/long_warning_fits:      open=OK decode=OK
warnings/long_warning_fits:      run=4086 markers=0 dangling=0
```

`run=0` on the first is the whole point: the decode produced no bytes at all, so
the drawing's geometry went down with the message.

### Green

```
warnings/long_warning_truncated: open=OK decode=OK
warnings/long_warning_truncated: run=2024 markers=1 dangling=0
warnings/long_warning_fits:      open=OK decode=OK
warnings/long_warning_fits:      run=4086 markers=0 dangling=0
limits/max_string_4096:          open=OK decode=LIMIT_EXCEEDED
```

The last line is the control I was asked to keep: a `Text` record the drawing owns
is still refused under the same bound, on `g13_long_text.dwg`.

Reading the truncated stream by hand:

```
message_len field 4095
external reference ../not-resolved/ééééé ... éééééééé [truncated]
longest run 2024 ends at 4863
after run: b' [truncated]\x00\x00\x7f\x00\x00\x18\x00\x00'
dangling 0
```

2024 filler characters is 4048 bytes; with the 35-byte head that survives and the
12-byte marker that is 4095 against a bound of 4096, and one more character would
not have fitted. The zeroes after the marker are the record's padding, so the
marker really is the end of the message.

### The fixture

`g13_xref_long.dwg`, `Corpus.WriteXref` with an 8192-byte `XRefPath`, which reaches
the wire as the `UNRESOLVED_BLOCK` message. The path is padded with U+00E9 rather
than an ASCII character, because the byte budget a 4096-byte bound leaves for it is
odd: a cut that counted bytes instead of characters would split one of those in
half and leave a lone lead byte at the end of a message the record declares as
UTF-8. `dangling=0` is that check, and it has a positive control beside it.

The scans over the raw stream are crude on purpose, the way `non_finite_exponents`
is: they count a two-byte pattern and a twelve-byte marker, they cannot be a
mis-parse of the wire, and each has a control that proves it finds what it looks
for.

### What has no fixture

The storage-time path. Producing one means a drawing that makes ACadSharp's reader
say something 100 KB long, and its own messages are short. What covers it is that
both paths cut through one helper, plus a source rule that keeps them there. I am
not claiming more than that.

## 2. `WIRE.md` stated no ceilings

Counts are `uint32`, a record's `length` is `uint32` in the frame, but the
producer's cursor is a signed 32-bit position, so no record longer than 2^31 - 1
bytes is ever emitted. The new section states the ceilings that fall out:
`point_count` at most 89,478,482 without bulges and 67,108,861 with, and the three
string lengths at 2^31 - 1 less the record's fixed prefix. A limit set above a
ceiling is a limit the wire cannot carry.

One correction to the brief, and it is in the document: the string ceilings are not
simply the subtraction. The padding to a multiple of four comes out of the same
budget, so each is that subtraction rounded down to where the padded record still
fits (2,147,483,580 / 2,147,483,572 / 2,147,483,612).

Nothing in `test_wire_ceilings.py` compares the document against a copy of the
document. The two `point_count` ceilings are recomputed from the `length` formula
`WIRE.md` states in its own `Polyline` section, and the three string ceilings from
the fixed prefixes `conformance.c` and `payload.rs` assert independently of each
other, with a check that those two agree.

`Cursor.Take` now honours it: a negative `n`, which is what `count * 8` becomes
once a vertex count is past about 268 million, is `LIMIT_EXCEEDED`, and `_pos`
advances inside a `checked` context. Before this it wrapped silently and the record
went out with a length header describing bytes nobody had written.

**I did not run that branch and I am not going to claim I did.** Observing the
failing state costs a record about 2 GiB long, which means the array behind it and
a container with room for both, for one branch. What I did instead was read the two
guards and rely on the `checked` context, which is not a comment about intent: it
is the arithmetic refusing to wrap, so an edit that gets a guard wrong faults there
rather than producing a wrong length. The test reads the same two things.

## 3. `DocumentEnd.total_records` did not say whether it counts itself

It does: every record in the stream, `DocumentBegin` and this record included, so
it is `ViewEnd`'s `record_count` plus two. `DecodeSession` computes `_items + 5`
against `_items + 3`, and both conformance consumers increment their tally before
comparing, so this is the document catching up with two oracles. The test checks
the sentence against `DecodeSession`'s own arithmetic rather than against a
number.

## Three more `WIRE.md` edits, handed over from #71

That lane owns the flattener change and not this file, so the wording comes from
its comment and the document lands here.

**Record 5 never said where angle zero is.** This is the real one: a hole rather
than drift. The record fixes the sense of its angles and not their origin, and a
consumer cannot draw an arc with one of those. Zero is along the x axis the
arbitrary axis algorithm gives for the record's own normal, and I wrote the
algorithm out rather than naming it, because this document's first paragraph
promises a consumer can be written from it alone. A normal of `(0, 0, 1)` puts
zero on world `+X` and a normal of `(0, 1, 0)` puts it on `(-1, 0, 0)`, plus the
formula for a point at angle `t`. The `1/64` threshold is stated as a real
number, with what an integer division does to it.

**The `ViewBegin` extents sentinel.** The finiteness section scoped its guarantee
to geometry for a good reason and that read as permission to put a `NaN` in the
extents. It is not: extents the source cannot give come across as the inverted
box, `min` at `1e20` and `max` at `-1e20`, which is AutoCAD's own `EXTMIN` and
`EXTMAX` for an empty drawing, so `min_x > max_x` is a comparison a consumer can
make without knowing the constants. The box does not say why, and the document
now says so.

**Warning 106.** "Not a similarity" includes a reflection, and a mirror preserves
an `Arc`, an `Ellipse` and a bulge exactly. The row now names what does raise it
and rules a reflection out by name.

`test_wire_conventions.py` runs the document's algorithm against the document's
own worked examples, so editing one and not the other fails rather than passing
quietly; neither is written out in the test. The threshold gets two controls:
with the band rounded to zero the world z normal crosses with itself and there is
no axis at all, and just off that axis the two branches come out a right angle
apart, which is the failure the band exists to prevent.

Once #71 landed there was finally a fixture with a normal that is not `+Z`, so the
algorithm is now run rather than only stated. `g13_ocs_rotated.dwg` carries an
`Arc` and a `Polyline` in one plane with one origin and a normal of `(0, 1, 0)`,
the case the document calls the one nobody guesses. The polyline's vertices are
the block's own x and y axes lifted, with no angle in them, so they fix the frame
without touching the document, and angle zero has to land on the first span and a
quarter turn on the second. The document's algorithm is also checked against
`g13_support.arbitrary_axis`, written from the same specification by the lane that
made the flattener run it, over seven normals including both sides of the band.

**That comparison found a real hole in my own wording.** `1/64` is a bound on a
direction cosine, so it means nothing on a normal that is not a unit vector, and I
had not said to normalize first: `(1/64 + 1e-9, 0, 1)` is outside the band as
written and inside it once normalized, and the two answers are ninety degrees
apart. The algorithm now opens with the normalize step and says why. Swapping the
two candidate axes in the document takes twelve assertions red, both fixture
checks included.

`WarningCodes.cs` carries row 106's sentence a second time and needed no edit in
the end: #71 replaced it in the same commit as the gate, with the same wording
plus a paragraph on why it changed. Grepped rather than assumed.

Nothing in this section moves the shim digest, so no capture was re-recorded for
it.

## What I re-recorded, and what I did not

`MANIFEST.json` and `g13_scenarios.json`, in the pinned arm64 SDK container, after
`git fetch origin && git rebase origin/main`, three times: #72 moved
`VersionGate.cs` and #71 rewrote `Flattener.cs`, both shim sources, so the digest
the captures are pinned to moved twice for reasons that have nothing to do with
this branch. The scenarios were re-run each time rather than carried across, since
they carry no shim digest and nothing in the suite would have caught a stale one.

The manifest diff against `4038a09` is three lines: the rollup and the two sources
this branch touches. Not one dump changed, the four new OCS ones included.

The `MANIFEST.json` diff is worth reading for what is not in it: three lines, the
rollup and the two sources I touched. Not one dump changed, not one `output_bytes`,
not one batch count. The truncation is reached only by a string past
`max_string_bytes`, the whole corpus records at the default 64 KiB, and the only
fixture that meets the new path is the one written for it.

The scenario file churns 652 lines because every run re-records its own timings
and RSS numbers, so I compared the outcomes rather than the diff. Across all 37
scenarios that existed before, every `open_code`, `decode_code`, `output_bytes` and
batch count is identical to `main`, and the only new entries are the two this PR
adds:

```
warnings/long_warning_fits       was None   now ('OK', 'OK', 9116 bytes, 1 batch)
warnings/long_warning_truncated  was None   now ('OK', 'OK', 4948 bytes, 1 batch)
```

`amplification.json` is not re-recorded. Nothing in it is pinned to the shim digest,
and the change adds two integer compares per write and no allocation at all on the
path that fits, so re-running it would replace measured numbers with
differently-noisy ones and call that a result.

The fixture generator builds Debug, so `RecordEncoder.Agree` runs on every record
and asserts the measuring walk and the writing walk agree on its length. The
truncating decode passes, which is that check applied to the new path.

## Files, including one I share

* `native/Wire/RecordEncoder.cs`, `native/Sources/AcadSharpSource.cs`,
  `acadsharp/docs/WIRE.md`
* the fixture, the corpus writer, the scenario recorder, `PROVENANCE.md`, and two
  new test modules

**I touched `AcadSharpSource.cs`**, only `OnNotification`, and another lane in this
wave is making `_closed` volatile in the same file. The two edits are in different
methods.

In `docs/WIRE.md` I added three sections and did not touch line 58.

## One thing I could not fold in

`docs/ABI.md` now under-describes `max_string_bytes`, and it is another lane's file
in this wave, so I left it alone rather than take a conflict.

The sentence after the limits table says "Exceeding any of them is
`VIPRS_ACAD_LIMIT_EXCEEDED`", which is no longer true for one case: a `Warning`
message past the bound is shortened and marked instead. The rest of that paragraph
still holds, including "never a silently truncated stream", because the stream is
complete and the message says it was cut. The table row itself ("the longest UTF-8
string a `Text` or `Warning` record may carry") is still exactly right.

What it wants is one clause, something like: *Exceeding any of them is
`VIPRS_ACAD_LIMIT_EXCEEDED`, with one exception: a `Warning` message past
`max_string_bytes` is shortened and marked rather than refused, because that
message is the decoder's own and refusing it would cost every record after it.
`docs/WIRE.md` has the rule.* Whoever lands the ABI.md change in this wave should
pick it up.

## CI

Full repo suite in the pinned container, arm64:

```
1791 passed, 9 skipped
```

`ruff check` and `ruff format --check` clean over 77 files, and
`tools/shellcheck-all.sh` clean.


spdrman:
Green, and rebased onto `main` at `65ec874`, which had not moved while I worked.

```
consumers    pass  2m21s
lint         pass  6s
shellcheck   pass  6s
test (3.12)  pass  1m19s
test (3.9)   pass  1m41s
```

The conformance run is the one worth naming: it built the arm64 archive from the
changed shim with NativeAOT, verified it, and ran both consumers against the
unpacked archive and against an `AbiTest` build. Nothing in the header moved, so
no other archive has to be rebuilt.

Two things I want on the record rather than buried in the body.

**I did not run the `Cursor.Take` overflow.** Observing it costs a record about
2 GiB long, the array behind it, and a container with room for both, for one
branch. I read the two guards and I am relying on the `checked` context to make
reading enough: it is the arithmetic refusing to wrap, not a comment saying it
should not. The test reads the same two things and says so in its own docstring.

**`docs/ABI.md` now under-describes `max_string_bytes`**, and it is another lane's
file in this wave so I left it alone. The sentence after the limits table says
"Exceeding any of them is `VIPRS_ACAD_LIMIT_EXCEEDED`", which is no longer true
for a `Warning` message: that is shortened and marked instead. The rest of the
paragraph still holds, "never a silently truncated stream" included, because the
stream is complete and the message says it was cut, and the table row itself is
still exactly right. The body has the clause I would add. Whoever lands the
ABI.md change in this wave should pick it up.

spdrman:
## The empty-view warning code: it has to wait for the code, and the repository already says so

#71 suggests an empty view wants a warning code of its own, so a consumer can
tell an empty drawing from a damaged one. I agree it belongs and I am not adding
the row, because a row now goes red:

`tests/test_warning_codes.py::test_the_document_defines_no_code_nothing_emits`

> WIRE.md defines {extra}, which nothing in this library produces. A specified
> code with no producer is a branch a consumer writes and never reaches, and it
> stops anyone noticing when the real one is missing.

That is the repository's own rule, with a test enforcing it in both directions:
every code the shim emits is documented, and every code documented is emitted.
So the row and the `WarningCodes.cs` constant have to land in the same change,
and that change is the flattener's, not this file's.

What I have done instead is put the gap in the document as a fact rather than a
plan, in the finiteness section:

> The box does not say why. A view that is empty and a view whose extents the
> drawing has damaged report the same box, and nothing else on this wire tells
> them apart.

So a consumer is told the limit of what the sentinel means, and whoever adds the
code deletes that last clause and adds a row in the same commit. Carrying it
forward rather than losing it.

## One twin I deliberately left stale, and why

`native/Adapter/WarningCodes.cs` carries row 106's sentence a second time, as the
C# spelling of the same table:

> An INSERT scale that is not a similarity, under which a circle is an ellipse
> and a bulge is an elliptical arc...

I own that file and I did not touch it. On **this** branch the gate really is
`IsSimilarity`, so replacing that comment would make it wrong in the other
direction: right about #71's code and wrong about the code in the same tree. It
also moves the shim digest, which forces a re-record of `MANIFEST.json` that
#71's rebase would immediately redo, for a comment.

It wants the same edit as the table row, in whichever of the two PRs lands
second. The replacement text is row 106's, minus the table pipes.

## What I verified, and the one thing I could not

The two worked examples are checked by running the algorithm the document states
against the examples the document states, in `test_wire_conventions.py`. Both
come out of the file; neither is written in the test. `(0, 0, 1)` gives
`(1, 0, 0)` and `(0, 1, 0)` gives `(-1, 0, 0)`, and the two controls on the
threshold hold: at a band of zero the world z normal crosses with itself and
there is no axis at all, and a normal of `(1e-3, 0, 1)` comes out a right angle
apart between the two branches.

**What I could not check here is that the shim emits that convention for a
non-`+Z` normal**, because for `+Z` the algorithm is the identity and every
normal in this branch's corpus is `+Z`. The four `g13_ocs_*.dwg` fixtures in #71
are what exercise it. What I can say from this branch is weaker and still worth
saying: this branch's flattener passes an arc's angles through unchanged from the
file, which is DXF's convention by construction, so the sentence is true here as
well as there.

These three edits describe #71's behaviour, so the two want to land in the same
wave. If this one merges alone, row 106 is briefly ahead of the gate that raises
it and the extents paragraph is ahead of the sentinel. Neither is a contradiction
a consumer can trip over: the old gate raises 106 in a superset of the new one's
cases, and no view on this branch reports the box at all.

spdrman:
The empty-view warning code is now tracked in #74, with your reasoning for not adding the row here. You were right that a row before a producer goes red against this repo's own rule, so it is filed as one change (producer plus row plus a scenario) rather than as a doc edit waiting for code.

spdrman:
Green again, now rebased onto `main` at `65f7ddc`, which is #72.

```
consumers    pass  2m41s
lint         pass  6s
shellcheck   pass  7s
test (3.12)  pass  1m11s
test (3.9)   pass  1m54s
```

#72 took a second copy of `AbiConstants` out of `VersionGate.cs`, which is a shim
source, so the digest `MANIFEST.json` pins the captures to moved under this branch
for a reason that has nothing to do with it. The rebase conflicted on exactly that
block and nowhere else. I took `main`'s side and re-recorded rather than resolving
by hand, because a hand-merged rollup is a number that is a run of neither tree.

The manifest diff against the new base is the same three lines as before: the
rollup and the two sources I touched. No dump changed.

I re-ran the scenarios as well rather than carrying them across the rebase. They
carry no shim digest, so nothing in the suite would have caught a stale one, and
the `gate/*` scenarios are the ones `VersionGate` feeds. They come back identical:
across all 37 scenarios that predate this branch, every `open_code`,
`decode_code`, `output_bytes` and batch count matches `main`, and the only new
entries are the two long-warning ones.

spdrman:
Rebased onto `main` at `4038a09`. Three findings, one of them about my own document.

## The `WarningCodes.cs` twin is already done, by #71

The reason I left it no longer holds, and neither does the job. #71 replaced that
comment itself, in the same commit as the gate:

> A block transform that does not scale an entity's plane uniformly, under which
> a circle is an ellipse and a bulge is an elliptical arc. The parameters still
> cross unchanged; this says they were measured in a frame the transform does not
> preserve.
>
> Not a reflection. A mirror was raising this while the angles of an `Arc` and the
> parameters of an `Ellipse` followed nothing...

That is my row 106 plus a paragraph on why it changed, so the two halves already
say the same thing and I have not touched the file. I grepped the tree for the old
wording rather than trusting that: the only occurrences left are in `Flattener.cs`
and `test_flatten_ocs.py`, both quoting what the message *used to* say in order to
assert it no longer does, which is correct as it stands.

So the digest cost you were budgeting for did not need spending, and I did not
spend it. The manifest moved for the rebase alone.

## The non-`+Z` coverage: cheap, and it found something

Very cheap, and worth more than I expected. `g13_ocs_rotated.dwg` carries an
`Arc` and a `Polyline` in the same plane, same origin, both with a normal of
`(0, 1, 0)`, which is the branch my worked example calls the one nobody guesses.

The polyline is the witness, and that is the whole trick: its three vertices are
the block's own x and y axes lifted, with no angle anywhere in it, so it fixes the
two axes of the frame without going near my document. Angle zero then has to land
on its first span and a quarter turn on its second. Both do, through `WIRE.md`'s
own formula, and the fixture's shape is asserted first so a fixture that changed
fails rather than quietly asserting something else.

I also put the document's algorithm against `g13_support.arbitrary_axis`, which
#71 wrote from the same specification and which the flattener's own tests run.
Seven normals, including both sides of the `1/64` band.

**That is what found the bug, and the bug was mine.** The band is a bound on a
direction cosine, so it only means anything on a unit vector, and I never said to
normalize before taking it. A normal of `(1/64 + 1e-9, 0, 1)` is outside the band
as I wrote it and inside it once normalized:

```
raw   nx = 0.015625001      >= 1/64 -> True
unit  nx = 0.015623094...   >= 1/64 -> False
```

and the two branches are ninety degrees apart there. A consumer written from the
document as it stood would have drawn those arcs in the wrong plane. The algorithm
now opens with `n = normalize(nx, ny, nz)` and says why, and says to do it even
though a producer promises a unit normal, because the bytes may not have come from
this producer. `g13_support` already normalized, which is exactly why comparing
two implementations caught it and comparing my prose against my own table did not.

Swapping the two candidate axes in the document takes twelve assertions red, both
fixture-backed ones included, so none of it passes over nothing.

## The rebase

#71 rewrote `Flattener.cs` and #72 moved `VersionGate.cs`, both shim sources, so
everything was re-recorded. The manifest diff is three lines: the rollup and the
two sources this branch touches. **Not one dump changed, the four new OCS ones
included**, which is the strongest thing I can say about an encoder change landing
under a rewritten flattener. All 37 scenarios that predate this branch match
`main` on every code, byte count and batch count.

My `OnNotification` edit and #71's changes to the same file merged without a
conflict; they are in different methods.

One thing the rebase cost: the two earlier re-record commits collapsed to nothing,
because their captures were superseded wholesale, so the tree no longer carries the
red `LIMIT_EXCEEDED` capture as a commit. The red run is in the PR body, which is
where anyone would read it anyway, and the first commit is still tests-only and
still red on its own.

spdrman:
Green on `4038a09`, six checks now that #73 added the musl link test:

```
consumers          pass  2m25s
musl-static-link   pass  1m38s
lint               pass  5s
shellcheck         pass  6s
test (3.12)        pass  1m28s
test (3.9)         pass  1m47s
```

2105 tests locally in the pinned arm64 container, `ruff check` and
`ruff format --check` clean over 81 files, `tools/shellcheck-all.sh` clean.
Up to date with `main`, `mergeStateStatus` CLEAN, `closingIssuesReferences`
empty so #59 stays open for whichever PR closes it.

PR #71, opened 2026-09-13
Lift object coordinates into world space, and make an arc follow a reflection

### LINKED ISSUE #65: OCS to WCS, arcs under mirrored insertions, and the two things beside them
Wire version 2 (#56) fixed one half of two problems and left the other half alone, on purpose. This is the other half, plus the two things beside it that turned out to be the same bug wearing different clothes.

## 1. OCS to WCS is never applied

A DWG entity's coordinates are in an object coordinate system defined by its extrusion direction, and the arbitrary-axis algorithm is what turns that into world space. The shim never runs it. Polyline vertices leave `Flattener.EmitPolyline`'s successor untouched, and `Arc`, `Circle` and `Ellipse` centres and normals pass through unchanged under 3D block rotations.

That is one problem across five record kinds, and it wants its own change because the fix is a single transform applied in five places rather than five separate corrections. #56 carried the normal onto records 4 and 9 specifically so the fix has somewhere to land: a consumer cannot do the transform itself without knowing the plane, and until wire version 2 a `Polyline` did not say.

## 2. An `Arc` under a mirrored insertion still bends the wrong way

#56 fixed this for a polyline's bulges. `Flattener.BulgeArray` negates every bulge when `Basis.Mirrored`, and `g13_mirrored_bulge.dwg` pins it: the arc's midpoint comes out at `(-12.5, 5)`, the mirror image of the unmirrored `(12.5, 5)`, where it used to come out at `(-7.5, 5)` on the wrong side of the chord.

The `Arc` record has the same defect and nothing in it follows a reflection. Its `start_angle` and `end_angle` are counter-clockwise about the normal, and a mirror flips what counter-clockwise means. `Flattener.Map`'s `Arc` arm adds `basis.Rotation` to both angles and that is all, so every arc in a mirrored block is drawn as its own reflection about the chord. `Circle` is unaffected because it has no angles, and `Ellipse` has the same problem in its `start_param` and `end_param`.

There is no fixture for it yet. `g13_mirrored_bulge.dwg` is a polyline; an arc, a circle and an ellipse in the same mirrored block would make the whole set checkable at once.

## 3. `NON_UNIFORM_BLOCK_SCALE`'s message is false for a mirror

The text reads "crossed under a block transform that is not a similarity, so its parameters describe a shape the transform does not preserve". A reflection preserves every shape exactly; what it does not preserve is handedness.

#56 took polylines and polygons off that warning, because with the bulge negated the record is exact, and `Basis.IsUniform` is the gate that allows a mirror. `Arc`, `Circle` and `Ellipse` still use `Basis.IsSimilarity` and still raise it under a pure mirror, which is correct today only because their angles really are wrong. Once item 2 is fixed they should move to `IsUniform` as well, and whatever warning remains for a genuinely non-uniform scale needs wording that is true.

## 4. `ViewBegin`'s extents are not covered by the finiteness guarantee

`docs/WIRE.md` now promises that no record of type 3 to 10 carries an `f64` that is `NaN` or infinite, and `Flattener.Finite` enforces it with `WarningCodes.NonFiniteGeometry`. `ViewBegin` is deliberately outside that range: its extents are a bounding box the source reports rather than a shape anybody draws, and a view holding nothing has no finite one, so the guarantee would mean inventing a number.

That is defensible and it is also a hole. A consumer that reads `min_x` as `NaN` gets exactly the failure the guarantee exists to prevent, and the record gives it no way to tell "this view is empty" from "this drawing is damaged". The options are a flag saying the extents are absent, a documented sentinel, or a warning code of its own, and picking one is a design call rather than a fix.

## Not in scope here either

Elliptical arcs from an anisotropic scale are still a warning rather than a record, and a record for them is a separate discussion. Polyline width, thickness and linetype are not on the wire in any version.

spdrman:
Closes #65

A DWG entity does not store world coordinates. It stores them in the object
coordinate system its extrusion names, and DXF's arbitrary axis algorithm is
what turns one into the other. The shim never ran it, and on a corpus where
every extrusion was +Z that was invisible, because for +Z the algorithm is
exactly the identity.

The four items in the issue turned out to be one change plus one that stands on
its own. A placement now carries the entity's own plane composed into it, and a
basis measured through that plane hands back four things at once: the two
in-plane scale factors, the plane's normal, the angle offset an `Arc`'s angles
need, and whether the transform reverses handedness. Items 1, 2 and 3 all fall
out of that.

## The fixtures, before any of it was fixed

Four new files, recorded against the unfixed flattener and committed in
d922f7d so the diff carries the before as well as the after.

`g13_ocs_plane.dwg` is an arc, a circle and a bulged polyline in a -Z plane,
plus a circle on an oblique extrusion, with a 2D and a 3D polyline beside them.
`g13_ocs_mirror.dwg` inserts one block twice, as itself and at `XScale -1`, so
the mirrored records are the unmirrored ones with x negated and the fixture
carries its own expected values rather than numbers I typed. `g13_ocs_rotated.dwg`
has an insertion whose own extrusion is +Y, which takes the block's plane to the
world's XZ plane. `g13_ocs_skew.dwg` composes a 3x scale with a 45 degree
rotation.

What they recorded before the fix:

```
# g13_ocs_plane: every centre is the coordinate the file holds
Arc      c=[(4.000000,3.000000,0.000000)]  normal=[(0.000000,0.000000,-1.000000)]
Circle   c=[(6.000000,-2.000000,1.000000)] normal=[(0.000000,0.000000,-1.000000)]
Circle   c=[(5.000000,0.000000,7.000000)]  normal=[(1.000000,2.000000,2.000000)]

# g13_ocs_mirror: the mirrored arc sweeps the far side of its circle, and all
# three curves raise a warning for a transform that preserves them exactly
Arc      c=[(10.000000,0.000000,0.000000)]  a0=0.000000 a1=1.570796
Warning  NON_UNIFORM_BLOCK_SCALE "ARC crossed under a block transform that is not a similarity..."
Arc      c=[(-10.000000,0.000000,0.000000)] a0=3.141593 a1=4.712389
Ellipse  major=[(-5.000000,0.000000,0.000000)] p0=0.000000 p1=1.570796

# g13_ocs_rotated: every record claims the XY plane
Circle   c=[(-3.000000,0.000000,0.000000)] normal=[(0.000000,0.000000,1.000000)]
Arc      c=[(0.000000,0.000000,0.000000)] a0=3.141593 a1=4.712389 normal=[(0.000000,0.000000,1.000000)]

# g13_ocs_skew: no warning at all, and a circle that is an ellipse
Circle   c=[(0.000000,0.000000,0.000000)] r=8.944272 normal=[(0.000000,0.000000,1.000000)]
```

## Red

```
$ python3 -m pytest tests/test_flatten_ocs.py -q
...
19 failed, 6 passed in 0.13s
```

The six that passed are the controls and they had to pass either way: the four
that pin the arbitrary axis helper against cases worked out by hand, the circle
under the mirror (a circle has no angles, so nothing in it follows a reflection
and nothing needed to), and the length of the oblique centre, which the lift
preserves because the lift is a rotation.

## Green

```
$ python3 -m pytest tests/test_flatten_ocs.py -q
26 passed in 0.04s

$ python3 -m pytest tests/ -q
1600 passed, 103 skipped in 4.72s
```

(1600 rather than the 1556 this branch started against, because #72 landed
under it and brought its own tests. Every one of them passes on the rebase.)

And what the four fixtures record now:

```
Arc      c=[(-4.000000,3.000000,0.000000)]
Circle   c=[(-6.000000,-2.000000,-1.000000)]
Circle   c=[(-2.138803,6.902735,4.666667)] normal=[(0.333333,0.666667,0.666667)]
Arc      c=[(-10.000000,0.000000,0.000000)] a0=1.570796 a1=3.141593
Ellipse  major=[(-5.000000,0.000000,0.000000)] p0=-1.570796 p1=0.000000
Circle   c=[(-3.000000,0.000000,0.000000)] normal=[(0.000000,1.000000,0.000000)]
Arc      c=[(0.000000,0.000000,0.000000)] a0=0.000000 a1=1.570796 normal=[(0.000000,1.000000,0.000000)]
Warning  NON_UNIFORM_BLOCK_SCALE "CIRCLE crossed under a block transform that does not scale its plane uniformly..."
```

The oblique circle is the one worth checking by hand, because its centre does
not depend on which x axis the algorithm picks: a point at OCS z = 7 lands on
the plane `p . N = 7`, and `(-2.138803, 6.902735, 4.666667) . (1,2,2)/3` is 7.
Unlifted it is 19/3.

## What a mirror does to an arc, and what it cannot do

Under a reflection the image of a counter-clockwise sweep runs the other way
round the circle, so the record's two ends swap: `start` becomes
`turn - end_angle` and `end` becomes `turn - start_angle`. The included angle
is preserved, the shape is exact. What a mirror cannot preserve is which end
the drawing called the start, and no record whose angles are counter-clockwise
about its normal has anywhere to say it. The `Ellipse` is the same statement in
its two parameters, and both are negated and swapped.

The probe matters here. The block's arc is a quarter turn from (14, 0) to
(10, 4) about (10, 0). Reflected that is (-14, 0) to (-10, 4), and the wrong
answer swept from (-14, 0) to (-10, -4): the point at (-14, 0) sits on the
mirror's own axis and comes out identical either way, so a test that probed the
endpoints alone could not fail. The midpoint is at 45 degrees, off both axes,
and it is what moves.

## Item 4, the `ViewBegin` extents

`docs/WIRE.md` leaves `ViewBegin` outside the finiteness guarantee because its
extents are a bounding box the source reports rather than a shape anybody
draws. The reasoning is right and there is a hole underneath it: a layout's
extents are four doubles a file holds, so a drawing can hand the shim a `NaN`
and it crosses. `g13_bad_extents.dwg` is that drawing, and before the fix the
capture read:

```
"view_extents": ["NaN,-Infinity,Infinity,NaN", "NaN,-Infinity,Infinity,NaN"]
```

```
$ python3 -m pytest tests/test_view_extents.py -q
...
4 failed, 67 passed in 0.04s
$ python3 -m pytest tests/test_view_extents.py -q
102 passed in 0.03s
```

Of the three options the issue lists I took the documented sentinel, because it
is the only one that needs no change to the wire format, the header or the
frozen record layout. Extents that are not four finite numbers report the
inverted box, `min` at `1e20` and `max` at `-1e20`, which is the pair AutoCAD
itself writes into `EXTMIN` and `EXTMAX` for a drawing with nothing in it.
`min_x > max_x` is then a comparison that works.

It does not separate an empty view from a damaged drawing: both come out as
that box. That wants a warning code of its own, and a code wants a row in
`docs/WIRE.md`, which is not mine this wave. There is a comment below with what
the row should say.

## Blast radius

The whole recorded corpus is unchanged apart from the four new fixtures and one
warning message, both real drawings included. Nothing that was already right
moved, which is the strongest thing I can say about a change that touches every
geometry arm in the flattener. Re-checked file by file against `65f7ddc` after
the rebase onto #72, rather than carried over from before it: one dump out of
twenty-seven moved, and the line in it is the warning's new wording.

`g13_scenarios.json` moved only in measurement noise, the two new fields, and
`warnings/nonuniform_block_scale`'s `output_bytes`, which grew by 108 because
the message it carries is longer. No scenario's `open_code`, `decode_code`,
`code`, record count or batch count changed.

The benchmarks are not re-recorded. They are bound to fixture digests rather
than to the shim, no benchmark fixture changed, and re-recording timing numbers
on a machine that is running four other lanes would be noise somebody then has
to review.


spdrman:
## Folded in: five things I found on the way

None of these is in #65. All of them are in this PR, because they are in the
code this PR rewrites and splitting them out would have meant two branches
editing the same forty lines.

**Handedness was measured in two dimensions.** `Basis.Mirrored` was
`(x.X * y.Y) - (x.Y * y.X) < 0`, which is the determinant of the top-left 2x2
and says nothing about a transform that leaves the XY plane. An insertion that
stands its block on its side measured exactly 0 there, and an insertion with
`ZScale -1`, which is a reflection, measured positive. It is the triple product
of the three transformed axes now, which is the determinant.

**`IsIdentity` never looked at z.** It compared the origin and the images of X
and Y and stopped, so `diag(1, 1, -1)` read as the identity and every arm that
short-circuited on it skipped the mirror handling entirely. The property is
gone: `Turn` is 0 for the identity by construction, so nothing needed the
special case.

**The uniform gate compared two lengths and nothing else.** A 3x scale composed
with a 45 degree rotation gives two in-plane axes of `sqrt(5)` each that are no
longer at right angles, so it passed as uniform and `NON_UNIFORM_BLOCK_SCALE`
said nothing. A circle under that is an ellipse, and the direction the miss
went is the damaging one: the consumer is told the parameters describe the
shape. `g13_ocs_skew.dwg` is that case and the gate now carries an
orthogonality term beside the length one.

**Upstream's arbitrary axis has an integer division.**
`Matrix4.GetArbitraryAxis` in CSMath writes its threshold as
`Math.Abs(zaxis.X) < (1 / 64)`. Both operands are `int` literals, so `1 / 64`
is 0 and the test never fires. The band exists because near the world z axis
the cross product with z shrinks towards nothing and its direction is decided
by the last few bits of the normal; with the band gone, a normal of
`(1e-3, 0, 1)` gets an x axis ninety degrees from the one DXF defines. This
file runs its own algorithm rather than calling upstream's, and builds the
INSERT transform with it too, which is identical to `Insert.GetTransform()` for
any insertion with a +Z extrusion. I have not reported it upstream; say the
word and I will.

**TEXT's insertion point was never lifted either.** It is in the plane the
entity's normal defines, exactly like an arc's centre, so it goes through the
same lift now. Its rotation does not and cannot: record 10 carries no normal,
so there is nowhere on the wire to tell a consumer which plane the angle is
measured in, and the in-plane angle is the closest this version gets. Giving
`Text` a normal is a wire change and is not this one. MTEXT's insertion point
is already world coordinates and is untouched.

spdrman:
## Three things in `docs/WIRE.md` that this PR makes wrong or leaves unsaid

`docs/WIRE.md` belongs to another lane this wave and I have not touched it.
Here is what needs changing in it, so whoever owns it can take the wording
straight from this comment.

**1. Row 106's meaning is now stale.** It reads:

> | 106 | `NON_UNIFORM_BLOCK_SCALE` | An insertion scale that is not a
> similarity, under which a circle is an ellipse and a bulge is an elliptical
> arc. The parameters still cross unchanged; this says they were measured in a
> frame the transform does not preserve. |

"not a similarity" includes a reflection, and after this PR a reflection raises
nothing: an `Arc`'s angles, an `Ellipse`'s parameters and a bulge all follow a
mirror now, so the records are exact under one. What raises it is a transform
that does not scale the entity's plane uniformly, which includes a uniform
scale composed with a rotation. Suggested replacement:

> | 106 | `NON_UNIFORM_BLOCK_SCALE` | A block transform that does not scale an
> entity's plane uniformly, under which a circle is an ellipse and a bulge is
> an elliptical arc. The parameters still cross unchanged; this says they were
> measured in a frame the transform does not preserve. A reflection is not this
> case: a mirror preserves every shape exactly and the records follow it. |

**2. The finiteness section should mention what `ViewBegin` now reports.** The
paragraph that begins "The guarantee is scoped to geometry on purpose" is still
true, and it is now half the story. A producer never puts a `NaN` or an
infinity in `ViewBegin`'s extents either; a view whose extents are not four
finite numbers reports `min_x = min_y = 1e20` and `max_x = max_y = -1e20`,
which is the pair AutoCAD writes into `EXTMIN` and `EXTMAX` for a drawing with
nothing in it, and which a consumer tests with `min_x > max_x`. Suggested
addition after that paragraph:

> `ViewBegin`'s extents are still never `NaN` or infinite. A view whose extents
> the source cannot give reports the inverted box instead, `min` at `1e20` and
> `max` at `-1e20`, and a consumer reads `min_x > max_x` as "this view has no
> usable extents". The box does not say why: an empty view and a damaged one
> report the same thing.

**3. Record 5 never says where angle zero is.** This is a hole rather than
staleness, and I hit it writing the fix. `Arc` is specified as

> Angles are radians, counter-clockwise, measured in the plane the normal
> defines.

which fixes the sense but not the origin, and a consumer cannot draw an arc
without both. The producer's convention, before and after this PR, is DXF's:
zero is along the x axis DXF's arbitrary axis algorithm gives for the record's
own normal, so for a normal of +Z it is world +X. That needs writing down, and
the algorithm named, because for a normal of (0, 1, 0) it is (-1, 0, 0) and
nobody guesses that. It is worth saying in the same breath that the threshold
in the algorithm is `1/64` as a real number: the version of it in the CSMath
this shim builds on writes it as integer division and gets zero (see the
fold-in comment above).

Nothing in this PR changes the convention. It was already what the shim emitted
for every +Z record, and the fix keeps it.

spdrman:
## Two things about the tests that are worth knowing before the next one

**`pytest.approx` does not reach inside a list of tuples.** It compares each
tuple with plain `==`, so

```python
[(1.0, 2.0)] == pytest.approx([(1.0, 2.0000001)], abs=1e-5)   # False
```

A nested comparison is an exact one wearing approx's clothes, and it passes for
as long as the numbers happen to round to the same six decimals the dump
prints. Three of my point assertions were exact comparisons that only looked
approximate; they went red the moment a value came out of an angle rather than
straight off the dump. Every multi-point comparison in `test_flatten_ocs.py`
flattens to one list of numbers first, through a helper that says why.

Nothing already in the tree is affected: `test_flatten_curves.py`'s point
assertions are all single tuples, which are flat sequences and do get the
tolerance.

There are two tolerances in the new file for the same reason. A coordinate read
off the dump is good to `1e-6`. A point reconstructed through an angle is not:
the dump prints six decimals, so an angle on it is good to `5e-7` and the
radius multiplies that into the position, which at `r = 5` is a couple of parts
in a million however right the shim is.

**One branch in the fix cannot be watched failing, and I want that on the
record rather than in a comment nobody opens.** The `IPolyline` arm takes a 3D
polyline's vertices from the plain placement rather than the lifted one,
because POLYLINE's 3D flag is exactly the flag that says its vertices are world
coordinates. `g13_ocs_plane.dwg` asks for an extrusion of -Z on its 3D polyline
and the record comes back with +Z: a DWG has nowhere to put a 3D polyline's
extrusion and upstream reads it as the default. So the wrong behaviour, lifting
by the entity's own normal, is the identity on any 3D polyline this format can
hold. The branch is right per DXF and this corpus cannot make it fail.

What the corpus can do is prove the other side runs, and that was missing at
first: an LWPOLYLINE and a POLYLINE reach the flattener through different cases,
and the LWPOLYLINE in the fixture says nothing about the shared arm. There is a
2D POLYLINE with a -Z extrusion in there now whose vertices do get lifted, so
the arm is exercised rather than assumed. Its z comes out 0 rather than the
elevation the fixture sets, because upstream's `DwgWriter` does not round-trip a
2D polyline's elevation; the x negating is what says the lift ran.

spdrman:
## The other convention, and why this is not it

There are two self-consistent ways to make a mirrored `Arc` come out right, and
the one this PR does not take is worth writing down because it is the first
thing a reviewer will reach for.

**Flip the normal.** Emit `normalize(L·u × L·v)` as the record's normal without
correcting its sign, so a reflection turns the normal over. Then
counter-clockwise about the *new* normal is the mirror image of
counter-clockwise about the old one, and the angles, the parameters and the
bulges all cross unchanged. No correction anywhere, and the arc keeps the
traversal order the drawing gave it.

It is exact and it is ruled out by what already shipped. #56 chose the other
convention for a polyline: `g13_mirrored_bulge.dwg` pins the mirrored record as
`normal=(0,0,1)` with `bulges=[0, -0.5, 0]`, so the normal is the transformed
plane normal and the bulge is negated. `docs/WIRE.md` then ties the two
together, "positive when the arc runs counter-clockwise about the normal,
matching the `Arc` record's angles", so the two records cannot use different
conventions without that sentence becoming false. Taking the flip would mean
re-recording a fixture and an expectation that landed yesterday and changing
what a bulge means on the wire, to buy something the records cannot express
anyway.

**So: keep the normal, correct the angles.** The emitted normal is the plane
normal on the same side the entity's own normal was on, which under a mirror is
`L^-T N` rather than `L·u × L·v`. An `Arc`'s two ends then swap and an
`Ellipse`'s two parameters are negated and swapped.

What that costs is the traversal direction: the mirror of a counter-clockwise
sweep is a clockwise one, and a record whose angles are counter-clockwise about
its normal has nowhere to put that. So the mirrored arc runs from what was the
end to what was the start. The shape is exact and the included angle is
preserved; only the word "start" moves. A polyline does not have this problem
because its vertex order carries the direction and only the bulge sign needs
correcting, which is exactly why #56 could be lossless here and this cannot.

The third option, normalising the emitted angles into `[0, 2*pi)`, I did not
take either. It would change the numbers on every arc under a rotated block for
no gain, and the sweep is what a consumer reads, not the absolute values.

spdrman:
## Run locally as well, because the conformance job is the slow one

The whole point of the archive is that a consumer links what we publish, so I
built it from this branch and ran both consumers against it the way
`acadsharp-conformance.yml` does, on arm64 with `--platform` passed explicitly.

```
$ python3 acadsharp/build_acadsharp.py --platform linux --arch arm64
  ...
  ---- shared smoke ----
  ABI_VERSION=1
  ABI_FINGERPRINT=a7805641d1a3b64f
  BACKING_VERSION=3.7.1
  ...
  verify_archive.sh: all invariants hold for acadsharp-linux-arm64.tgz
  ok  acadsharp-linux-arm64.tgz  (15034 KB)  [built]

$ VIPRS_LIB_DIR=.../bin/unpacked/acadsharp-linux-arm64 acadsharp/tests/conformance/c/run.sh
  158 checks, 0 failures

$ VIPRS_LIB_DIR=.../bin/unpacked/acadsharp-linux-arm64 acadsharp/tests/conformance/rust/run.sh
  137 checks, 0 failures
```

Both skip the exception block, because an archive is not the test
configuration; CI publishes `AbiTest` afterwards and runs them a second time
against that. The fingerprint is unchanged, which it should be: nothing in this
PR touches the header.

The numbers worth knowing are that the NativeAOT publish is clean, so none of
the new geometry is AOT-hostile, and that the two consumers still agree with
the library on every record shape after the flattener was rewritten under them.
That was the one thing the pytest suite genuinely could not tell me, because
every test in it reads a committed capture.

The local run is arm64 native. Nothing here was measured on x64; the conformance
job runs on `ubuntu-24.04-arm`, which is the same architecture.

spdrman:
## Rebased onto `65f7ddc`, and the corpus claim re-checked against it

#72 made `VersionGate`'s two constants aliases of `AbiConstants`, which is a
change under `native/`, so the shim rollup moved and the rebase conflicted on
exactly one line of `MANIFEST.json`. Regenerated rather than hand-merged: a
digest somebody typed is a digest that means nothing.

Two lines move, and only two:

```
-    "sha256": "fcab1bed...ae9de4f5",
+    "sha256": "392fccb9...b53e36175",
-      "native/Adapter/VersionGate.cs": "8b6d2ce1...6dea9331",
+      "native/Adapter/VersionGate.cs": "203178cf...407029fa",
```

No expectation moved with it, which is the check worth having: the alias
carries the same 1014 and 1032 the literals did, so `gate/signature_out_of_range`
still records `UNSUPPORTED_FORMAT` with `AC1009 is outside AC1014 to AC1032`,
and the scenario capture beside it is still a recording of this code rather
than of the one before the rebase.

### The claim, re-run against the new main rather than carried over

Every pre-existing expectation dump, checked one at a time against `65f7ddc`:

```
$ for f in <every .txt in tests/expectations on main>; do
      git diff --quiet origin/main -- "$f" || echo "MOVED: $f"
  done
MOVED: acadsharp/tests/expectations/g13_nonuniform.txt
```

One file out of twenty-seven, and the only line in it is the warning message
getting its new wording. `g13_mirrored_bulge.txt`, `g13_hatch.txt`,
`real_AC1032.txt` and `real_AC1018.txt` are byte for byte what main has.

`pytest acadsharp/tests` is 1600 passed / 103 skipped on the rebased tree, up
from 1556 because #72 brought its own tests with it. The whole repo is 1910
passed / 103 skipped. `ruff check`, `ruff format --check` and
`tools/shellcheck-all.sh` are clean.

#72's four new rules in `test_shim_source_rules.py` are all about the DWG range
in `Abi.cs` and `VersionGate.cs`, neither of which this PR touches, so there is
no interaction between the two changes beyond the digest.

PR #72, opened 2026-09-13
The DWG read range as a measured fact, and stage.sh out of the Python string

spdrman:
The DWG read range as one shim constant and a measured fact, `stage.sh` out of
the Python string, and the capabilities symbol taken from the header rather
than typed.

Refs #59. **PR B** and **PR D** from the triage table on that issue, plus a
third piece handed over from the ABI lane because both files are this lane's.
No closing keyword: #59 has several PRs and only the last one closes it.

## Checklist items covered

- [x] ~~The DWG read range lives in five uncoordinated copies~~, done the way
      the triage reshaped it: no macros in the header, the conformance
      consumers keep their literals, `VersionGate`'s second copy collapses.
- [x] `STAGE_SH` is ~185 lines of shell inside a Python string, invisible to
      `tools/shellcheck-all.sh`, carrying `# shellcheck disable` comments
      nothing will ever read.
- [x] `LINKINFO.md`'s `schema_version` rule contradicting itself (found during
      triage, folded in here because this PR is what adds a field).
- [x] `archive_smoke_source()` hardcodes the symbol it resolves (found by the
      ABI lane, handed over on #69).

## B: the range

Three places wrote the range down and nothing compared them.
`AbiConstants.DwgVersionMin`/`Max` is what `viprs_acad_capabilities_v1`
answers with; `VersionGate.MinVersion`/`MaxVersion` were literals of their own
and are what decides whether a file is opened at all, so a bump applied to one
pair gives a library that advertises a range it does not enforce, and neither
side fails to compile; and `release-acadsharp.yml` regexed an English sentence
out of the frozen header with `test_release_workflow.py` pinning that wording,
so rewriting a comment in a contract document was a release failure.

The range is a fact about the backing reader, not a clause in the ABI, so it
is measured rather than declared:

- `archive_smoke.c` already calls capabilities against the staged library. It
  prints `DWG_VERSION_MIN=` and `DWG_VERSION_MAX=` beside `BACKING_VERSION=`
  now, and refuses a library answering with a range that holds no drawing.
- `stage.sh` records both as facts, the way it records the backing version.
- `make_linkinfo` writes `dwg_version_min` and `dwg_version_max` as integers,
  refusing a value that was not measured, is not an integer, is not a DWG
  version signature, or is inverted.
- `verify_archive.sh` compares both against what the shipped library answers,
  in the block that already links and runs it for the fingerprint. This is the
  one manifest field nothing in the bytes can settle: the header does not
  state the range, so a manifest claiming a range the library does not read is
  well-formed, and the number a consumer refuses a drawing on is the wrong one.
- The release notes read the range out of the published archives and fail the
  run when the five disagree, naming both ranges and which archives carry each.
- `VersionGate`'s constants are aliases of `AbiConstants` now, held to one copy
  by a source rule.
- The conformance consumers keep `1014`/`1032` and each says why: independent
  oracles of the pinned build, the same argument `layout_table.h` makes.
- `LINKINFO.md` gets two rows, keeps `schema_version` at 1, and loses the
  sentence saying adding a field moves the version.

## D: `stage.sh`

See the separate comment for the shellcheck output, which is the first time
that script has ever been linted. In short: it is a tracked, executable
`acadsharp/scripts/stage.sh`, the two placeholders are `${RUNTIME_ARCHIVES:?}`
and `${STATIC_SYSTEM_LIBRARY_LADDER:?}` set as `ENV` lines in
`make_dockerfile` and in the mac `env`, the one disable earns a reason line,
and the one I nearly added was measured and dropped.

## The third piece: the capabilities symbol

`archive_smoke_source()` took ten of eleven names from the header and
hardcoded the eleventh. #69 renames that call, and from the moment it lands
the smoke resolves NULL, exits 7 and the driver refuses every archive. The
handover carried a two-line literal swap; I took the name out of the header
instead, so this branch asks for today's spelling and tomorrow's with no
second edit and no ordering between the two PRs. The same literal was in the
verifier's probe, put there by this branch, where it would have failed to
compile rather than to resolve. The guard that closes the class: every name
the smoke hands to `dlsym` is one the header declares, plus a test that
generates the smoke from a renamed list and asserts it follows.

## Red, then green

Every item was watched failing first.

| item | the red | the green |
| --- | --- | --- |
| one shim copy | `test_only_abi_cs_assigns_the_range` names `['Abi.cs', 'Adapter/VersionGate.cs']` | one name |
| the manifest field | 45 failed in `test_build_manifests.py`, `make_linkinfo` had no such argument | all green |
| the verifier | 6 failed: two frozen-field cases, the shape cases, the live one and its control | 121 passed |
| the workflow | 6 failed: the header regex still there, no range in the notes | 99 passed |
| `stage.sh` tracked | `test_git_actually_tracks_it` fails until `git add` | green |
| the dlsym guard | `test_a_renamed_capabilities_call_carries` fails against the hardcoded literal, put back to check | green |

The live range case is the one worth reading: hand-edit an archive's
`dwg_version_max` to `1035` and the verifier refuses by name, saying the
library answers `1032`. The control beside it asserts the good archive's range
*is* reported, because a check that only fires on a number somebody broke could
be firing on every number, and that line is what proves the probe reached the
library at all.

Locally: `1496 passed, 5 skipped`, `tools/shellcheck-all.sh` clean, `ruff
check` and `ruff format --check` clean, all in a `linux/arm64` container.

## The move changed nothing, on three platforms

The static ladder differs per platform, so the environment variable had to be
proven where it is read. Same tree built before and after the move:

| cell | where | facts.txt |
| --- | --- | --- |
| linux/arm64 | this Mac, `--platform linux/arm64` | identical |
| linux/x64 | the NAS, native x86_64, in a container | identical |
| musl/x64 | the NAS, native x86_64, in a container | identical |

On arm64 I unpacked both archives and compared them: the shared library, both
static archives, the header, the three contract documents and the README are
byte for byte the same. The only differences were `BUILDINFO`'s timestamp and
one line in `THIRD_PARTY_NOTICES`, which is its own finding and its own commit
(see the comment below).

A real musl archive's manifest out of that run, which is also the range
arriving in a published artifact:

```json
  "target": "x86_64-unknown-linux-musl",
  "static_certified": true,
  "static_system_libraries": ["m"],
  "dwg_version_min": 1014,
  "dwg_version_max": 1032
```

Both NAS scratch trees and both builder images were removed afterwards.


spdrman:
## Things I found on the way, folded in or written down

Six of these, in the order I hit them. Three are fixed in this branch, two are
limits of what I built that the next person should know about, and one belongs
to files another lane owns.

### 1. The archive was not reproducible, and the release notes publish its digest (fixed here)

Found while proving the `stage.sh` move changed nothing: I built the same tree
twice and the archives differed. The only difference besides `BUILDINFO`'s
timestamp was one line in `LICENSES/THIRD_PARTY_NOTICES`:

```
  Restored /work/native/Viprs.ACadSharp.Native.csproj (in 80 ms).
  Restored /work/native/Viprs.ACadSharp.Native.csproj (in 198 ms).
```

`dotnet list package` restores first and says how long that took, so the notices
file, `CHECKSUMS.txt` and the archive's own sha256 all depended on how loaded the
machine was. `release-notes` puts those digests on the release page, so two runs
of one commit produce two sets of numbers with nothing to say which is right.
Fixed in its own commit, filtered out of the block, and there is a test. What
belongs there is which packages were restored, never how long it took.

### 2. The issue's count of `# shellcheck disable` comments is one too many

The lows say `STAGE_SH` carries "two `# shellcheck disable` comments nothing will
ever read". There is one, `SC2086` on the static smoke's link line. It is
load-bearing, measured by taking it out (two SC2086 warnings come back), so it
earns a reason line rather than going. The other two live in
`verify_archive.sh`, which is a tracked `*.sh` and has always been linted.

I also reached for a second disable on the new `for name in ${RUNTIME_ARCHIVES:?}`
loop, on the assumption that an unquoted expansion needs one, and then checked:
shellcheck does not raise SC2086 on a `for` list. So it would have been exactly
the kind of line the issue is complaining about, and the comment there now says
that, with the measurement.

### 3. The fixture library had to learn to answer, or the new check would have passed over anything

`test_verify_archive.py`'s stand-in library defined every export as
`uint32_t name(void) { return 0u; }`. The verifier now asks the library what
AC10xx range it reads, and a stub like that answers whatever the register
happened to hold. So `viprs_acad_capabilities_v1` is written out properly in the
fixture, filling the struct the shipped header declares, in a translation unit of
its own because every other stub's signature is one the real header contradicts.
Without that, `test_a_read_range_the_library_does_not_report_is_rejected` would
have passed for the wrong reason.

### 4. The live range check does not run on every archive, and that is not a hole I can close here

`verify_archive.sh` can only ask the library when it can link and run it: the
host has to be able to build for the target, and the archive has to carry the
static half. So the live comparison covers a Linux cell verified on its own
architecture, and never a mac archive, which is shared-only. Verifying a
linux/arm64 archive on this Mac prints
`static_certified: not link-tested (this host cannot build for linux/arm64)` and
skips it.

What covers every target is the build's own smoke, which dlopens the staged
library in the container before the archive is packed, and the driver turns a
failed smoke into a hard error. The manifest's shape (integers, AC10xx codes, low
not above high) is checked from the bytes on every archive on every host. I would
rather state that than pretend the live half is universal.

### 5. `MANIFEST.json` will conflict with the geometry lane

Editing `VersionGate.cs` makes every capture a recording of a program that is no
longer in the tree, so this branch reruns `regenerate.py --only expectations` and
carries a new `MANIFEST.json`. `.lane-65-geometry` was running the same generator
at the same time on this machine, adding a fixture and a `views_without_extents`
field. Both branches touch the `shim` block. Whoever merges second re-records
rather than resolving the digests by hand: a hand-merged digest is a number
nobody measured, which is the one thing that file exists to prevent.

### 6. Not mine to change: the header and `ABI.md` still state the range in prose

`acadsharp/include/viprs_acadsharp.h` says, around line 151, "the numeric AC10xx
codes, so 1014 and 1032 for ACadSharp 3.7.1", and `docs/ABI.md` has
`1014 today` / `1032 today` beside the two struct fields. Both belong to the ABI
v2 lane, so I have left them alone.

They are no longer load-bearing: nothing greps either of them now, which was the
point of retiring the workflow regex. But they are still two copies of a fact
about the backing reader sitting inside a document whose own preamble says it
describes a VIPRS boundary and never ACadSharp, and they will go stale the day
the pinned reader moves. **Suggested change for whoever owns those files:** in
the header, say that the range is reported by `viprs_acad_capabilities_v1` at run
time and recorded in each archive's `LINKINFO.json`, and drop the two numbers; in
`ABI.md`, do the same for the two field comments. A1 can now reword that
paragraph freely without breaking a release, which it could not do before this PR.

spdrman:
## Composing with #73 (the musl unwinder lane), measured rather than guessed

#73 edits `STAGE_SH` and this PR deletes it, which is the one pairing that
rebases badly: git sees an edit against a deleted region and hands you a
conflict with no hint about where the block belongs. So I went and looked at
their branch rather than waiting for it.

**Their block survives the move and the lint.** I took `STAGE_SH` from
`origin/pr73` at `bd1e36d`, applied this PR's two substitutions
(`${RUNTIME_ARCHIVES:?}` and `${STATIC_SYSTEM_LIBRARY_LADDER:?}`) and ran
shellcheck over the result as a file:

```
$ shellcheck composed-stage.sh
exit=0
```

That answers the third of the three things their rebase note asks for: the
`privatise_unwind.sh` block is clean once it is a tracked `*.sh`, so nothing
in it has to change to survive being linted for the first time.

### If this PR lands first, #73 rebases (cheaper, and what I would do)

Their `STAGE_SH` hunk becomes a plain edit to `acadsharp/scripts/stage.sh`, in
the same window, between `ranlib "$MERGED"` and `echo "---- static smoke ----"`.
Then three one-liners:

- `make_dockerfile`'s `COPY` line: add `privatise_unwind.sh` (this PR left the
  line otherwise as it was).
- `_write_build_context`: this PR already turned the `retain_sections.py`
  `copy2` into a loop, so it becomes
  `for name in ("retain_sections.py", "privatise_unwind.sh", "stage.sh")`, with
  the `chmod` on `stage.sh` kept.
- `test_privatise_unwind.py`: `ba.stage_script("musl")` becomes
  `ba.stage_script()`. The function still returns the script's text, which is
  what their note asks for, but it no longer takes a platform: nothing is
  substituted per platform any more, the two lists are environment variables.
  There are six call sites, all in that file, and `test_the_stage_script_calls_it`
  loops over platforms for a string that is now the same for all of them.

### If #73 lands first, I rebase

Same three edits in the other direction, and I carry their block into
`stage.sh` by hand rather than letting git decide. I have the composed file
already, so this is a paste, not an exercise.

### Either way, two collisions to know about

**`verify_archive.sh`'s invariant list.** They add an invariant 8 (the bundled
libunwind ships under private names) and so does this PR (the read range is the
range the library answers with). Whoever is second renumbers theirs to 9. The
code they add is in the `walk_archive` branch and the two `else` arms at the end
of the `static_certified` section; mine is the `PROBE_STATUS` arm just above
them and the `cc` line that compiles the probe. Adjacent, not overlapping.

**`tools/shellcheck-all.sh` picks up three new files between us**, not one:
their `privatise_unwind.sh` and `verify_archive_matched_host.sh`, and this PR's
`stage.sh`. All three are clean today, theirs on their branch and mine on this
one, and the composed `stage.sh` above is clean too.

spdrman:
## PR D: `stage.sh` stops being a Python string

`build_acadsharp.py` held 185 lines of shell in `STAGE_SH`, with two placeholder
words substituted into it before it was written into the build context.
`tools/shellcheck-all.sh` discovers its work with `git ls-files '*.sh'`, so the
most intricate shell in this repository was the one part of it no linter had
ever opened, and the `# shellcheck disable` in it was addressed to nobody.

It is `acadsharp/scripts/stage.sh` now, tracked and executable. The two
placeholders became `${RUNTIME_ARCHIVES:?}` and
`${STATIC_SYSTEM_LIBRARY_LADDER:?}`, set as `ENV` lines in `make_dockerfile` and
put in the process environment for the mac path, so the file is valid shell as
it sits in the tree and what runs in the container is byte for byte what is in
the repository. `:?` rather than a bare expansion on purpose: unset,
`for name in $RUNTIME_ARCHIVES` is an empty loop, the merge produces an archive
of one object, the static smoke fails to link it and the cell ships shared-only,
which is a recorded outcome nobody reads.

### Shellcheck, the first time it has ever seen this script

Run on the extraction before a single edit, which is what the issue asked for:

```
$ shellcheck acadsharp/scripts/stage.sh
In acadsharp/scripts/stage.sh line 170:
            for name in RUNTIME_ARCHIVE_LIST; do
                        ^------------------^ SC2043 (warning): This loop will
                        only ever run once. Bad quoting or missing glob/expansion?
```

One finding, and it is the placeholder: `RUNTIME_ARCHIVE_LIST` is a bare word
until Python substitutes the list in, so shellcheck reads the loop as running
once. `${RUNTIME_ARCHIVES:?}` is what removes it. With optional checks on
(`-o all`, which CI does not run) there are 124 SC2250 brace-style suggestions
and 5 SC2312 masked-exit notes, and nothing else.

The issue says two `# shellcheck disable` comments. There is one in this script,
`SC2086` on the static smoke's link line, and it is load-bearing: removing it
produces two SC2086 warnings, so it earns its reason line. The other two live in
`verify_archive.sh`, which shellcheck has always seen. I also reached for a
second disable on the `for name in ${RUNTIME_ARCHIVES:?}` loop and then measured
it: shellcheck does not raise SC2086 on a `for` list, so the directive would
have been another line nothing reads, and the comment there says so.

`tools/shellcheck-all.sh` now lists and lints the file, and passes.

spdrman:
## Scope note: five hunks in `LINKINFO.md`, not two

My brief said to touch the field table and the `schema_version` rule only,
because that document has other lanes in it. I touched five places, and the
three extra ones are here so somebody can say no to them:

| lines | what | why |
| --- | --- | --- |
| 48 | the `schema_version` rule | in scope: the sentence that said adding a field moves the version, three lines above the rule saying fields are only ever added within one |
| 94 | the field table | in scope: the two new rows |
| 384, 409 | both worked manifests | **extra.** The section is called "A complete manifest" and both examples are marked as always-present-field complete. Leaving the two new fields out would make the only worked example in the contract an invalid manifest, which is what a consumer author copies |
| 447 | one sentence in "Verifying an archive before trusting it" | **extra.** It enumerates what `verify_archive.sh` checks, and this PR adds a check: that the library answers with the range the manifest states |

None of them is the linking recipe or the musl section, which is where #73 is
working. I have not touched `docs/ABI.md` or the header at all.

spdrman:
## Two more found by re-reading my own diff, both fixed here

Neither came from a test going red. Both came from reading the two most
delicate files once more after the branch was green, which is the reading that
found them.

**The notes generator raised over exactly the archive it exists to report.** It
groups the published archives by the range each recorded and sorts the groups.
A stale asset that a `--clobber` did not replace carries no range, so one key is
a pair of `None`s, and in Python a tuple holding `None` does not compare with a
tuple holding integers. `TypeError`, the step dies, and the release page says
nothing at all. My own test missed it by staging one archive, where nothing is
ever compared. The case that actually happens is four cells rebuilt and one
stale, so that is what the test stages now, and the sort goes through a key.

**The capabilities lookup could abort the verifier.** The line that reads the
call out of the shipped header's declarations is a `grep | head -1`, and
`verify_archive.sh` runs under `set -euo pipefail`. A header declaring no such
call makes grep exit 1, pipefail carries it through `head`, and the assignment
takes the whole verification down three lines before the refusal that would
have explained it. The failure would have been an empty exit 1. `|| true` on
the pipeline, and the test pulls that line out of the script and runs it under
the same options rather than restating it, which is red without the guard.

Both are mine, introduced in this branch, and both are the same shape: a path
that only opens when something else is already wrong, which is the path nobody
exercises.

spdrman:
## The live check, on a real archive, on a real runner

The fixture suite proves the verifier refuses a manifest the library disagrees
with. This is the other half: a real NativeAOT build, on `ubuntu-24.04-arm`,
verified on a host that matches the target so the probe actually links and runs.
From the `consumers` job on this branch:

```
consumers  Build the linux/arm64 archive
  [linux/arm64] #16 1.908 DWG_VERSION_MIN=1014
  [linux/arm64] #16 1.908 DWG_VERSION_MAX=1032
  [linux/arm64]   static_certified holds: the archive links and runs here
  [linux/arm64]   the library reads AC1014 to AC1032, which is what LINKINFO.json says
  [linux/arm64]     DWG_VERSION_MIN=1014 DWG_VERSION_MAX=1032
  [linux/arm64] verify_archive.sh: all invariants hold for acadsharp-linux-arm64.tgz

consumers  Verify the archive before anything runs against it
    static_certified holds: the archive links and runs here
    the library reads AC1014 to AC1032, which is what LINKINFO.json says
```

Three readings of the same fact, in order: the smoke asking the staged library
in the container, the driver's own verify of the packed archive, and the
workflow's second verify before anything is run against it. The first is what
gets recorded in `LINKINFO.json`; the other two are what hold it to the library.

That run was cancelled at a later step when I pushed again, which is why it
shows as cancelled. The steps above had already passed, and nothing since has
touched the code they exercise.

spdrman:
## Green, on `65fd8ab`

```
consumers      pass  2m26s
lint           pass  9s
shellcheck     pass  5s
test (3.12)    pass  1m38s
test (3.9)     pass  1m44s
```

The `consumers` run is the one that exercises the substance rather than the
text: it builds a real linux/arm64 archive through the moved `stage.sh`, on a
runner whose architecture matches the target, so the verifier's probe links and
runs the library rather than skipping itself.

```
consumers  Build the linux/arm64 archive
  [linux/arm64]   the library reads AC1014 to AC1032, which is what LINKINFO.json says
  [linux/arm64] verify_archive.sh: all invariants hold for acadsharp-linux-arm64.tgz

consumers  Verify the archive before anything runs against it
    the library reads AC1014 to AC1032, which is what LINKINFO.json says
  verify_archive.sh: all invariants hold for acadsharp-linux-arm64.tgz

consumers  Run the generated Rust consumer against the test configuration
  149 checks, 0 failures
```

Locally, in a `linux/arm64` container: `1809 passed, 6 skipped` repo-wide,
`tools/shellcheck-all.sh` clean, `ruff check` and `ruff format --check` clean,
and one more full arm64 archive built from this exact commit whose `facts.txt`
is identical to the one built before any of this landed.

PR #73, opened 2026-09-13
Rename the runtime's libunwind so musl can link, and link-test musl on a musl host

spdrman:
Refs #59

Both musl archives have shipped a static cargo recipe that nothing had ever run, and when I ran it, it failed. This closes the gap that hid that, and fixes what it revealed.

## The gap

`verify_archive.sh` holds `static_certified: true` to an archive that links and runs, but only when the host can build for the target. Both musl cells in `release-acadsharp.yml` build on glibc runners, so what they got was

```
static_certified: not link-tested (this host cannot build for musl/x64)
```

written into a log nobody reads, and a green job.

`acadsharp/scripts/verify_archive_matched_host.sh` picks the host instead of accepting whatever host it is on: a musl archive gets verified inside a pinned Alpine container of the same architecture, on the same runner, where the script's own musl detection fires and cc, cargo and rustc are all musl-native. It exports `VIPRS_REQUIRE_LINK_TEST`, which turns that skip line into a failure, so a cell that somehow still cannot link says so. The four container cells in the release workflow go through it, and the conformance workflow gains a `musl-static-link` job so a regression is a red pull request rather than a bad release.

It never emulates an architecture. A cpu that does not match the host is a refusal, for the reason ADR 0001 gives.

## What running it showed

The runtime bundles its own llvm-libunwind inside the merged archive, and rustc links a `self-contained/libunwind.a` of its own for every musl target and for no glibc one. 59 symbols collide, all of them `__unw_*`, `unw_*` or `libunwind::*` internals, and the link dies on all of them.

I reproduced it myself rather than inheriting the numbers, with the glibc archive through the same harness on the same machine as the control every time:

| | arm64 (this Mac, native, `--platform linux/arm64`) | x64 (the NAS, native x86_64) |
|---|---|---|
| glibc archive, cargo recipe | links and runs | links and runs |
| musl archive as published, cargo recipe | 59 collisions, link fails | 59 collisions, link fails |
| musl archive, C recipe | links and runs | links and runs |

rustc 1.98.1 both times, against the published `3.7.1-viprs.1` archives, GNU ld from the Alpine toolchain. The `-crt-static` default and an explicit `+crt-static` fail identically.

## The fix

`acadsharp/scripts/privatise_unwind.sh` renames our copy with `objcopy --redefine-syms`, applied to the merged archive rather than to its members, which is what makes definitions and references move together in one pass. The runtime keeps its own unwinder under private names and rustc's copy sits beside it.

`_Unwind_*` is deliberately not in the selection. That is the public personality ABI a landing pad calls by name, and renaming it would split the language runtimes instead of the unwinder. Measured, the merged archive defines none of them anyway: every collision is an internal name.

After the rename, on both architectures:

| | arm64 | x64 |
|---|---|---|
| collisions with rustc's libunwind | 59 -> 0 | 59 -> 0 |
| cargo recipe | links and runs | links and runs |
| C recipe, 3000 managed throws | passes | passes |
| both unwinders live: 3000 managed throws interleaved with 2000 Rust panics through `catch_unwind`, one binary | passes | passes |
| with `RUSTFLAGS=-Ctarget-feature=+crt-static` | passes | passes |
| member order | unchanged, 247 / 266 members | unchanged |
| `SHF_GNU_RETAIN` on `__modules` | still set | still set |

The last two matter because the linker emits `.init_array` in the order it pulls members, and because `__modules` has to survive `--gc-sections`. Both are checked against the shipped bytes, not assumed.

The panic half is the one that earns its place. Linking is not the question; whether two unwinders can coexist at run time is, and only interleaving real .NET exception handling with real Rust unwinding in one process answers it.

## Two fixes that are measured dead, and stay dead

Dropping `libRuntime.WorkstationGC__libunwind.cpp.o` from the merge breaks the C recipe too: the runtime reaches into that copy's C++ internals, not only its `__unw_*` C API. And no consumer-side flag helps: `-C link-self-contained=no` breaks the build script's own link and `-C link-self-contained=-unwind` is rejected. `acadsharp/docs/LINKINFO.md` keeps that reasoning so nobody spends the day again.

## The invariant that catches a regression from anywhere

`verify_archive.sh` now reads the shipped symbol index for the unwinder's names. No linker, no cargo, no matching architecture, so it holds in the x64 cell for the arm64 archive and on a Mac for both, which is exactly the coverage the consumer link cannot give.

Two halves, and the second is the one that earns its place: the original names are a refusal, **and** an archive carrying neither name is a refusal. Checking only for the old names passes an archive that bundles no libunwind at all, which is also what the rename step silently not running looks like from here. The zero needs a positive control.

## Tests

35 new tests, red on the first commit of this branch and green on the second. They cover the rename over a real archive (references in other members move, mangled names still demangle, `_Unwind_*` untouched, member order preserved, idempotent), both new refusals in the verifier against real fixture archives, the strict flag, and the wrapper and workflow wiring. The verifier fixture grew an unwinder so it looks like the real archive.

Full suite: 1486 passed, 6 skipped. shellcheck and ruff clean.

## Regions of the three shared files I touched

The lane in `.lane-59b-dwgrange` owns these and is moving `STAGE_SH` into `acadsharp/scripts/stage.sh`. I kept to three small regions and expect to rebase.

- **`acadsharp/build_acadsharp.py`**: one block inside `STAGE_SH`, between `ranlib "$MERGED"` and `echo "---- static smoke ----"`; one filename added to the Dockerfile's `COPY` line; the `shutil.copy2` of `retain_sections.py` in `_write_build_context` turned into a two-entry loop. Nothing else.
- **`acadsharp/scripts/verify_archive.sh`**: the invariants comment at the top; one block inside the `if walk_archive ...` branch for the static archive; the two `else` arms at the very end of the `static_certified` section.
- **`.github/workflows/release-acadsharp.yml`**: the `run:` line of build-linux's Verify step, plus a comment above it. One word of behaviour.


spdrman:
## Everything I measured, with the commands, including the things that did not work

The rename works. This is the whole record, so nobody has to re-derive it.

### The end-to-end proof, on a real build

The one measurement that covers the whole chain: I built a musl/arm64 archive with the driver as changed by this PR, on this arm64 Mac, and verified it through the new wrapper.

```
python3 acadsharp/build_acadsharp.py --platform musl --arch arm64
bash acadsharp/scripts/verify_archive_matched_host.sh bin/acadsharp-musl-arm64.tgz musl arm64
```

The build log carries the step, and the fact:

```
privatise_unwind: renamed 79 libunwind symbols in 2 archive(s)
privatised_unwind_symbols	79
static_ok	1
retained_module_sections	1
```

The wrapper's output, in full:

```
verify_archive_matched_host.sh: this host is mac/arm64 and the archive is
  musl/arm64, so verifying inside rust:1.98.1-alpine (linux/arm64)
...
---- acadsharp-musl-arm64/lib/libacadsharp_native.a ----
  members: 249 (247 objects)
  symbol index defines every entry point the header declares
  the bundled libunwind is private (119 symbols), so rustc's own copy
  can sit beside it
  the encapsulation sections are retained, so --gc-sections cannot collect them
  static_certified holds: the archive links and runs here
    ABI_VERSION=1 ABI_FINGERPRINT=a7805641d1a3b64f
  the documented cargo recipe links and runs against this archive

verify_archive.sh: all invariants hold for acadsharp-musl-arm64.tgz
```

The same wrapper against the **published** `acadsharp-musl-arm64.tgz` fails on both new checks, which is the control for the run above:

```
FAIL: libacadsharp_native.a defines 119 libunwind symbols under their original names, ...
FAIL: the documented cargo recipe fails against this archive: ... multiple definition of `__unw_regname' ...
```

### The measurements behind the fix

Every one has the glibc archive through the same harness on the same machine as its control. arm64 is this Mac, native, `--platform linux/arm64` passed explicitly on every `docker run`. x64 is the NAS, native x86_64, in a container. rustc 1.98.1 both times, `rust:1.98.1-alpine`, GNU ld from the Alpine toolchain (2.45.1 on x64, 15.2.0's ld on arm64).

| | arm64 native | x64 native |
|---|---|---|
| glibc archive, cargo recipe (control) | links and runs | links and runs |
| musl archive as published, cargo recipe | fails, 59 collisions | fails, 59 collisions |
| after `privatise_unwind.sh` | 0 collisions, links and runs | 0 collisions, links and runs |
| C recipe, 3000 managed throws, before | passes | passes |
| C recipe, 3000 managed throws, after | passes | passes |
| 3000 managed throws + 2000 Rust `catch_unwind` panics, one binary | passes, 1000 caught | passes, 1000 caught |
| the same with `RUSTFLAGS=-Ctarget-feature=+crt-static` | passes | passes |
| ar member order | unchanged, 247 | unchanged, 266 |
| `SHF_GNU_RETAIN` on `__modules` | still set | still set |

The panic half is the one that earns its place. Linking is not the question; whether two unwinders can coexist at run time is. The probe the verifier links calls one export that returns a constant and cannot tell a working unwinder from a broken one, so I wrote one that drives real .NET exception handling (3000 `viprs_acad_open_memory` calls on a megabyte of garbage, each raising inside ACadSharp and coming back as `VIPRS_ACAD_UNSUPPORTED_FORMAT`) interleaved with real Rust unwinding.

### The three fixes that do not work, re-measured rather than inherited

I did not take these on trust, because a confident reproduction narrative is evidence to check, not a fact to inherit. All three on native arm64 musl, against the published archive.

**`-C link-self-contained=-unwind` is rejected by rustc itself**, before any link:

```
RUSTFLAGS="-Clink-self-contained=-unwind" cargo build ...
error: failed to run `rustc` to learn about target-specific information
```

**`-C link-self-contained=no` breaks the build script's own link**, because rustc still passes `-lunwind` and there is then nothing to find:

```
RUSTFLAGS="-Clink-self-contained=no" cargo build ...
ld: cannot find -lunwind: No such file or directory
ld: have you installed the static version of the unwind library ?
```

**Dropping the libunwind member breaks the C recipe too**, so it is not a musl-only trade:

```
ar d lib/libacadsharp_native.a libRuntime.WorkstationGC__libunwind.cpp.o && ranlib ...
cc probe.c -Wl,--whole-archive ..._init.a -Wl,--no-whole-archive ....a -lm
undefined reference to `libunwind::LocalAddressSpace::sThisAddressSpace'
```

That last one is why the rename had to reach the mangled C++ names and not just the `__unw_*` C API. The runtime reaches into that copy's internals.

### Things I learned on the way that are not obvious from the diff

**The merged archive defines zero `_Unwind_*`.** Measured on both architectures. Every single collision is an internal `__unw_*`, `unw_*` or `libunwind::*` name, which is what makes the rename safe: the public personality ABI a landing pad calls is untouched, and the language runtimes are not split.

**119 and 79 are the same set.** The verifier reports 119 and `privatise_unwind.sh` reports 79. The ar symbol index carries one entry per definition per member, and `nm --defined-only -g | sort -u` counts unique names. Anyone comparing the two numbers and concluding the check over-matches would be wrong: measured on the patched archive, old-named goes 119 to 0 and privatised goes 0 to 119.

**`verify_archive.sh`'s archive index has one leading underscore stripped** (`sed 's/^_//'`, a Mach-O convention that applies to the ELF path too). Any pattern written against `static-index.txt` has to account for that, which is why the new check strips the rest with `sed 's/^_*//'` and matches the bare name.

**`objcopy --redefine-syms` over an archive preserves member order and `SHF_GNU_RETAIN`.** Both measured rather than assumed, and both are load-bearing: the linker emits `.init_array` in the order it pulls members (stage.sh has the long comment about a binary that segfaults on the way out), and the retain flag is what keeps `__modules` alive under `--gc-sections` (#67).

**`rust:*-alpine` does not set `RUSTFLAGS`**, so the musl default applies and the produced binary is statically linked. I checked both that and an explicit `+crt-static` because they are different link lines, and the collision and the fix behave identically under both.

**Alpine's busybox is enough for `verify_archive.sh`.** Its hand-rolled byte readers (`dd`, `od -An -v -tx1`) work under busybox, so the wrapper's container needs only `bash build-base binutils python3 file`. I checked this rather than assuming, because a byte reader silently misreading would have produced a pass, not an error.

### Things I deliberately did not do

**The driver's own `verify_archive()` still calls `verify_archive.sh` directly, not the wrapper.** Routing it through the wrapper would make `build_acadsharp.py --platform musl` self-proving, which is tempting, but the same call verifies linux archives built in containers on a macOS host, and there the wrapper would refuse with "no container is defined for platform 'linux'". That breaks the normal local flow for a gain the CI paths already cover. If someone wants it, the honest shape is a wrapper that falls back to the plain verifier for a platform it has no container for, and that fallback is exactly the silent skip this PR exists to remove.

**The new conformance job does not run the C or Rust conformance consumers against a musl shared library.** `test_conformance_workflow.py` asserts each runner appears exactly twice, against the unpacked archive and against the test configuration, and that pairing is deliberate. The musl gate this lane needs is the static link, which the wrapper runs. Behaviour coverage on musl is a separate and reasonable want.

**The rename runs on glibc too, not only musl.** Nothing collides there, so it changes nothing a glibc consumer can see, and it means the cheap linux cells exercise the same code path the musl ones depend on. A musl-only branch would be a code path that only ever runs in the two slowest cells.

### Files I touched outside the three shared ones

Listed for whoever rebases. `acadsharp/scripts/privatise_unwind.sh` and `acadsharp/scripts/verify_archive_matched_host.sh` are new. `acadsharp/tests/test_privatise_unwind.py` is new. `acadsharp/tests/test_verify_archive.py` gained an `unwind` and a `plat` parameter on `_build_linux_tree` plus the `UNWINDER` source. `acadsharp/tests/test_release_workflow.py` gained `VERIFIER_FOR`. `acadsharp/tests/test_abi_docs.py`'s `TestTheMuslLimitationIsStated` became `TestTheUnwinderRenameIsStated`, because it asserted the limitation that this PR removes. `acadsharp/docs/LINKINFO.md`'s musl section now describes what the archive carries.

`.github/workflows/acadsharp-conformance.yml` is not in the shared three and gained one job.

spdrman:
## CI is green on a real runner, and the musl job is not a fast pass

All six checks pass on `bb65da2` plus the LINKINFO commit. The new `musl-static-link` job took 1m53s, which looked far too quick for a NativeAOT build, so I read the log rather than the verdict. It did the work:

```
[musl/arm64] #14 [ 9/11] COPY archive_smoke.c ... privatise_unwind.sh /work/
[musl/arm64] #16 21.03 privatise_unwind: renamed 79 libunwind symbols in 2 archive(s)
[musl/arm64] #16 21.50 privatised_unwind_symbols	79
...
verify_archive_matched_host.sh: this host is linux/arm64 and the archive is
  musl/arm64, so verifying inside rust:1.98.1-alpine (linux/arm64)
  the bundled libunwind is private (119 symbols), so rustc's own copy
  can sit beside it
  the encapsulation sections are retained, so --gc-sections cannot collect them
  static_certified holds: the archive links and runs here
    ABI_VERSION=1 ABI_FINGERPRINT=a7805641d1a3b64f
  the documented cargo recipe links and runs against this archive
```

That last line is the one that has never appeared for a musl archive before. `ubuntu-24.04-arm` is simply much faster at this than Docker Desktop on a Mac, where the same build takes about fifteen minutes.

`consumers`, which this PR does not change, is still green at 2m33s.

## A note for whoever rebases the `STAGE_SH` move

The lane moving `STAGE_SH` out into `acadsharp/scripts/stage.sh` needs three things to survive the move, and none of them is obvious from the diff alone:

1. The rename block itself, which sits between `ranlib "$MERGED"` and `echo "---- static smoke ----"`. It has to stay in that window: after `ar qc` so one pass reaches every member, before the smoke so the C link is measured against the archive that ships.
2. `privatise_unwind.sh` in the Dockerfile's `COPY` line and in `_write_build_context`, or `stage.sh` calls a script the image does not have.
3. Once `stage.sh` is a tracked `*.sh`, `tools/shellcheck-all.sh` finds it by its own glob. The block I added is `sh`-compatible and shellcheck-clean today as part of the Python string, but it has never been linted as a file.

`acadsharp/tests/test_privatise_unwind.py::TestTheBuildRunsTheRename` holds all three through `ba.stage_script()`, so if that function keeps returning the script's text after the move, those tests keep working and will say which of the three broke.

## The docs commit

`bd1e36d` rewrites the musl section of `acadsharp/docs/LINKINFO.md`, which shipped inside every archive saying the cargo recipe did not work there. It now describes what the archive carries, keeps the collision and both dead fixes, and `acadsharp/tests/test_abi_docs.py`'s class for it was rewritten to assert the new claims: that `_Unwind_*` is untouched, that the consumer has nothing to do, and that the dead fixes stay written down.

spdrman:
Behaviour coverage on musl, the thing you scoped out deliberately, is now tracked in #74 along with the note that adding a third runner is a real change to `test_conformance_workflow.py`'s exactly-twice rule rather than a line. Your other two deliberate omissions I did not file: the wrapper fallback because you argued it back to the silent skip this PR removes, and running the rename on glibc too because that is a decision rather than deferred work.

spdrman:
## Rebased onto `65f7ddc`, and the rename block survived the move

#72 took `STAGE_SH` out of the Python string and into a tracked `acadsharp/scripts/stage.sh`, which is exactly the collision the PR body predicted. The block now lives in that file, between `ranlib "$MERGED"` and `echo "---- static smoke ----"`, which is the same window it occupied before and for the same two reasons: after `ar qc` so one pass reaches every member, before the smoke so the C link is measured against the archive that ships.

A pure move over an edit is how a block like that disappears silently, so I did not trust the rebase and re-ran the end-to-end check on a real build.

```
python3 acadsharp/build_acadsharp.py --platform musl --arch arm64
bash acadsharp/scripts/verify_archive_matched_host.sh bin/acadsharp-musl-arm64.tgz musl arm64
```

```
[musl/arm64] privatise_unwind: renamed 79 libunwind symbols in 2 archive(s)
[musl/arm64] privatised_unwind_symbols	79
[musl/arm64] retained_module_sections	1
[musl/arm64] static_ok	1
...
  the bundled libunwind is private (119 symbols), so rustc's own copy
  can sit beside it
  the encapsulation sections are retained, so --gc-sections cannot collect them
  static_certified holds: the archive links and runs here
  the library reads AC1014 to AC1032, which is what LINKINFO.json says
    ABI_VERSION=1 ABI_FINGERPRINT=a7805641d1a3b64f
    DWG_VERSION_MIN=1014 DWG_VERSION_MAX=1032
  the documented cargo recipe links and runs against this archive

verify_archive.sh: all invariants hold for acadsharp-musl-arm64.tgz
```

**79 and 119 are both unchanged** from before the rebase, which is the number I would have told you about rather than adjusted. #72's new read-range probe runs beside mine on the same archive, so the two lanes compose.

## `shellcheck acadsharp/scripts/stage.sh` is clean with the block in it

That file was a Python string until #72 and no linter had ever seen it. It is discovered by `git ls-files '*.sh'` now, so my block is linted for the first time as part of this push:

```
$ bash tools/shellcheck-all.sh
acadsharp/scripts/link_consumer_smoke.sh
acadsharp/scripts/privatise_unwind.sh
acadsharp/scripts/stage.sh
acadsharp/scripts/verify_archive.sh
acadsharp/scripts/verify_archive_matched_host.sh
...
SHELLCHECK-OK
```

Clean, no `# shellcheck disable` added, and `${RUNTIME_ARCHIVES:?}` and `${STATIC_SYSTEM_LIBRARY_LADDER:?}` are untouched, so the file is still valid shell as it sits in the tree.

## What else the rebase touched

**`verify_archive.sh`'s invariant list.** #72 added the read-range check as invariant 8 and I had also written an 8. Mine is 9 now, and the `VIPRS_REQUIRE_LINK_TEST` sentence stayed on 7 where it belongs. Both invariants are in the file and both fire.

**`test_verify_archive.py`'s fixture builder.** #72 gave it a capabilities unit (`with_caps`, `caps_objs`); I gave it an unwinder and a `plat`. The resolution keeps all four, so the static archive is now `[obj, register_obj, unwind_obj] + caps_objs` and the shared library is `with_caps(symbols, shared_src)` built under `shared_library_name(plat)`.

**Two signatures, in `f15755d`.** `stage_script()` lost its platform argument and `_write_build_context()` lost its plat. The tests follow.

**One test is new rather than adjusted.** `stage_script()` reads a file off disk now, so every other assertion in `TestTheBuildRunsTheRename` is only as good as that file. `test_the_stage_script_in_the_tree_is_the_one_that_runs` pins that what the function returns is `acadsharp/scripts/stage.sh` itself, so the next move of that file cannot quietly turn the rest of the class into assertions about nothing.

**`privatise_unwind.sh` is in the build context and the Dockerfile.** #72 rewrote both, so this needed redoing rather than merging: the copy loop is now `("retain_sections.py", "stage.sh", "privatise_unwind.sh")` with a chmod loop over the two shell scripts, and the `COPY` line carries it.

Full suite after the rebase: 1537 passed, 8 skipped. ruff clean.

Issue #74, opened 2026-09-13
Deferred from the #59 parallel wave: three things we chose not to do, and why


spdrman:
Three pieces of work came out of the five-lane wave on #59 and #65 that were deliberately not done, each for a reason worth keeping. Everything else those lanes found was fixed in the PR where it was found. These are the exceptions, so they get a home rather than living in a PR comment nobody will scroll back to.

## 1. An empty view has no warning code, so a consumer cannot tell empty from damaged

Raised in #71, declined in #70, and the decline was right.

A consumer reading a view with no records cannot tell a drawing that is genuinely empty from one whose contents failed to flatten. That wants a warning code of its own.

It could not go in with the rest, because `WIRE.md` cannot document a code before something emits it. The repository enforces that in both directions:

```
tests/test_warning_codes.py::test_the_document_defines_no_code_nothing_emits
  WIRE.md defines {extra}, which nothing in this library produces. A specified
  code with no producer is a branch a consumer writes and never reaches, and it
  stops anyone noticing when the real one is missing.
```

So this is one change, not two: the producer in the shim and the `WIRE.md` row together, plus a fixture that produces an empty view and a scenario recording the code.

- [ ] a warning code for an empty view, emitted and documented in the same change

## 2. The conformance consumers never run against a musl shared library

Raised in #73, and scoped out of it deliberately.

#73 makes the static cargo recipe work on musl and proves it with a link and a runtime probe that interleaves managed exception handling with Rust unwinding. What it does not do is run the C and Rust conformance consumers against a musl **shared** library, so musl has linking coverage and no behaviour coverage.

`test_conformance_workflow.py` asserts each runner appears exactly twice, against the unpacked archive and against the test configuration, and that pairing is deliberate, so adding a third is a real change to the workflow's shape rather than a line.

- [ ] run both conformance consumers against a musl shared library, and decide what the runner-count rule becomes

## 3. The reader allocates from sizes the file declares, before any bound applies, and there is no patch mechanism

Raised during the #59 triage. This is the one with teeth.

`getSectionBuffer18`/`21` call `HugeMemoryStream.Create(DecompressedSize x count)` and the LZ77 decompressor allocates `new byte[decompressedSize]`, all straight from header fields. `max_input_bytes` does not reach any of it, which is the amplification the adversarial review measured at 387x. A hostile or corrupt file can ask for an allocation before a single limit is consulted.

[...]

PR #75, opened 2026-09-13
Run the conformance consumers against the musl shared library

spdrman:
The second item in #74: the conformance consumers never ran against a musl shared library.

## What was true

#73 made the static cargo recipe work on musl and proved it with a link and a runtime probe that interleaves managed exception handling with Rust unwinding. It stopped there on purpose. So `musl-static-link` link-tested the archive and never ran a consumer against it, and the musl `.so` we ship had never been asked to answer the ABI handshake, report its capabilities or refuse a malformed file. Linking coverage, no behaviour coverage.

The shape rule stood in the way of adding one. `test_conformance_workflow.py` asserted each runner appeared exactly twice, and the two were never the number: one run against what a consumer downloads, one against the only configuration carrying `viprs_acad__test_throw` and `viprs_acad__test_live_handles`. A count cannot say that, and it refuses a third run that is a genuine gap.

## What this does

One run per consumer against the unpacked musl archive, in `rust:1.98.1-alpine` on the same arm64 runner, inside the job that already builds and verifies that archive. No second AbiTest publish: what musl can answer differently is whether the `.so` loads and answers at all, and the exception and handle-leak cases are managed logic that cannot differ by libc, so a musl AbiTest build would cost a second NativeAOT publish for no musl-specific answer. The runners already skip those sections against a Release archive and that is what they do here.

The image is the glibc job's image one libc over, deliberately. If the two containers also differed by compiler version, a check that passes on one and fails on the other would no longer say which of the two it was, and there is a test holding that.

The rule becomes "at least twice, and against a musl lib dir at least once", written as positive assertions rather than a loosened count: each runner has a lib dir that is an unpacked archive, one that is the AbiTest build, and one that is the musl archive of the cell that job builds, derived from the build step rather than spelled out again.

## Red, then green

The test change went in first, against the unmodified workflow:

```
FAILED test_conformance_workflow.py::TestTheMuslLibraryIsRunAndNotOnlyLinked::test_both_consumers_run_against_the_musl_archive
FAILED test_conformance_workflow.py::TestTheMuslLibraryIsRunAndNotOnlyLinked::test_the_musl_consumers_run_in_a_musl_container
FAILED test_conformance_workflow.py::TestTheMuslLibraryIsRunAndNotOnlyLinked::test_both_libcs_run_the_same_toolchain
3 failed, 24 passed in 0.16s
```

with the messages they are supposed to have:

```
E  AssertionError: acadsharp/tests/conformance/c/run.sh never runs against
   acadsharp-musl-arm64 in musl-static-link: []. musl then has a link test and
   no behaviour test, which is the state #73 left and #74 records
E  AssertionError: musl-static-link runs its consumers in '', which is not a
   musl image. A musl .so asks for /lib/ld-musl-*.so.1 and a glibc container
   has no such loader, so this is the difference between running the musl
   library and failing to
E  AssertionError: the workflow names ['rust:1.98.1'], which is not one glibc
   image and one musl image
```

Green after the workflow commit, and the whole suite with it, in a container that has gcc so the compiled-fixture tests actually run rather than skipping:

```
27 passed in 0.10s          # test_conformance_workflow.py
2146 passed, 10 skipped     # the repo
```

## The positive control

A run that quietly used a glibc library would look exactly like a run that used the musl one, so there is a step ahead of both that reads the library's own NEEDED entries. Against the musl archive, on arm64:

```
.../bin/unpacked-musl/acadsharp-musl-arm64/lib/libacadsharp_native.so needs:
  libc.musl-aarch64.so.1
```

Pointed at an unpacked glibc archive instead, the same step fails:

```
.../glibc/acadsharp-linux-arm64/lib/libacadsharp_native.so needs:
  libm.so.6
  libc.so.6
  ld-linux-aarch64.so.1
the library about to be run does not ask for a musl libc,
so the runs below would prove nothing about musl
identity step rc=1
```

And the consumer itself cannot be fooled either. The C runner, same Alpine container, pointed at the glibc `.so` built from this same tree, does not reach the program:

```
aarch64-alpine-linux-musl/bin/ld: .../libacadsharp_native.so: undefined reference to `pthread_rwlock_rdlock@GLIBC_2.34'
aarch64-alpine-linux-musl/bin/ld: .../libacadsharp_native.so: undefined reference to `__errno_location@GLIBC_2.17'
collect2: error: ld returned 1 exit status
rc=1
```

155 undefined references, every one of them a GLIBC-versioned libc symbol and none of them a `viprs_acad_*` entry point, so the refusal is about the libc and nothing else. That last part is worth stating because the first time I ran this control I used an archive left over from an earlier branch, which also failed, partly for a reason that had nothing to do with libc: it predates the ABI v2 rename, so it was missing `viprs_acad_get_capabilities_v1` as well. An archive built from this tree gives the clean version.

The real runs, locally, arm64, against the musl archive built from this tree:

```
library: .../bin/unpacked-musl/acadsharp-musl-arm64/lib/libacadsharp_native.so
image:   rust:1.98.1-alpine (linux/arm64)
test exports:  absent, so the cases that need them are skipped
166 checks, 0 failures      # C consumer
145 checks, 0 failures      # generated Rust consumer
```

## What the first musl run found

The generated Rust consumer did not run at all the first time, and the way it failed is worth keeping.

A musl toolchain links `crt-static` by default. This crate links a shared library, so the result has to be a dynamic executable, but rustc omits `-dynamic-linker` under `crt-static` and `ld` falls back to its own aarch64 default of `/lib/ld-linux-aarch64.so.1`, which Alpine does not have. The binary builds and links cleanly and the kernel then refuses to exec it:

```
error: could not execute process /tmp/target/debug/conformance (never executed)
Caused by:
  No such file or directory (os error 2)
```

which reads like a missing build far more than like a missing loader. `readelf -p .interp` had it:

```
before:  /lib/ld-linux-aarch64.so.1     NEEDED: libviprs_acadsharp.so
after:   /lib/ld-musl-aarch64.so.1      NEEDED: libviprs_acadsharp.so
                                                libgcc_s.so.1
                                                libc.musl-aarch64.so.1
```

`acadsharp/tests/conformance/rust/run.sh` now asks for a dynamic musl when the toolchain host is a musl one, appended to any `RUSTFLAGS` a caller set rather than replacing them. That belongs in the runner and not in the workflow, because anyone running this by hand on Alpine hits the same wall. The consumer then needs `libc.musl-aarch64.so.1`, the same libc as the shim it is testing, which is a second piece of evidence that the two are in the same world.

## Three smaller things that came with it

- The `VIPRS_CONFORMANCE_IMAGE` checks read the first match of a regex. That said nothing once there were two images, so they read every match now.
- The verify-before-run ordering was global across the file, which the glibc job satisfied on its own no matter what any other job did. It is per job.
- A job that runs a consumer without naming `VIPRS_CONFORMANCE_IMAGE` falls back to a local-only image name. That is a test failure now rather than a pull error in CI.

## Cost

The musl archive build already dominates that job. The additions are an unpack, one `readelf` in a container, and the two runs. Locally with the image warm both runs are about a second each (the crate has no dependencies, so cargo finishes in 0.8s); on the runner the added cost is dominated by pulling `rust:1.98.1-alpine` once. The job's 90 minute budget is untouched and no second NativeAOT publish was added, which was the point of the reduced shape.

Refs #74


spdrman:
## Green on the runner, and what the musl runs actually printed

Run [34779596250](https://github.com/libviprs/libviprs-dep/actions/runs/34779596250), both jobs success, on the tip of this branch. `CI` (ruff, pytest, shellcheck) is green on the same commit.

The positive control, on the runner rather than on my machine:

```
/home/runner/work/libviprs-dep/libviprs-dep/bin/unpacked-musl/acadsharp-musl-arm64/lib/libacadsharp_native.so needs:
  libc.musl-aarch64.so.1
```

and then the two runs against exactly that file:

```
Run the C consumer against the unpacked musl archive
  library: .../bin/unpacked-musl/acadsharp-musl-arm64/lib/libacadsharp_native.so
  image:   rust:1.98.1-alpine (linux/arm64)
  test exports:  absent, so the cases that need them are skipped
  166 checks, 0 failures

Run the generated Rust consumer against the unpacked musl archive
  library: .../bin/unpacked-musl/acadsharp-musl-arm64/lib/libacadsharp_native.so
  image:   rust:1.98.1-alpine (linux/arm64)
  libc: musl, so the consumer is linked against a dynamic one
  145 checks, 0 failures
```

## The cost, measured rather than estimated

Step times off the two runs, `musl-static-link` on `main` at `9e21857` against this branch:

| step | main | here |
| --- | --- | --- |
| Build the musl/arm64 archive | 75s | 88s |
| Verify it on a musl host | 16s | 15s |
| Unpack it the way a consumer would | | 0s |
| Name the libc the library asks for | | 0s |
| Run the C consumer | | 1s |
| Run the generated Rust consumer | | 2s |
| **job total** | **96s** | **110s** |

So the four new steps measure 3 seconds between them, on a job with a 90 minute budget. The 13 seconds the archive build moved is runner variance and not something I added.

Two reasons it lands that cheaply. `verify_archive_matched_host.sh` already pulls `rust:1.98.1-alpine` in the step immediately before, so the image is warm and the pull costs nothing extra, which is a second reason to use that image rather than pick another one. And the Rust consumer crate has no dependencies, so cargo builds it in under a second.

## The glibc side is untouched, and I checked rather than assumed

`acadsharp/tests/conformance/rust/run.sh` now sets a RUSTFLAGS entry, which is a shared file. The gate on it is the toolchain host triple, so on `rust:1.98.1` the case does not match and nothing changes. The `consumers` job proves that end to end: both of its archive runs and both of its AbiTest runs are green on this branch.

## What this does not cover

musl/arm64 only. The musl/x64 cell still has link coverage and no behaviour coverage, because it lives in `release-acadsharp.yml` on an x64 runner and this workflow deliberately runs one architecture with no emulation, for the reason ADR 0001 gives. Same shape would work there, and it is a separate change to a separate file.

## One thing I got wrong first

My first pass at the negative control used an `acadsharp-linux-arm64.tgz` left over from an earlier branch. It failed, which is what I wanted, but 11 of its undefined references were `viprs_acad_*` entry points rather than libc, because that archive predates the ABI v2 rename. A control that fails for two reasons proves neither. Built fresh from this tree, all 155 undefined references are GLIBC-versioned libc symbols and none are entry points, so the refusal is about the libc and nothing else. That is the version quoted in the description.

PR #76, opened 2026-09-13
A fixture cannot lie about the ABI, and a skipped compiled fixture is not a pass

spdrman:
The fourth item on #74: a test fixture is allowed to lie about the ABI, and a skipped compiled-fixture suite reads as a pass. Refs #74.

## What was true

`test_verify_archive.py` builds a real stub library, and its `_stub_source` answered `return 1u` for `viprs_acad_abi_version` while the shipped header declares `VIPRS_ACAD_ABI_VERSION 2u`. #69 fixed the capabilities-call instance by pulling the name and the two version macros out of the header, but the generic arm could still disagree, and `STUB_DWG_MIN`/`STUB_DWG_MAX` were hand-recorded as `1014`/`1032`. #72's guard cannot cover any of this: it compares the **generated** smoke against the header, and a fixture is not generated from the header at all.

Then I went looking for who would have caught it downstream, and nobody would have. The verifier's probe prints `ABI_VERSION=` and throws the number away. It holds the *manifest* to the header (invariant 4) and it holds the *manifest* to the library's read range (invariant 8), but it never held the *library* to the header. So an archive whose `viprs_acad_abi_version()` answers 1 next to a header declaring 2 passed every check `verify_archive.sh` has. That is the same class, one layer out, and I folded the fix in here.

### 1. The skip gate

Disabled the gate (`toolchain.required()` returns `False`) and put one suite back to a bare `pytest.skip("cargo is not installed on this host")`:

```
FAILED test_verify_archive.py::TestASkippedCompiledFixtureIsNotAPass::test_the_flag_turns_it_into_a_failure
FAILED test_verify_archive.py::TestASkippedCompiledFixtureIsNotAPass::test_ci_is_an_environment_that_claims_it_can_compile
FAILED test_verify_archive.py::TestEveryToolchainGateGoesThroughTheDoorway::test_no_suite_skips_on_a_missing_build_tool_by_hand
E       AssertionError: these gates skip on a missing build tool without going through
E       toolchain.missing_tool(), so VIPRS_REQUIRE_COMPILED_FIXTURES cannot see them and
E       the tests behind them vanish from a green run:
E           test_build_link_consumer.py: "cargo is not installed on this host"
3 failed, 4 passed, 132 deselected in 0.53s
```

### 2. The header-agreement guard

Reintroduced the literal, exactly as asked. `_stub_source` back to `return 1u`:

```
E       assert 'uint32_t viprs_acad_abi_version(void) {\n\treturn 2u;\n}' in '#include <stdint.h>\n...'

    def test_the_stub_answers_what_the_header_declares(self, tmp_path, good_tree):
        answers = _probe_the_library(str(tmp_path), good_tree)
        assert int(answers["entry_points"]) == len(ENTRY_POINTS)
>       assert int(answers["abi_version"]) == STUB_ABI_VERSION, answers
E       AssertionError: {'entry_points': '11', 'abi_version': '1', 'abi_fingerprint': '0502ac0f61611530', 'caps_abi_version': '2', ...}
E       assert 1 == 2

FAILED ...::test_the_generated_stub_states_the_headers_abi_version
FAILED ...::test_the_stub_answers_what_the_header_declares
2 failed, 3 passed, 134 deselected in 0.76s
```

`'abi_version': '1'` next to `'caps_abi_version': '2'` is the bug itself, in one line: the same library answering two different numbers to the same question because #69 fixed one of them and not the other.

### 3. The verifier holding the library to its header

Removed the new comparison from the probe and ran the refusal tests against an archive whose library answers `abi_version 7` and `wire_version 7` to a header declaring 2:

```
E           static_certified holds: the archive links and runs here
E           the library reads AC1014 to AC1032, which is what LINKINFO.json says
E             ABI_VERSION=2 ABI_FINGERPRINT=0502ac0f61611530
E             DWG_VERSION_MIN=1014 DWG_VERSION_MAX=1032
E             agreed: abi_version 7 and wire_version 7 are what the shipped header declares
E           cargo recipe: not run (no cargo on this host)
E
E         verify_archive.sh: all invariants hold for acadsharp-linux-arm64.tgz
E       assert 0 == 1

FAILED ...::test_an_abi_version_the_header_does_not_declare_is_refused
FAILED ...::test_capabilities_versions_the_header_does_not_declare_are_refused
2 failed, 1 passed, 136 deselected in 0.99s
```

Green, repo-wide, in `viprs-dep-gate-cc:arm64` (gcc and ar present), with ruff and shellcheck:

```
2159 passed, 6 skipped in 21.29s
All checks passed!
82 files already formatted
ALL-GREEN
```

## The skip count, with and without a compiler

Same tree, same command, two images:

| image | compiler | result |
|---|---|---|
| `viprs-dep-gate-cc:arm64` | gcc, ar | `1849 passed, 6 skipped` |
| `viprs-dep-gate:arm64` | none | `1729 passed, 126 skipped` |

120 tests vanish, and before this branch nothing said so. Now the run ends with:

```
!!!!!!!!!!!!! 125 compiled-fixture tests did not run on this host !!!!!!!!!!!!!!
     1  cargo is not on this host, so the consumer link this suite is about never runs
   115  cc is not on this host, so the fixture here is a real ELF built on this host
     9  cc is not on this host, so the fixtures here are real ELF objects
  these read a real ELF built by cc and ar. Set VIPRS_REQUIRE_COMPILED_FIXTURES=1 to make this a failure instead of a line.
1729 passed, 126 skipped in 4.69s
```

And with the flag set in the image that has no compiler, which is the proof that it bites:

```
3 failed, 1729 passed, 1 skipped, 122 errors in 4.64s

cc is not on this host, so the fixture here is a real ELF built on this host.
VIPRS_REQUIRE_COMPILED_FIXTURES is set (or this is CI), which says this host builds the
compiled fixtures, so a skip here is the suite silently dropping the tests that read real
bytes rather than a host that cannot run them. Install cc, or set
VIPRS_REQUIRE_COMPILED_FIXTURES=0 to go back to skipping.
```

Local Mac runs keep skipping quietly. This machine has no host toolchain by design and a gate that failed here would make the suite unrunnable on the machine it is written on.

## How the negative fixtures stay exempt

Not by a list of names, which is the next thing to drift. By construction: there is no knob a fixture can pick up by accident.

`_build_linux_tree` takes `abi_version` and `caps_versions`, both defaulting to the header's own numbers. A tree built without naming them agrees with the header, full stop. The two trees that disagree say so in the call that builds them:

```python
return _build_linux_tree(str(tmp_path_factory.mktemp("abilie")), abi_version=STUB_ABI_VERSION + 5)
```

The existing negative fixtures (`tree_missing_a_shared_export` and friends) disagree about *symbols* and *initialisers*, which are entirely different knobs, so they never touch this and need no exemption at all. And the lying values are derived too (`STUB_ABI_VERSION + 5` rather than `7`), so the day the header reaches 7 those fixtures are still wrong rather than accidentally right.

## What is in here

- `acadsharp/tests/toolchain.py`, new. `missing_tool()` is the one doorway a missing build tool goes through, and the terminal-summary hook that prints the census. The gate defaults **on** when `CI` is set rather than needing a workflow to opt in, because a flag that has to be added is the flag a new workflow forgets, which is how this happened. `VIPRS_REQUIRE_COMPILED_FIXTURES` overrides in both directions.
- `acadsharp/tests/conftest.py` loads it by path and registers the hook, the same way it already loads the build driver.
- `test_privatise_unwind.py` and `test_build_link_consumer.py` route their tool gates through the doorway. Platform skips stay skips: a mac cell genuinely cannot produce an ELF and no flag changes that.
- `test_verify_archive.py`: the numbers derived, the guard probe, the two lying fixtures, and a source-rule test that reads every suite here for a hand-rolled skip on a missing build tool, in the style `test_shim_source_rules.py` already uses. That one has a control that plants an offender, because a scan whose pattern never matches reports a clean suite forever.
- `scripts/verify_archive.sh`: invariant 10, the library held to the header it ships, through both the entry point and the capabilities call.


spdrman:
Three things worth calling out separately, two of which I found on the way and folded in rather than filing.

**I did not edit a workflow, and I do not think this one needs to.** The brief was to set the flag in CI, and the obvious way is an `env:` line on the `test` job in `.github/workflows/ci.yml`. I went the other way: `toolchain.required()` derives its default from `CI` being present in the environment, so every GitHub Actions run gets the strict behaviour without anything being added to a workflow file. The reason is the failure mode itself. `ci.yml` already carries a comment about `pyyaml` being missing and 13 tests skipping on every run while looking green, and this item exists because a container without a compiler did the same thing to 114 more. A flag a workflow has to remember to set is a flag the next workflow forgets. If you would rather have it stated explicitly as well, this is the whole diff and it is inert given the derivation:

```yaml
      - run: pytest -v
        env:
          VIPRS_REQUIRE_COMPILED_FIXTURES: "1"
```

One consequence to be aware of before merging: five `test_build_link_consumer.py` tests currently skip for want of `cargo`. `ubuntu-latest` ships Rust, so they should run and pass in CI, but if any runner image ever drops it, those five become a failure rather than a skip. That is the intended behaviour, and I would rather it be visible here than discovered later.

**The verifier had the same hole the fixture did, and I fixed it here.** This started as "the fixture is allowed to lie", and the obvious next question was what would have caught the lie if the fixture had shipped. Nothing would. `verify_archive.sh` holds the *manifest* to the header's `VIPRS_ACAD_ABI_VERSION` (invariant 4) and holds the *manifest* to the library's AC10xx range (invariant 8), and its probe prints `ABI_VERSION=` and discards it. So the library's own answer was never compared to the header packed beside it in the same archive, and an archive whose `viprs_acad_abi_version()` said 1 next to a header declaring 2 passed everything. That is invariant 10 now, checked through the entry point and again through the capabilities call, because a consumer reads the pair out of the struct rather than out of the entry point and a library that agrees one way and disagrees the other is one half its callers refuse. The red run for it is in the PR body: `verify_archive.sh: all invariants hold` over an archive answering 7 to a header saying 2.

**The read range cannot come from the header, so it comes from `AbiConstants`.** The brief said to drive the DWG range from the header like the versions, and I could not: the header states the range only in prose (`so 1014 and 1032 for ACadSharp 3.7.1`), deliberately, and the verifier's own comment says so in as many words, which is exactly why it has to ask the running library rather than read the bytes. `native/Abi.cs` is the only machine-readable statement of it, so `_abi_constants_dwg_range()` reads it there, the same way `test_release_workflow.py` already reads `AbiConstants` for the manifest. `test_the_read_range_is_the_librarys_own_rather_than_a_pair_of_digits` pins that, so a rename in `AbiConstants` has to be followed here rather than absorbed.

One smaller thing: routing the skips through `missing_tool()` moves the reported skip location to `toolchain.py` instead of the call site. The reason string names the tool and the census lists every one of them, so I left it rather than reaching for tracebackhide.

spdrman:
A fourth instance of the same class, found while checking whether anything else here is allowed to disagree with the header. Pushed as `00fabe2`.

`TestEveryWireVersionDeclarationAgrees` pins the **wire** version across seven declarations, `native/Abi.cs` included, with a positive control on each regex. Nothing pins the **ABI** version at all. So bumping `VIPRS_ACAD_ABI_VERSION` in the header and leaving `AbiConstants.AbiVersion` behind is a change no test in this repository could see, and the fingerprint does not cover it: that hashes the header, and a stale `AbiConstants` is compiled from the same header, so the hashes match and the numbers do not. The library would answer 1 to a header declaring 2. That is this fixture's own bug, in the real shim.

Red, with `AbiVersion` moved to `3u` and the header left at `2u`:

```
E       AssertionError: AbiConstants.AbiVersion is 3 and the header declares
E       VIPRS_ACAD_ABI_VERSION 2. The library answers the first and every consumer
E       compiles against the second.
E       assert 3 == 2
FAILED ...::TestTheFixtureAgreesWithTheHeaderItShips::test_abi_constants_declares_the_headers_abi_version
```

I put it in `test_verify_archive.py` rather than in `test_conformance_consumers.py` next to the wire-version one, for two reasons. It keeps me inside the files this lane owns while two other lanes are live, and it sits directly beside `_abi_constants_dwg_range()`, which already reads `native/Abi.cs` for the same reason. If you would rather it lived with its wire-version sibling, it is a self-contained method and moves cleanly.

Worth saying where this now gets caught, because there are two layers and they are not redundant. Invariant 10 catches it at the archive, which is where it counts and where the pair actually shipped, but only on a host that can link for the target. This catches it at the source, where it is a one-line diff instead of a build, and it needs no compiler so it runs on the dev Mac too.

One related thing I did **not** touch, because `build_acadsharp.py` belongs to another lane this round: both generated smokes there (`archive_smoke_source` and `static_smoke_source`) print `ABI_VERSION=%u` and never compare it to `VIPRS_ACAD_ABI_VERSION` either. They check the fingerprint and stop. That is now covered downstream, since the driver runs `verify_archive.sh` on every archive it produces and invariant 10 is in there, so it is a redundancy rather than a hole. Worth a line in whichever lane owns that file next.

PR #77, opened 2026-09-13
Bound the reader's declared-size allocations, and give an empty view a warning code

### LINKED ISSUE #74: Deferred from the #59 parallel wave: three things we chose not to do, and why

spdrman:
Two items from #74, in one branch because they both rewrite the generated corpus and cannot be recorded independently: the reader's unbounded allocation (item 3) and the missing empty-view warning code (item 1).

Refs #74, which stays open: it has several PRs and Rom will retire it with the last one.

### The number in the issue is the benign one

#74 cites 387x. That is real and I reproduced it: `cancel/never` on `g13_many_inserts.dwg` allocates 69,298,944 bytes across the open of a 179,150-byte file, 386.8 times its size. It is managed churn across the whole open, not one buffer, and it is unchanged by this PR.

The adversarial ceiling is not a multiplier. I took `g13_line.dwg`, the smallest fixture in the corpus at 10,539 bytes, rewrote the four bytes of one page header's declared decompressed size, and measured what the open allocated. Before the patch:

| derivative | declared | alloc_open_bytes | x file | open |
| --- | --- | --- | --- | --- |
| page_map_64MiB | 67,108,864 | 67,157,576 | 6,372 | CORRUPT_INPUT |
| page_map_256MiB | 268,435,456 | 268,484,168 | 25,475 | CORRUPT_INPUT |
| page_map_1GiB | 1,073,741,824 | 1,073,790,528 | 101,887 | CORRUPT_INPUT |
| page_map_int32_max | 2,147,483,647 | 38,688 | 3.7 | OUT_OF_MEMORY |
| section_map_64MiB | 67,108,864 | 70,907,664 | 6,728 | **OK** |
| section_map_256MiB | 268,435,456 | 272,233,056 | 25,831 | **OK** |
| section_map_1GiB | 1,073,741,824 | 1,077,539,656 | 102,243 | **OK** |
| section_map_int32_max | 2,147,483,647 | 42,920 | 4.1 | OUT_OF_MEMORY |

Read the `section_map` rows first. A 10 KB file declaring a gigabyte allocated a gigabyte, **opened successfully and decoded**. The allocation tracks the declared field exactly and linearly. The ceiling is not a ratio at all, it is whatever fits in a 32-bit field, so the honest statement is that the allocation was unbounded from declared fields, and that is what I have written everywhere rather than 387x.

Two things fell out of the measurement:

- `peak_rss_kb` stayed at 36 MB while a gigabyte was allocated, because the pages are never touched. Anyone measuring this with RSS sees nothing. `GC.GetAllocatedBytesForCurrentThread` is the number that shows it.
- **The `OutOfMemoryException` catch at `AcadSharpSource.cs:186` is not inert.** #74's briefing says no scenario has ever produced `VIPRS_ACAD_OUT_OF_MEMORY`, and that was true, but the two `int32_max` rows produce it. `new byte[2147483647]` is above `Array.MaxLength` so the runtime refuses without allocating. That is the runtime declining, not a bound, and it is the one value in the whole 32-bit range where the old code was safe.

### The shim cannot bound any of it, so this builds the patch mechanism

`acadsharp/patches/` has held a `.gitkeep` and nothing else since it was created, and `build_acadsharp.py` had no step that would apply anything in it. It does now, shaped like pdfium's: every `.py` in that directory is copied into the build context and applied to the unpacked tarball, after the CSUtilities clone (which removes and recreates `src/CSUtilities`) and before the publish.

One deliberate difference from pdfium's scripts, which print `WARNING: ... already patched?` and return 0 when an anchor has moved. Mine refuses. A patch that reports success without applying produces an archive nobody can tell from a patched one, and that is the failure the whole mechanism would otherwise have.

`BUILDINFO.json` now carries `source_patches`, each patch's name and sha256, beside the existing source pins, and `verify_archive.sh` has the field in its frozen list so every published archive is checked for it. `regenerate.py` applies the same patches to its own ACadSharp checkout before building, so a capture is a run of the reader the archive ships rather than of the pristine tarball.

### The ceiling

`patches/allocation_ceiling.py` bounds every declared size the reader allocates from against **64 times the input stream's length, or 16 MiB, whichever is larger**. The factor comes from a measurement rather than a guess: I instrumented every allocation site and ran the whole corpus through it.

| | largest single allocation | x its own file |
| --- | --- | --- |
| `g13_wide_polyline.dwg` | 89,088 | **3.82** (the worst in the corpus) |
| `real_AC1018.dwg` | 1,306,624 | 1.19 |
| `real_AC1032.dwg` | 1,217,536 | 1.13 |
| `g13_dimension_deep.dwg` | 2,345,984 | 2.37 |

3.82x is the worst ratio anything asks for, and two of those files are drawings AutoCAD itself wrote. 64 is sixteen times that. The 16 MiB floor is there because the format's default page is 29,696 bytes, so a 10 KB drawing would otherwise be held to 640 KB.

Refusing is `CORRUPT_INPUT`, not `LIMIT_EXCEEDED`, and I think that is the right call: `LIMIT_EXCEEDED` means one of the six fields in `viprs_acad_limits_v1` and invites the caller to raise one and retry, and there is no field here that would change the answer. A file whose own header describes a section larger than the file could hold is malformed. It also needs no new mapping: `InvalidDataException` reaches the shim's existing reader catch-all.

After the patch, all eight derivatives refuse:

```
page_map_1GiB        alloc_open=39528 (3.8x)  CORRUPT_INPUT
  System.IO.InvalidDataException: this drawing declares 1073741824 for its LZ77 page,
  past the ceiling of 16777216 this build derives from an input of 10539 bytes
section_map_1GiB     alloc_open=43760 (4.2x)  CORRUPT_INPUT
```

1,077,539,656 bytes down to 43,760. And `declared/unmodified_control`, the same file with nothing rewritten, still opens and decodes, which is the pair that says the ceiling is not simply set below the corpus.

### What it does not do, said in ABI.md

The ceiling is per allocation and it scales with `max_input_bytes`. A caller allowing the default 512 MiB input is allowing up to 32 GiB for a single section buffer. This makes allocation a function of a number the caller sets; it does not make it small, and `max_input_bytes` is the field that moves it. That is written into the limits section rather than left for someone to discover.

### Beyond the sites #74 names

Two more in the same class, guarded here and called out because neither is reachable by a fixture, so they ship as defence rather than as something I proved:

- `DwgReader.cs:918`, the local section page loop, runs `PageCount` times where `PageCount` is an unvalidated `Int32`, and a `DecompressedSize` of 0 makes `while (decompressSizeCounter < localmap.Offset)` an infinite loop appending objects. Guarded by `PageCount * DecompressedSize <= ceiling`, a page-size floor of 1, and a page count ceiling of 2^20 (the largest in the corpus is 79).
- the same loop's gap fill, where a declared `Offset` is a loop count.

Both live inside the LZ77-compressed section map, so reaching them needs a re-encoded section map rather than a byte edit, which is why the fixtures stop at the two plain page headers. If you want that proven rather than guarded, it wants its own issue.

## Item 1: `EMPTY_VIEW`, code 108

Option A as decided: one code, fired when a view emits zero geometry records.

The detection site is **not** `Flattener.cs`. It is `DecodeSession.Stream()`, now `Compose()`, which already tracks `_items` and `_warnings`. It is the difference and not the count: the reader's notifications head every view's stream whether or not the view holds anything, so a drawing that could not be read at all arrives with four items and no geometry among them. `_items - _warnings == 0` is the test.

`ViewEnd`'s `record_count` was `_items + 3` and the new record is between `ViewBegin` and `ViewEnd` like any other, so the warning counts itself into `_items` and `_warnings` rather than being yielded past them. Both conformance consumers verify `record_count` against their own count, so an unadjusted one would have failed there.

### What moved in the corpus, and what did not

The new fixture is `g13_empty_view.dwg`, the default document with no entity in it. `Corpus.cs:827` says `g13_bad_extents` adds a line on purpose "so the view is not empty", so nothing in the corpus covered this, and nothing could have covered it by accident.

Then three fixtures that were already there turned out to produce no geometry at all, each for a different reason, and each says what that reason was in the same stream:

| fixture | gains | alongside |
| --- | --- | --- |
| `g13_empty_view.dwg` (new) | EMPTY_VIEW | nothing but reader notifications |
| `g13_unsupported.dwg` | EMPTY_VIEW | UNSUPPORTED_ENTITY |
| `g13_xref.dwg` | EMPTY_VIEW | UNRESOLVED_BLOCK |
| `g13_nan_bulge.dwg` | EMPTY_VIEW | NON_FINITE_GEOMETRY |

That is the whole distinction the code exists for, arriving with a worked example of each side, and it is the entire expectation churn: three files gain one line, one file is new. Every other dump is byte for byte what it was, which I checked against both the patched and unpatched reader before recording anything. The benchmarks are not re-recorded: they are pinned to fixture digests that did not move, and re-running them would churn run-to-run RSS numbers for no signal.

`g13_scenarios.json` is the exception and its diff is large, but almost none of it is a change in behaviour. The whole file is one recording, so re-running it to add the twelve new scenarios moved every existing scenario's `alloc_open_bytes`, `peak_rss_kb` and `decode_micros`, which are what a managed runtime does between two identical runs. Every `open_code` and `decode_code` in it is what it was.

One mechanical change was needed. The canonical dump walked `source.EnumerateView` directly, a layer below where this warning is composed, so the dump was the one artefact that could not see it. It walks a fresh `DecodeSession.Compose()` now and drops the framing records, which is why every existing expectation is unchanged.

## Folded in, from the fixtures lane's finding

Both generated smoke programs printed `ABI_VERSION=` and never compared it to anything: the same value-read-then-discarded shape the verifier had. Both compare it now, and the shared one asks through the capabilities struct as well as the export, because an earlier round fixed one of those paths and not the other. `header_versions()` had its default argument bound at import, which made the static smoke's substituted version look like it tracked the header while being impossible to prove; it is resolved on the call now, and there is a test that moves the header and watches the number follow.

## Red, then green

Item 3, the mechanism, before any of it existed:

```
FAILED test_acadsharp_patches.py::TestThereIsAPatchToApply::test_the_driver_knows_where_the_patches_live
FAILED test_acadsharp_patches.py::TestThereIsAPatchToApply::test_the_directory_holds_at_least_one
FAILED test_acadsharp_patches.py::TestThereIsAPatchToApply::test_every_named_script_is_in_the_tree
FAILED test_acadsharp_patches.py::TestTheDockerfileAppliesThem::test_the_patch_is_copied_into_the_build_context
FAILED test_acadsharp_patches.py::TestTheDockerfileAppliesThem::test_the_patch_runs_against_the_unpacked_source
FAILED test_acadsharp_patches.py::TestTheDockerfileAppliesThem::test_the_patch_runs_after_the_submodule_clone
FAILED test_acadsharp_patches.py::TestTheDockerfileAppliesThem::test_the_patch_runs_before_the_publish
FAILED test_acadsharp_patches.py::TestTheDockerfileAppliesThem::test_the_build_context_carries_them
FAILED test_acadsharp_patches.py::TestTheArchiveSaysWhatWasApplied::test_buildinfo_carries_the_patch_list
FAILED test_acadsharp_patches.py::TestTheArchiveSaysWhatWasApplied::test_it_names_every_patch_with_its_digest
FAILED test_acadsharp_patches.py::TestTheArchiveSaysWhatWasApplied::test_the_verifier_reads_the_same_field
FAILED test_acadsharp_patches.py::TestAPatchThatDoesNotApplyIsARefusal::test_an_empty_tree_is_refused
FAILED test_acadsharp_patches.py::TestAPatchThatDoesNotApplyIsARefusal::test_a_tree_with_the_right_files_and_the_wrong_contents_is_refused
13 failed, 1 skipped
```

Item 1, with the `WIRE.md` row and the constant in and the producer out, which is the guard #74 quotes:

```
FAILED test_warning_codes.py::TestTheSpecificationAndTheShimAgree::test_the_document_defines_no_code_nothing_emits
E  AssertionError: WIRE.md defines [108], which nothing in this library produces. A
   specified code with no producer is a branch a consumer writes and never reaches,
   and it stops anyone noticing when the real one is missing.
1 failed, 10 passed
```

There is a second guard I did not know about until it caught me, `test_flatten_warnings.py::TestEveryCodeHasAFixture`, which insists every declared code has a fixture behind it. It went red on `EMPTY_VIEW` before the fixture existed.

Green, whole repository, in a container that has gcc, because without one the compiled-fixture tests skip and a skip reads as a pass:

```
2239 passed, 5 skipped
ruff check: All checks passed!
ruff format --check: 84 files already formatted
tools/shellcheck-all.sh: ok
```

Baseline on `main` for `acadsharp/tests` alone was 1834 passed, 6 skipped; it is 1925 passed, 5 skipped now.

I also built the real `linux-arm64` archive end to end so the new Dockerfile step is not shipped untested, and compiled both generated smoke programs with `-Wall -Wextra`.

This clears the last of #74's checkboxes: items 2 and 4 landed in #75 and #76, and this PR carries items 1 and 3.

Closes #74


spdrman:
Three notes for whoever reviews this, and one thing I want a decision on rather than a merge.

**How to reproduce the measurement.** The eight derivatives are not committed. They are derived at capture time from `g13_line.dwg` by `g13_support.declared_size_inputs()`, which rewrites the four bytes after a page header's magic, and `test_limits_declared_sizes.py` recomputes the derivation and checks the sha256 the capture recorded, so the capture is provably a run of a file anybody can rebuild. The magic has to appear exactly once or the derivation refuses, which is what makes it a rewrite rather than a guess at which field was meant. To see the before, unpack the pinned tarball without applying `patches/allocation_ceiling.py` and build the fixture generator against it: `regenerate.py` applies the patch now, so you have to go round it deliberately.

**Two traps that cost me time, in case they cost anyone else any.**

`peak_rss_kb` is useless for this. It sat at 36 MB while a gigabyte was allocated, because nothing ever touches the pages. Anyone auditing this with RSS or with `time -v` concludes there is no problem. `GC.GetAllocatedBytesForCurrentThread` is the only number in the harness that shows it, and it happens to be the one the path-versus-memory work added for an unrelated reason.

The other is that a single fixture can only ever prove the first guard it reaches, because the reader refuses at the earliest one. That is why there are two headers and not one: `page_map` exercises `DwgReader.cs:831` and `section_map` exercises `:875`, and the `section_map` ones are the interesting rows because those open **successfully** after allocating a gigabyte. The `page_map` ones refuse afterwards for an unrelated reason, so on their own they look like the reader coping.

**What I did not prove, and would like a tracked issue for rather than a quiet merge.** Two guards in this patch cover sites no fixture reaches:

- `descriptor.PageCount` and `descriptor.DecompressedSize` in the AC18 descriptor loop, where `PageCount` is an unvalidated `Int32` driving a loop and a zero `DecompressedSize` makes the gap fill `while (decompressSizeCounter < localmap.Offset)` run forever;
- `localmap.Offset` in that same gap fill, which is a declared field used as a loop count.

Both fields live inside the LZ77-compressed data section map, so reaching them needs a re-encoded section map rather than the four-byte edit the other fixtures use. Writing an LZ77-AC18 encoder and a synthetic section map is a real piece of work and it is not this PR. The guards are cheap, they cost nothing on the corpus, and I would rather ship them than leave the sites open, but they are defence and not evidence and I have said so in the patch's own comments too. If you want them proven, that is an issue of its own.

Everything else in the branch has a failing state I watched before the fix, and the two red runs are in the description.

PR #79, opened 2026-09-13
Run the musl/x64 library, not only link it, and settle where the fixture gate is turned on

spdrman:
Two coverage gaps that earlier lanes named and left. Both live in workflow files.

## A. musl/x64 had a link test and no behaviour test

#75 put both conformance consumers in front of a musl shared library, in `acadsharp-conformance.yml`, which runs arm64 and nothing else on purpose. Its own note said what was left: musl/x64 is built in `release-acadsharp.yml`, on an x64 runner, and had never once been asked to load and answer.

`build-linux` now unpacks the archive it just verified, proves the library's libc, and runs both consumers against it, between Verify and Upload so a library that loads and then refuses the handshake keeps its archive off the release page.

No `if:` narrowing that to the one cell. linux/x64 had exactly the same gap and nobody had named it, the arm64 cells re-ask the question against the tarball that is about to be uploaded rather than against a branch build, and a rule with no exception cannot be got wrong by editing the matrix. Each cell carries the image its consumers run in, which is `rust:1.98.1` in that cell's libc, so a check that passes on one side and fails on the other still says which of the two it was.

### Red, then green

Tests first, against the unedited workflows:

```
FAILED ...TestEveryContainerCellIsRunAndNotOnlyLinked::test_both_consumers_run_in_the_job_that_builds_the_archives
FAILED ...TestEveryContainerCellIsRunAndNotOnlyLinked::test_the_consumers_are_pointed_at_the_archive_the_cell_unpacked
FAILED ...TestEveryContainerCellIsRunAndNotOnlyLinked::test_the_archive_is_unpacked_before_a_consumer_looks_for_it
FAILED ...TestEveryContainerCellIsRunAndNotOnlyLinked::test_every_cell_names_a_container_whose_libc_is_the_cell_s
FAILED ...TestEveryContainerCellIsRunAndNotOnlyLinked::test_the_job_takes_that_image_from_the_cell
FAILED ...TestEveryContainerCellIsRunAndNotOnlyLinked::test_both_libcs_run_the_same_toolchain
FAILED ...TestEveryContainerCellIsRunAndNotOnlyLinked::test_the_image_is_pinned_and_is_not_the_local_default
FAILED ...TestEveryContainerCellIsRunAndNotOnlyLinked::test_the_container_architecture_follows_the_cell
FAILED ...TestTheLibcOfTheLibraryUnderTestIsProved::test_the_control_is_there
FAILED ...TestTheLibcOfTheLibraryUnderTestIsProved::test_it_runs_before_the_consumers_it_is_vouching_for
FAILED ...TestTheLibcOfTheLibraryUnderTestIsProved::test_it_reads_the_library_the_consumers_will_load
FAILED ...TestTheLibcOfTheLibraryUnderTestIsProved::test_it_asks_about_the_cell_it_is_in
FAILED ...TestTheLibcControlCanFail::test_the_script_is_there_and_executable
FAILED ...TestTheLibcControlCanFail::test_a_musl_library_satisfies_a_musl_cell
FAILED ...TestTheLibcControlCanFail::test_a_glibc_library_satisfies_a_glibc_cell
FAILED ...TestTheLibcControlCanFail::test_a_glibc_library_does_not_satisfy_a_musl_cell
FAILED ...TestTheLibcControlCanFail::test_the_musl_job_uses_it_rather_than_its_own_copy
FAILED ...TestTheLibcControlCanFail::test_the_control_runs_before_the_consumers_it_vouches_for
18 failed, 135 passed
```

then `153 passed` on the same two files, and `2203 passed, 5 skipped` repo-wide with `CI=true` in a container that has gcc **and** cargo, so nothing in this run is a compiled fixture that quietly did not compile. The five skips are shellcheck and ruff, both of which I ran separately (shellcheck rc=0 over 14 scripts, `ruff check` and `ruff format --check` clean).

Three of those tests passed vacuously in the red run, because a script that does not exist exits 127 and 127 is not 0. They are real now, and two more that were green throughout I watched fail by mutation: an `if: matrix.platform == 'musl'` on the consumer step and a musl cell given a glibc image.

```
FAILED ...::test_nothing_conditions_the_consumers_onto_a_subset_of_the_matrix
FAILED ...::test_every_cell_names_a_container_whose_libc_is_the_cell_s
2 failed, 115 passed
```

### The positive control, and the proof it really used the musl library

A musl `.so` and a glibc `.so` are both ELF, `find_shim.sh` finds either, and a consumer linked against either reports the same passes. So before the consumers there is a step that reads the library's own `NEEDED` entries and refuses anything that is not this cell's libc.

#75 had that control inline, in one step of one job. It was right, and it had never refused anything, because the only way to run it was to run the job. It is `acadsharp/tests/conformance/expect_libc.sh` now, both workflows call it, and it decides both ways.

I built both x64 archives on the NAS, which is native x86_64, from this tree with this driver, and ran exactly the steps the workflow runs:

```
================ cell musl/x64 on x86_64 ================
library: .../bin/unpacked/acadsharp-musl-x64/lib/libacadsharp_native.so
the library asks for:
  libc.musl-x86_64.so.1
libc: musl, which is the one this cell is about
---- C consumer ----      166 checks, 0 failures
---- Rust consumer ----   145 checks, 0 failures

================ cell linux/x64 on x86_64 ================
the library asks for:
  libm.so.6
  libc.so.6
  ld-linux-x86-64.so.2
libc: linux, which is the one this cell is about
---- C consumer ----      166 checks, 0 failures
---- Rust consumer ----   145 checks, 0 failures

================ the control, pointed the wrong way ================
the library asks for:
  libm.so.6
  libc.so.6
  ld-linux-x86-64.so.2
Error: the library about to be run does not ask for libc.musl-<arch>.so.1,
  so a musl consumer run against it would prove nothing about musl.
exit 1
```

`libc.musl-x86_64.so.1` is the answer that was wanted: that soname is written by the linker that built the library, and the glibc build of the same shim from the same tree says `libc.so.6` instead and gets refused.

### Runner time

Measured on the NAS with the images already present: the libc control 1s, the C consumer 1s, the Rust consumer 2s. About 4 seconds of new work per cell.

The image is a pull only on the glibc cells: the musl cells go through `verify_archive_matched_host.sh` one step earlier, which already pulls `rust:1.98.1-alpine`. `docker pull rust:1.98.1` took 30s on that machine, `rust:1.98.1-alpine` 24s. The job timeout goes 60 to 75 so a slow build plus the new work cannot turn into a cancelled release.

### What the same run found

`rust/run.sh` asks for a dynamic musl, because on aarch64 the crt-static default produces a binary the kernel refuses to exec and cargo reports `(never executed)`. x86_64 does not behave that way. I took the case back out and ran it against the musl/x64 archive:

```
test result: ok. 4 passed; 0 failed; ...
VIPRS CAD ABI conformance, generated consumer
--- handshake
Segmentation fault (core dumped)
exit 139
```

It links, it execs, all four layout tests pass, and then it dies on the first call into the library, because what is running holds two libcs. Same missing flag, and nothing in the symptom points at it, so `run.sh` now records the x64 half beside the aarch64 half. With the flag, 145 checks and no failures.

## B. The strict-skip gate stays implicit, and ci.yml says so

#76 left this as a decision: the gate derives from `CI` being present, and the alternative was one `env:` line on ci.yml's test job.

I did not add the line. In a container with gcc and no cargo, on this tree:

```
no CI, no flag (a dev machine)                 2166 passed, 6 skipped
CI=true (what Actions sets)                    5 failed, 2166 passed, 1 skipped
VIPRS_REQUIRE_COMPILED_FIXTURES=1, no CI       5 failed, 2166 passed, 1 skipped
CI=true + the env: line #76 offered            5 failed, 2166 passed, 1 skipped
CI=true + the opt-out a cell without cc needs  2166 passed, 6 skipped
```

Row four is row two. The line changes nothing an Actions run does, and what it would add is a second statement of one rule, in a file where the only value worth turning it to is the one that switches the rule off.

What it would genuinely have bought is visibility: there was nothing in ci.yml saying the run is strict, and no way to find out short of reading a module three directories away. A comment buys that too and cannot be given a value, so ci.yml now says where the strictness comes from and names the override a runner with no compiler would need.

Pinned as tests rather than as this paragraph: an Actions run is strict, no workflow quietly opts out, and ci.yml explains it without also setting it. Red first on the third:

```
FAILED ...::TestCiYmlSaysWhereTheStrictnessComesFrom::test_it_names_the_switch
FAILED ...::TestCiYmlSaysWhereTheStrictnessComesFrom::test_it_names_the_file_that_decides
3 failed, 9 passed
```

and the opt-out guard watched to fail, by putting `VIPRS_REQUIRE_COMPILED_FIXTURES: "0"` on the pytest step:

```
E   AssertionError: these workflows switch the compiled-fixture gate off:
E       ci.yml:62: VIPRS_REQUIRE_COMPILED_FIXTURES: "0"
FAILED ...::TestNoWorkflowOptsOutOfIt::test_no_workflow_turns_the_gate_off
FAILED ...::TestCiYmlSaysWhereTheStrictnessComesFrom::test_it_says_so_without_also_setting_it
2 failed, 10 passed
```

`TestTheGateActuallyBites` spends a subprocess on the consequence rather than on `required()`: a suite needing a tool no host has is `1 skipped` without CI, `1 failed` with it, and a skip again with the opt-out.

### The cargo exposure

Five `test_build_link_consumer.py` tests want cargo, so under the gate they are a failure rather than a skip on any runner image without Rust. `ubuntu-latest` ships Rust, and this PR's own CI run is the check: those five run there rather than skipping. The day an image drops it they go red here instead of going quiet, which is the intended behaviour and is why I would rather it be visible than discovered.

While measuring this I got it wrong once, and the mistake is worth recording. My first run said the derivation did not engage under `CI=true`, which would have sunk the whole decision. zsh does not word-split an unquoted `$e`, so `docker run $e` passed `-e CI=true` as one argument and the variable was named with a leading space. Re-running before diagnosing is what caught it.

## Scope

`.github/workflows/**`, the workflow tests, and two files under `acadsharp/tests/conformance/`: the new `expect_libc.sh`, and a comment in `rust/run.sh`. Nothing here touches `build_acadsharp.py`, `acadsharp/patches/`, `native/`, `docs/`, `scripts/verify_archive.sh` or the generated corpus.

`release-acadsharp.yml` only fires on a push to `release` or a dispatch, so CI on this PR does not exercise it. The evidence for that half is the NAS run above plus the workflow tests. `acadsharp-conformance.yml` does fire on this push, so the `expect_libc.sh` swap in the musl job is exercised for real.

Refs #74


spdrman:
Four things worth calling out separately, three of which I found on the way and folded in rather than filing.

**CI is green on both halves, and one of those runs is the evidence for item B.** `CI` finished `2207 passed, 1 skipped` on ubuntu-latest, with no compiled-fixture census line, which is the check that the five `test_build_link_consumer.py` tests ran rather than skipped. So the cargo exposure #76 warned about is not live on the runners this repo uses: `ubuntu-latest` ships Rust, those five execute under the strict gate, and they go red here on the day an image drops it. That is the intended behaviour and it is why the gate is worth having on by default.

`acadsharp conformance` also ran, because this push touches `acadsharp/**`, so the `expect_libc.sh` swap in the musl job is exercised for real rather than only by a test:

```
Name the libc the library asks for
  the library asks for:
    libc.musl-aarch64.so.1
  libc: musl, which is the one this cell is about
Run the C consumer against the unpacked musl archive       166 checks, 0 failures
Run the generated Rust consumer against the unpacked musl  145 checks, 0 failures
```

**The libc control had never refused anything, and now it can be made to.** #75's version was inline in one step of one job. It was correct, and the only way to run it was to run the job, so nothing had ever pointed it at a glibc library to watch it say no. Pulling the decision into `expect_libc.sh` was not tidying: item A needed the same control in a second workflow, and the choice was one guard both call or two copies that drift. It now decides both ways, which the inline one did not: a musl library fails a glibc cell too, an empty `NEEDED` list is a refusal rather than a pass (a `readelf` whose output never arrived prints exactly that), and a platform it has no expectation for is a refusal rather than a shrug. `TestTheLibcControlCanFail` runs all six cases without a container.

**x86_64 musl fails differently from aarch64, and the symptom does not point at the cause.** `rust/run.sh` asks for a dynamic musl because on aarch64 the crt-static default leaves the binary with no interpreter and the kernel refuses to exec it, which cargo reports as `(never executed)`. I took the case back out on a native x86_64 host and ran it against the musl/x64 archive: it links, it execs, `cargo test --tests` passes all four layout cases, and then it dies on the first call into the library with a plain `Segmentation fault (core dumped)`, exit 139. What is running holds two libcs, its own static one and the one the shim asks for. Same missing flag, no hint of it in the symptom, so `run.sh` now records the x64 half beside the aarch64 half rather than leaving the next person to find it twice.

**A dispatch could have asked the consumers an unanswerable question.** Both runners read the expected upstream version out of `acadsharp/VERSION`, which is right for a push, and this job can be dispatched with a `version` input that the Build step honours instead. A dispatch whose upstream half differed from the file would build one ACadSharp and run a consumer expecting another, and the cell would fail for a reason that has nothing to do with the archive. It cannot happen today, and that is a fact about the pin table rather than about the workflow: the preflight refuses an upstream with no `SOURCE_SHA256` entry and there is exactly one entry. I pinned the coupling instead of leaving it implicit, so it goes red on the day a second pin lands, with the instruction in the message. Red with `3.8.0` added beside `3.7.1`:

```
E  AssertionError: acadsharp/VERSION is 3.7.1 and the driver pins ['3.7.1', '3.8.0'].
   A dispatch can now name an upstream the conformance consumers in this same job do
   not expect ...
FAILED ...::test_only_the_upstream_the_consumers_expect_can_be_built
```

That test reads `build_acadsharp.py` and changes nothing in it, which keeps it inside this lane while the driver belongs to another.

**Two smaller things.** Shellcheck caught a real break rather than a style point: a comment I added inside `sh -euc '...'` contained an apostrophe, which terminates the single-quoted string, so `run.sh` would have failed at parse the next time it ran. And `release-acadsharp.yml` fires only on a push to `release` or a dispatch, so nothing in this PR's CI exercises it. The evidence for that half is the two x64 archives I built on a native x86_64 host from this tree with this driver, and the workflow tests. I did not dispatch the release workflow, because a dispatch publishes to the existing tag.

Issue #78, opened 2026-09-13
Two allocation-ceiling guards ship unproven: the descriptor loop and the gap fill


spdrman:
## Progress

The ceiling patch from #77 installs **four guard functions across eight call sites**. Three of those sites are proven by fixtures that were watched failing; five are not. I built this checklist by reading the patch against the recorded scenarios rather than from the PR description, and doing that turned up a third category the original text of this issue missed.

### Proven, and watched failing first

- [x] **The LZ77 page sink** (`Decompress`, `DwgReader.cs:831`/`:875`). Four scenarios, `declared/page_map_{64MiB,256MiB,1GiB,int32_max}`, each a four-byte edit to the plaintext page-header declared size. All refuse with `CORRUPT_INPUT` at 39,528 to 39,544 bytes. Before the patch the same edits allocated in step with the field: 64 MiB declared gave 67,157,576 bytes and 1 GiB gave 1,077,539,656 with the open succeeding.
- [x] **The AC18 section-buffer product** (`getSectionBuffer18`, `DwgReader.cs:355`), which is `DecompressedSize * LocalSections.Count`. Four scenarios, `declared/section_map_{64MiB,256MiB,1GiB,int32_max}`, refusing at 43,752 to 43,768 bytes.
- [x] **A benign control alongside them.** `declared/unmodified_control` opens `OK` at 3,802,200 bytes, so the refusals are not a guard that refuses everything.
- [x] **Independent corroboration of the gaps.** A second agent, working from a digest-verified pristine tree with no sight of the patch, concluded unprompted that the product, the `PageCount` inflation and the zero-page loop were what a single-field bound would leave. Same set, opposite direction.

### Not proven, and this is what the issue is for

- [x] **`CheckDescriptor`: `PageCount` and `DecompressedSize` in the AC18 descriptor loop.** Done in #81. Five forged AC1018 inputs, one per branch: page size at 64 MiB and at a gigabyte, page size zero, page count, and the product with both factors legal. Unpatched, 64 MiB allocated 70,079,672 bytes out of a 440-byte file and a gigabyte allocated 1,076,712,160. The zero case is recorded patched and never run without the guard, because its absence is a hang and `regenerate.py` has no timeout by design; the mechanism is cited instead. The page-count branch is the weak one and #81 says so: unpatched it throws `EndOfStreamException` rather than allocating, so it is defence in depth plus a better message.
[...]

PR #80, opened 2026-09-13
Forge an AC1021 drawing so the four AC21 ceilings can be watched refusing

spdrman:
Four of the eight guards `acadsharp/patches/allocation_ceiling.py` installs sit on
the AC1021 side of `DwgReader`, and none of them had ever been watched refusing
anything. This lands a fixture that reaches all four, records the refusals, and
records a control beside them.

Refs #78

## They were not cheap, and the reason is three separate things

#78 expected these to fall to the four-byte page-header rewrite that proved the
AC18 sites, applied to `real_AC1032.dwg`. I checked that before grinding, and it
does not work, for three reasons stacked on top of each other.

**`real_AC1032.dwg` is not an AC21 file as far as this reader is concerned.**
`readFileHeader` sends `AC1024`, `AC1027` and `AC1032` to `readFileHeaderAC18`,
and `getSectionStream` sends them to `getSectionBuffer18`. Only `AC1021` itself
reaches `readFileHeaderAC21`, `getPageBuffer` and `getSectionBuffer21`. So the
corpus has no AC21 fixture at all, rather than one without scenarios.

**There cannot be one that the generator wrote, either.** Upstream's `DwgWriter`
throws `CadNotSupportedException` for `AC1021`, so the tool that writes every
`g13_*.dwg` cannot produce the version these guards need.

**And the fields are not in plaintext.** `PagesMapSizeCompressed`,
`PagesMapSizeUncompressed` and `PagesMapCorrectionFactor` live in a 0x110-byte
metadata block that is LZ77 compressed and then spread across the 0x400 bytes at
file offset 0x80. No four-byte edit reaches them even given a file to edit.

## But the Reed-Solomon worry was wrong, and that is what made this possible

The reason these were expected to need an RS re-encode is that the AC21 maps are
described as Reed-Solomon coded. In this reader they are not.
`DwgReader.reedSolomonDecoding` is a stride-`factor` gather: it corrects nothing,
it detects nothing, and there is no codeword to invalidate. The CRCs the header
carries are read into locals and never compared against anything either. The AC21
LZ77 has a literal run, which is a store mode, and the byte permutation its `copy`
applies is a bijection.

Both transforms are therefore invertible and neither authenticates anything, so
the file can be written as a function of the numbers it is supposed to declare.
`acadsharp/tests/ac21_forge.py` does that, and the `copy` permutation is checked
to be a bijection at every length before it is used, because a mistranscription
there would be a fixture quietly carrying bytes nobody chose.

## Red, then green

Measured on a pristine 3.7.1 tree against the same tree with the patch applied.
`alloc_open_bytes`, not RSS: peak RSS sat at 38 MB while a gigabyte was committed,
which is the third time that trap has been reported in this repo.

| scenario | declared | unpatched | patched |
| --- | --- | --- | --- |
| `ac21_page_block_64MiB` | 71,601,705 | 138,755,776 | 39,736 |
| `ac21_page_block_1GiB` | 1,145,624,220 | **2,219,411,248** | 39,744 |
| `ac21_compressed_page_wrapped` | 4,294,967,288 | 44,112 | 39,784 |
| `ac21_decompressed_page_64MiB` | 67,108,864 | 67,154,320 | 40,072 |
| `ac21_decompressed_page_1GiB` | 1,073,741,824 | 1,073,787,280 | 40,080 |
| `ac21_section_buffer_64MiB` | 67,108,864 | 137,191,632 | 2,962,280 |
| `ac21_section_buffer_1GiB` | 1,073,741,824 | **2,150,470,760** | 2,962,136 |
| `ac21_control` | 255 | 2,987,672 | 2,975,248 |

The gigabyte page-map case is 2.07 GiB out of a 1,152-byte file, 1.93 million
times its own length, which is the largest amplification anything in this repo
has measured. It is that big because the one declared field is committed twice:
once as the block buffer the interleave is read into and once as the page decoded
out of it. Patched, all four refuse with `CORRUPT_INPUT` naming the site, the
declared size and the ceiling.

The 2.96 MB on the section-buffer rows is not the declared size leaking through.
It is what walking a whole AC21 header costs, and the control beside it allocates
the same 2.97 MB while declaring 255 bytes. `test_the_allocation_does_not_track_the_declared_size`
is the check that says so: a gigabyte of difference in the declared field moves
the allocation by 144 bytes.

## The compressed-page guard, and why its input looks odd

`Check(totalSize, ..., "AC21 compressed page")` is dominated by the guard before
it for every value that does not wrap. `lenght` is `ceil(totalSize / 239) * 255`,
so it is always larger than `totalSize`, and both are measured against the same
ceiling: the block-buffer guard always refuses first.

The one way through is the int cast. `(int)(totalSize + blockSize - 1L)` narrows a
long before the division, so a `totalSize` near 2^32 produces a small factor, not a
large one. At `0xFFFFFFF8` the block buffer comes out at zero, passes, and the
compressed page is the guard that refuses. Unpatched, that same file throws
`OverflowException` on `new byte[(int)totalSize]` going negative, so this is the
one row where the guard is not preventing an allocation. It is preventing an
arithmetic accident from being the thing that saves us, and the refusal says what
happened rather than "arithmetic operation resulted in an overflow". There is a
test pinning the inequality, so if upstream widens that cast the input stops
reaching the guard it is named after and this file says so rather than passing.

## The controls

`declared/ac21_control` is the same forge with nothing inflated. It gets past all
four ceilings, the section buffer is assembled out of a real page, and what refuses
it is `DwgHeaderReader` running off the end of 255 bytes of nothing. That is the
pair the four refusals need: a ceiling that simply refused every AC1021 file would
produce exactly the same captures without it.

`declared/real_ac1032_control` records the real 2018 drawing opening `OK` and
decoding `OK` at 14,152,736 bytes, and a test asserts that none of the four AC21
site names appears anywhere in its capture. That is the correction to #78 made
into a check rather than a reading of a switch statement.

## How I know the baseline was unpatched

Not the shared builder image. `acadsharp-builder-3.7.1-viprs.1-linux-arm64`
legitimately carries the patch now and an earlier agent nearly measured a "before"
against it.

I unpacked two trees from a tarball whose sha256 I checked against
`SOURCE_SHA256["3.7.1"]` in `build_acadsharp.py`
(`0c6b9de91b4f41f355b8caa665bb9904be627a9288fefc6716de8a204a7847a7`), applied
`allocation_ceiling.py` to one of them, and grepped both for `ViprsAllocationCeiling`:
zero files in the pristine tree, three in the patched one. Both belts. The
`CSUtilities` submodule the tarball leaves empty came from an existing checkout,
and I verified the two independent copies on this machine are byte identical to
each other before using one.

## What this does not do

The two AC18 items on #78, `CheckDescriptor` and `CheckPageOffset`, are untouched.
Those still want a re-encoded AC18 LZ77 section map. The uncatchable-death path is
also still open. This ticks one box.


spdrman:
Three things I found on the way that are not the headline, folded into this PR rather than into new issues.

**The AC21 maps are not integrity checked at all, and that is a statement about the reader, not just about this fixture.** `DwgReader.reedSolomonDecoding` is named after an error-correcting code and is a stride-`factor` gather. It has no syndrome, no parity, no rejection path: whatever bytes are in the file come out de-interleaved. The 5 CRC-shaped fields `readFileHeaderAC21` pulls out of the decoded header, and `HeaderCRC64`, `PagesMapCrcCompressed`, `SectionsMapCrcUncompressed` and the rest of the metadata block, are read into locals and properties and never compared against anything. The `Checksum` and `CRC` on every AC21 page entry are the same. So for an AC1021 file the page map and the section map are attacker-controlled structures with nothing standing between them and the allocator except the ceiling this PR is proving. That is the reason the forge works, and it is also the reason the forge was the right amount of effort rather than a week of Galois field arithmetic. I have not filed it separately; it belongs to whoever next looks at what else those maps drive.

**`DwgLZ77AC21Decompressor._copyMethods[31]` looks wrong and is not.** It reads `src[0..30]` and writes `dst[0..30]` in five pieces with a two-byte reversal at the end, which is unlike every other entry in the table, and the first thing I did was assume I had mistranscribed it. I had not. It is a bijection, like all 31 of them, and `copy_permutation` asserts that at every length before it writes a byte, because a mistranscription there would have produced a fixture that silently carried different bytes from the ones its digest claims.

**`acadsharp/docs/ABI.md` now carries the AC21 number.** The section on what no field in `viprs_acad_limits_v1` bounds had the AC18 measurement, a hundred thousand times the file's own length. The AC21 one is 1.9 million times, and the doc said nothing about that path existing. It does now, including the sentence that matters for anyone auditing this from the switch statement: AC1024 and later are on the AC18 path, so the AC21 sites are not reachable from anything in the corpus.

On the local gate: `2365 passed, 6 skipped` in `viprs-dep-gate-cc:arm64`. The 6 are host tooling, 5 needing `cargo` and 1 needing `ruff`, both of which are on the runner, and none of them are in a file this PR touches.

PR #81, opened 2026-09-13
Forge an AC1018 drawing so the two descriptor ceilings can be watched refusing

### LINKED ISSUE #78: Two allocation-ceiling guards ship unproven: the descriptor loop and the gap fill

spdrman:
The last two guards from #77 that nobody had watched fail: `CheckDescriptor`
in the AC18 descriptor loop and `CheckPageOffset` in the gap fill it runs.
Both fire inside `readFileHeaderAC18`, at open, before a section is decoded,
and nothing in the corpus reaches either.

## Why a byte edit cannot get there, and why a forge can

The proven AC18 sinks were reached by rewriting four bytes of a page header
in `g13_line.dwg`. These two fields are not in a page header. They are in the
data-section map, which is stored LZ77 compressed, and the page header in
front of it is the only part of that page in the clear. `test_the_fields_
these_reach_are_not_in_plaintext` checks that against the fixture rather than
leaving it as a claim: compression type 2, compressed size not equal to
decompressed size.

So the file is written rather than derived, the way `ac21_forge.py` writes the
AC1021 side. It is cheaper than it sounds, because nothing between those bytes
and the descriptor loop authenticates anything:

- The 0x6C bytes at 0x80 that `DwgReader` calls encrypted are XORed with a
  fixed keystream. `CRC32StreamHandler`'s array constructor derives it from
  `randSeed = randSeed * 0x343fd + 0x269ec3` seeded with 1, taking bits 16 to
  23, so it is the same 108 bytes in every DWG ever written and 8 lines of
  Python reproduce it. `keystream` is checked against the magic sequence
  `DwgReader.cs:753` prints in its own comment before anything uses it.
- The file ID string inside that block is compared, and a mismatch raises a
  notification rather than an exception. Even the one check there is advisory.
- Every CRC and checksum in the page headers is read into a local and never
  compared against anything.
- `DwgLZ77AC18Decompressor` has a store mode, and unlike AC21's it has no
  permutation: `copy` is a `src.Read` straight into a `dst.Write`. A payload of
  1 to 238 bytes stores as `[len + 0x11]`, the bytes, `[0x11]`, and above that
  the `literalCount` continuation form has no ceiling at all.

`lz77_store` checks itself the way `ac21_forge.copy_permutation` proves its
bijection: every run it writes is decompressed again through `lz77_decompress`,
a transcription of `DecompressToDest`, and the bytes have to come back.
`read_back` then walks a finished file the way `readFileHeaderAC18` does,
through the header block, the page map and the section map, so a test can say
the number a capture names is the number a reader would find.

That oracle is also checked against data nobody here authored, and it needed
to be: my first transcription of `twoByteOffset` returned the second of its
two bytes where upstream returns the first, which is the byte that becomes the
next opcode. No input this forge writes enters the back-reference loop, which
is exactly why the part nothing reaches was the part that was wrong. So
`g13_line.dwg`'s own page map and section map, written by upstream's writer
with a dictionary and back references, now go through it and have to come out
at the lengths their page headers declare: 120 and 1,380 bytes. The section map
then parses under the same layout the forge writes, adds up to its own length,
and comes out as twelve descriptors whose `AcDb:Header` declares the 0x7400
page size the format uses by default. That is the layout check as well as the
decompressor check.

## Red, then green

The unpatched build is the pinned tarball, not the shared builder image: that
image already carries the ceiling and measuring a "before" against it measures
the fix. I unpacked `v3.7.1.tar.gz`, checked its sha256 against
`SOURCE_SHA256["3.7.1"]` (`0c6b9de9...7847a7`), cloned CSUtilities at
`fdf1403e`, and grepped the whole tree for `ViprsAllocationCeiling` before
building: absent. The patched tree is a second unpack with `patches/` applied.

Signal is `alloc_open_bytes` throughout. `peak_rss_kb` is in the table because
it says something here, not because it is the measurement.

| input | bytes | unpatched alloc | patched alloc | refused at |
|---|---|---|---|---|
| `ac18_descriptor_page_size_64MiB` | 440 | 70,079,672 | 45,928 | descriptor page size |
| `ac18_descriptor_page_size_1GiB` | 440 | 1,076,712,160 | 45,952 | descriptor page size |
| `ac18_descriptor_page_size_zero` | 440 | hangs, never run | 48,776 | descriptor page size |
| `ac18_descriptor_page_count` | 440 | 50,576, `EndOfStreamException` | 45,920 | descriptor page count |
| `ac18_descriptor_product` | 472 | 28,136,544 | 48,832 | 3 pages of 8388608 |
| `ac18_page_offset_64MiB` | 440 | 78,993,144 | 46,176 | page offset |
| `ac18_page_offset_1GiB` | 440 | 1,219,322,360 | 46,192 | page offset |
| `ac18_page_offset_gap` | 440 | 260,525,624 | 46,280 | page offset gap |
| `ac18_control` | 1,510 | 2,994,576 | 2,995,152 | a text height, both times |

The 1 GiB offset case is 2.8 million times its own file length. The two
branches that take a size cost 45,928 against 45,952 and 46,176 against 46,192
for declared numbers a gigabyte apart, which is the claim one number cannot
make on its own.

Two of these say something the byte ceiling alone does not:

- **The product case.** 3 pages is nowhere near `MaxSectionPages` and 8 MiB is
  half the ceiling, so every branch in front of the multiplication passes it.
  Nothing but the product refuses that file, and unpatched it allocated 28 MB
  out of 472 bytes.
- **The gap case.** An offset of 16,000,000 against an eight-byte page size.
  The offset is under the ceiling, and so is the buffer the 2,000,000 empty
  pages it creates add up to (16,000,008). What is not under anything is the
  number of objects, and unpatched that cost 260,525,624 bytes and 289 MB
  resident. This is the one `MaxSectionPages` exists for, and it is why that
  bound is not the byte ceiling written twice.

And one says less, which is worth stating rather than letting the table imply
otherwise. **The page-count case is the weak one.** Two million pages of one
byte isolates the branch, because one byte is far under the ceiling and two
million of them still are, so nothing but the count can refuse it. But
unpatched it throws `EndOfStreamException` on the second loop iteration rather
than allocating, since the file declares two million page entries and carries
one, and a file that really carried them would be thirty-two megabytes, past
the ceiling it is trying to get around. That branch is defence in depth plus a
better message, the same shape #80 found in the AC21 compressed-page guard. It
is recorded anyway: the first version of this work had a coverage check that
asserted every branch of both guards had an input, against a set that quietly
left this one out, and a completeness claim that names its own exceptions is
not one.

## The zero case, and why it is cited rather than run

`while (decompressSizeCounter < localmap.Offset)` at `DwgReader.cs:929` adds an
empty page and then advances the counter by `descriptor.DecompressedSize`,
which is the only thing that advances it. At zero the condition never changes
and `LocalSections` grows until the process dies, so the absence of this guard
is a hang and not an allocation. `regenerate.py`'s `run()` is
`subprocess.run(cmd, capture_output=True)` with no timeout, so recording that
would be a recorder that never comes back. The refusal is recorded, the
mechanism is cited, and no scenario in this repository hangs. The forged input
declares an offset above zero on purpose, because a zero page size with a zero
offset does not hang, it divides by zero further down where `sizeLeft` is
taken, and `test_the_forged_input_declares_an_offset_the_counter_cannot_reach`
pins that.

## The control

The same forge with nothing inflated: a 1 KB page size, one page, and an offset
one page further in, so the gap fill runs, adds one empty page and stops.
`getSectionBuffer18` allocates two pages' worth, zero-fills the empty one,
decompresses the real one into the rest, and `DwgHeaderReader` then reads a
text height of zero out of 2 KB that is not a header. `ceiling` appears nowhere
in that detail and the open costs three million bytes, sixty times the largest
refusal, which is what says the file went past the guards rather than being
stopped by one. A ceiling that refused every hand-written AC18 map would
produce the other seven captures without this one.

`declared/unmodified_control` and `declared/real_ac1032_control` still open and
decode `OK` beside them, and neither guard's site string appears anywhere in
their captures.

## Two things worth folding in

**The AC18 declared size is resident, not merely committed.** #77 recorded RSS
sitting at 36 to 38 MB while a gigabyte was allocated, and that has been the
standing warning on this issue. It holds for the sinks that note was measured
at, and it does not hold here: unpatched, `ac18_descriptor_page_size_1GiB`
peaked at 1,094,932 KB and `ac18_page_offset_1GiB` at 1,229,160 KB. The
difference is `HugeMemoryStream.Create`, which for anything up to
`int.MaxValue` hands back `new byte[(int)length]` rather than a chunk list, and
that array is resident on this runtime. I have not chased why beyond that, and
it does not change the rule: `alloc_open_bytes` is still the signal, because
RSS would have said nothing at all about the product case or the gap case.

**`HugeMemoryStream.Create` returns a fixed-length stream, which is load
bearing for the forge.** The page map is read with
`while (decompressed.Position < decompressed.Length)`, and `Length` is the
declared decompressed size rather than the bytes written. Declare one byte more
than the payload and the loop reads a tail of zeroed records, every one
numbered 0, and `Records.Add(0, ...)` throws on the second. Every page header
this forge writes declares exactly what its payload decompresses to, and
`page_header`'s docstring says why.

## Recorded, run and checked

- `tests/fixtures/gen/regenerate.py --only scenarios` against the patched tree,
  last, after `git fetch origin`. `origin/main` had not moved off `e724c79`, so
  there was nothing to rebase onto.
- Every other scenario's `result` moved by timing, RSS and managed-KB noise,
  which is what a re-record does. I diffed every non-numeric field across the
  whole file: no code, detail, count or digest changed anywhere else.
- `MANIFEST.json` is untouched, as expected. `shim_digest()` covers
  `native/{Probe.cs, Abi.cs, Adapter/**, Sources/**, Wire/**}`, and none of this
  is in there. No committed `.dwg` either: the forged files are built at record
  time and only their sha256 lands in the capture.
- `pytest acadsharp/tests` in `viprs-dep-gate-cc:arm64`: 2143 passed, 6 skipped.
  The skips are the 5 cargo consumer-link fixtures and one "not a git checkout",
  both pre-existing and environmental in that image.

Closes #78


spdrman:
Seven things this turned up that were not in the brief, all folded into the
branch rather than filed:

1. **The AC18 header block is not decrypted by `DwgReader` at all.** It reads
   the 0x6C bytes raw and hands them to `CRC32StreamHandler`, whose array
   constructor is where the XOR lives, with a keystream from an LCG seeded with
   1. So the "encrypted data" the comment at `DwgReader.cs:748` describes is a
   fixed 108 bytes, identical in every DWG. `keystream` reproduces it and is
   checked against the magic sequence in that same comment.

2. **The AC18 literal run has no upper length.** The brief's `[len + 0x11]`
   form tops out at 238 bytes, which is enough for a section map but not for
   the control's 1 KB data page. An opcode with a zero high nibble goes through
   `literalCount`, which reads its own continuation bytes, so the store mode is
   unbounded. `lz77_store` uses the short form where it fits and that one
   otherwise.

3. **`HugeMemoryStream.Create` returns a fixed-length stream below
   `int.MaxValue`.** It hands back `new MemoryStream(buf, 0, len, true, true)`,
   so a page header's declared decompressed size is the decompressed stream's
   `Length`, not just its capacity. The page-map loop runs
   `while (Position < Length)`, so declaring one byte more than the payload
   makes it read a tail of zeroed records, all numbered 0, and
   `Records.Add(0, ...)` throws on the second. Every page header the forge
   writes declares exactly what its payload decompresses to.

4. **The standing "`peak_rss_kb` is useless here" note does not hold on this
   path.** It holds where #77 measured it. Here the 1 GiB cases peaked at
   1,094,932 KB and 1,229,160 KB unpatched, because of (3): that is a real
   `byte[]` and it is resident. `alloc_open_bytes` is still the signal, since
   RSS would have said nothing about the product case or the gap case, but the
   note is advice rather than a fact about every sink and the PR body says so.

5. **The `0x2ba84` offset in the issue is wrong.** It is 178,820, past the end
   of a 10,539-byte file. The real ones are `0x2660` and `0x2840`, each once.

6. **`CheckDescriptor`'s page-count branch is defence in depth.** Covered
   anyway, because the first version of this had a coverage check asserting
   every branch had an input against a set that quietly left that one out.

7. **My own `twoByteOffset` transcription was wrong** and nothing the forge
   writes would ever have caught it, which is the point. It is fixed and the
   oracle now has to decompress `g13_line.dwg`'s own upstream-written page map
   and section map to their declared 120 and 1,380 bytes, and the section map
   has to parse under the same layout the forge writes.

Nothing here needs a new issue. `MANIFEST.json` is untouched and no `.dwg` is
committed: the forged files are built at record time and only their sha256
lands in the capture.

Issue #90, opened 2026-09-13
SHAPE is refused, and its geometry is in an external SHX this decoder should not resolve


spdrman:
`SHAPE` is refused. 1 on `real_AC1032.dwg`.

A SHAPE is a glyph from a compiled SHX shape file, placed at a point with a size and a
rotation. It is the smallest gap on the refused list and the only one whose geometry is
not in the DWG at all: the outline lives in an external `.shx`, and the entity carries a
name and an index into it.

## What acadrust models

```rust
pub struct Shape {
    pub common: EntityCommon,
    pub insertion_point: Vector3,
    pub size: f64,
    pub shape_name: String,
    // ...
}
```

Which is the placement, and not the outline. acadrust does not render SHX either. Neither
implementation can produce the geometry from the file in hand.

## The honest position

This is a `wontfix` candidate rather than a backlog item, for the same reason as 3DSOLID
in #88: the work is not "flatten a SHAPE", it is "parse and evaluate the SHX format, and
find the font file the drawing refers to". That is an external-resource problem with the
same security shape described in the raster issue.

SHAPE is also rare in practice, largely superseded by blocks and by TrueType text.

## What to do

Record the decision. Either:

- Close as `wontfix`, with a sentence in `docs/` so the next reader does not re-derive
  it, or
- Emit a placement record carrying the insertion point, size, rotation and the shape
  name as an opaque string, and never resolve the file. A consumer holding the right
  SHX can then do something with it.

Doing nothing and leaving it in the default arm is the only outcome worth avoiding,
because it looks like an oversight rather than a judgement.

## Acceptance

- [ ] Either closed with the reasoning written down, or a placement record with no
      external resolution.

spdrman:
## Fixtures

**SHAPE is writable** (`DwgObjectWriter.Entities.cs`, `case Shape shape:`), which is mildly
surprising given the geometry lives in an external SHX, and it means a fixture is available
if this is ever taken past `wontfix`.

| Fixture | Source | Notes |
| --- | --- | --- |
| `g13_shape.dwg` | generated | One `new Shape` naming a shape that **no SHX in the environment provides**. |
| SHAPE in situ | `real_AC1032.dwg` handle `889` | Already refused and recorded. |

### The fixture's one job

Name a shape file that is not present. That is not a convenience, it is the assertion: this
issue's position is that the decoder emits a placement and never resolves the external
resource, and a fixture naming a shape the machine happens to have cannot distinguish a
[...]

Issue #88, opened 2026-09-13
3DSOLID, REGION and MESH are refused; MESH is tractable and the other two need ACIS


spdrman:
`3DSOLID`, `REGION` and `MESH` are refused. On `real_AC1032.dwg`: 2 3DSOLID, 1 REGION,
2 MESH.

They are grouped because two of them share one hard dependency and the third does not,
and that distinction is the useful part of this issue.

## 3DSOLID and REGION need ACIS

Both store their geometry as an embedded ACIS stream (SAT/SAB), which is a proprietary
boundary-representation format. ACadSharp surfaces the stream; it does not evaluate it.
Neither does acadrust. There is no tessellation to be had from either library, so
"flatten a 3DSOLID" means writing or vendoring a B-rep evaluator, which is a
categorically larger undertaking than anything else on the refused list.

**Recommendation: keep refusing these two, and record that as a decision** rather than
leaving them in the default arm as though nobody had looked. The honest position is that
a decoder built on ACadSharp cannot render them, and saying so is more useful than an
open issue implying it is merely unscheduled.

## MESH is different and is worth doing

A MESH is an explicit vertex list plus face indices, with an optional subdivision level.
No proprietary format, no evaluator: the faces are in the file. It lowers to `Polygon`
records, one per face, exactly as #84 proposes for SOLID and 3DFACE.

The subdivision level is the thing to decide. Emitting the base mesh is honest and
cheap; evaluating subdivision to match what AutoCAD displays is a smoothing algorithm
and a much weaker promise. Emit the base mesh and say the level was ignored.

## Acceptance

- [ ] MESH emits one `Polygon` per face, with the subdivision level ignored and said so.
- [ ] 3DSOLID and REGION keep warning, and a short note in `docs/` records that this is
      a decision about ACIS rather than a backlog item.
- [ ] A MESH fixture whose faces are not all coplanar, so a flattener that collapses
      them is red.

spdrman:
## Fixtures

The split in the body maps exactly onto what our writer can emit, which is a useful
coincidence.

| Kind | Writable? | Fixture |
| --- | --- | --- |
| MESH | **yes** (`case Mesh mesh:`) | `g13_mesh.dwg`, generated |
| 3DSOLID | **no** | `real_AC1032.dwg` handles `D65`, `D6A` |
| REGION | **no** | `real_AC1032.dwg` handle `D69` |

`Solid3D` and `Region` have no case in `DwgObjectWriter` and fall to the
[...]

Issue #86, opened 2026-09-13
MULTILEADER, LEADER and TOLERANCE are refused; 19 entities on the real drawing


spdrman:
`MULTILEADER`, `LEADER` and `TOLERANCE` are refused. On `real_AC1032.dwg` that is 15
MULTILEADER, 1 LEADER and 3 TOLERANCE, 19 of the 88 refused entities and the second
largest group after POINT.

They belong together: all three are annotation composites that expand into primitives
this version already emits, and DIMENSION has already established how that is done here.

## Why DIMENSION is the precedent and the warning

`Flattener.Walk` dispatches DIMENSION onto its own stack rather than expanding it inside
`Map`, and the comment there says why: expanding a composite from inside `Map` put a
file-controlled recursion on the CLR stack. Any work on these three inherits that
constraint. A MULTILEADER carrying an MTEXT block, on a drawing an attacker supplies, is
the same shape of problem.

## What each needs

**MULTILEADER** is the hard one and the most common. It carries leader lines, an
optional landing, an arrowhead, and content that is either an MTEXT or a block
reference. The text and the lines are already expressible; the arrowhead is a block and
goes through the same expansion INSERT does.

**LEADER** is the older, simpler form: a vertex list plus an annotation reference. The
vertices lower to a `Polyline` directly.

**TOLERANCE** is a feature-control frame: text in boxes, with the frame geometry derived
from the dimension style's text height. It lowers to `Line` and `Text`, but the frame
geometry has to be computed rather than read, which makes it the one most likely to
disagree with what AutoCAD draws.

## Suggested order

LEADER first: it is the smallest and it proves the expansion path outside DIMENSION.
Then MULTILEADER, which is where the 15 are. TOLERANCE last, and possibly not at all,
since computing frame geometry to match AutoCAD is a much weaker promise than reading
geometry a file already holds.

## Acceptance

- [ ] Expansion goes through `Walk`'s own stack, never recursion inside `Map`.
- [ ] A depth cap, matching the one block expansion already has.
- [ ] One fixture per kind, plus a nested case for MULTILEADER with block content.
- [ ] Whatever is not expanded still warns, rather than emitting partial geometry.

spdrman:
## Fixtures

The three split cleanly by whether our own writer can emit them.

| Kind | Writable? | Fixture |
| --- | --- | --- |
| LEADER | **yes** (`case Leader l:`) | `g13_leader.dwg`, generated |
[...]

Issue #85, opened 2026-09-13
RAY and XLINE are refused, and the wire format has no unbounded record to put them in


spdrman:
`RAY` and `XLINE` are refused. One of each on `real_AC1032.dwg`.

They are one piece of work: both are a base point and a direction, and both have
infinite extent. A RAY runs from its base point in one direction forever; an XLINE runs
both ways forever.

## What acadrust models

```rust
pub struct Ray   { pub common: EntityCommon, pub base_point: Vector3, pub direction: Vector3 }
pub struct XLine { pub common: EntityCommon, pub base_point: Vector3, pub direction: Vector3 }
```

## Why this one is a design decision and not a mapping

Every record in `docs/WIRE.md` today is bounded, and these are not. There is no honest
way to put an infinite line into a `Line` record, and the three ways out are not equal:

- **Clip to the drawing extents and emit a `Line`.** Produces something renderable, and
  quietly turns a construction line into a finite one. A consumer that later re-extends
  or re-frames the view gets a line that stops in mid-air for no visible reason. It also
  puts a decode-time dependency on extents, which `g13_bad_extents.dwg` exists precisely
  to show cannot be trusted.
- **A new unbounded record** carrying base point and direction, leaving the clip to the
  consumer that knows its viewport. Honest, and costs a wire change.
- **Keep refusing them,** but say so deliberately in the docs rather than by falling
  through to the default arm.

The middle option looks right, and the first is the one to be careful of: it is the
easiest to implement and the hardest to notice being wrong.

Construction lines are also frequently on a non-plotting layer, so a consumer may well
want to drop them anyway. That is a consumer's decision to make and needs the record to
exist before it can be made.

## Acceptance

- [ ] A decision recorded in `docs/`, with the extents-clipping trap named.
- [ ] If a record is added: `docs/WIRE.md` versioned, fixtures for RAY and XLINE.
- [ ] If they stay refused: the refusal is explicit rather than a fall-through.

spdrman:
## Fixtures

Both are writable (`DwgObjectWriter.Entities.cs`, `case Ray ray:` and `case XLine xline:`),
so this gets a real fixture.

| Fixture | Build | Why this shape |
| --- | --- | --- |
[...]

PR #94, opened 2026-09-13
Add an isolated .dwg fixture per refused entity kind

### LINKED ISSUE #93: Add an isolated .dwg fixture per refused entity kind, where the writer can produce one
#82 through #90 each describe a kind the flattener does not emit, and each of them can
only be probed today against `real_AC1032.dwg`, a 341-record drawing. That is a poor test
surface: a change to its expectation is hard to review, and a fixture that holds one kind
in isolation is what the rest of this corpus already provides for every kind the flattener
does handle.

So: one small DWG per refused kind, where the writer can produce one.

`DwgObjectWriter` can emit POINT, SOLID, RAY, XLINE, MESH, TOLERANCE, SHAPE, LEADER,
PolyfaceMesh, PolygonMesh and the wipeout/image base. It has no case for MULTILEADER,
MLINE, 3DFACE, 3DSOLID, REGION or PDFUNDERLAY, and the final arm throws
`NotImplementedException`, so those six cannot be generated here and keep using the real
drawing.

## Acceptance

- [ ] A fixture per generatable refused kind, written by `tests/fixtures/gen/`.
- [ ] Each is shaped so the plausible wrong implementation fails it, rather than being a
      file that merely contains the entity.
- [ ] `PROVENANCE.md` records them.
- [ ] Both ACadSharp and an independent reader open every one.

spdrman:
Seven `.dwg` files, one per refused entity kind the writer can produce, so #82 through #90
have a small input to probe instead of only `real_AC1032.dwg` and its 341-record
expectation.

| Fixture | Kind | Serves |
| --- | --- | --- |
| `g13_point.dwg` | 4 POINT | #83 |
| `g13_solid.dwg` | 3 SOLID | #84 |
| `g13_ray_xline.dwg` | RAY, XLINE, LINE | #85 |
| `g13_polyface_mesh.dwg` | POLYFACE_MESH | #82 |
| `g13_polygon_mesh.dwg` | POLYGON_MESH | #82 |
| `g13_mesh.dwg` | MESH | #88 |
| `g13_tolerance.dwg` | TOLERANCE | #86 |

## Each one is shaped so the plausible wrong implementation fails it

This is the rule `g13_ocs_*` already states, applied again. A file that merely contains
the entity is satisfied by any output at all, which converts "untested" into "covered".

- **`g13_solid`**'s first quad is **asymmetric**. DXF stores corners 3 and 4 swapped, so an
  implementation emitting 1,2,3,4 draws a bow-tie, and over a square a bow-tie passes any
  assertion on area or bounding box. It also carries the triangle case and a non-Z normal.
- **`g13_point`** carries a non-Z normal, because a point's location is in the plane its
  normal defines. The origin is included and deliberately not alone: it is a fixed point of
  every transform.
- **`g13_polyface_mesh`**'s six vertices are ordered so a line threaded through them in
  storage order **self-intersects and leaves the surface**. If storage order happened to be
  a sensible path, #82's defect would produce something that looked plausible.
- **`g13_ray_xline`** uses non-axis-aligned directions, because an axis-aligned
  construction line clipped to an axis-aligned extents box lands on fixed points. The
  bounded line is there to give the drawing finite extents at all.
- **`g13_mesh`** is at subdivision level 2 with non-coplanar faces, so an implementation
  that collapses the mesh or silently starts evaluating subdivision is red.
- **`g13_polygon_mesh`** is 3×4 rather than square, so an M/N transposition shows up.

## Six kinds are deliberately absent

MULTILEADER, MLINE, 3DFACE, 3DSOLID, REGION and PDFUNDERLAY. ACadSharp's
`DwgObjectWriter` has no case for any of them and its final arm throws
`NotImplementedException`, so this generator cannot write one. Those stay on
`real_AC1032.dwg`, and the handles are named in each issue.

## Both halves: top level and block-expanded

The first round of these held their entity at **top level only**, so every record any of
them could produce had `flags` 0. `docs/WIRE.md` defines bit 0 as set when a record came
from expanding a nested insertion, and nothing reached it. That let through the defect
that actually ships: the kind gets implemented, the top-level case works, the fixture is
green, and instances inside an INSERT are dropped or emitted untransformed. The corpus
already holds the *handled* kinds to a higher bar, since `g13_insert` and
`g13_ocs_rotated` carry `flags=1` records.

So each file now carries a second half: one block, inserted **twice**, and neither
insertion is the identity.

- One is **non-uniform**, the only thing in these files that reaches the flattener's
  `NonUniform` warning path.
- One is **mirrored**, because a mirror reverses what counter-clockwise means, and that is
  what decides a mesh's face winding and a solid's corner order.

A uniform upright insert would have exercised the flags bit and nothing else, which is the
same mistake one level up.

The four files that sat at the default +Z normal (polyface mesh, polygon mesh, mesh,
tolerance) gained a **non-Z** case too. At +Z the arbitrary-axis algorithm is exactly the
identity, so they could not tell an implementation that runs the lift from one that never
wrote it. Point and solid already had one.

## Deliberately not varied

Layer, colour, linetype, lineweight, thickness. The wire carries only `handle` and `flags`
from the common entity fields, and the flattener reads `Thickness` nowhere:

```
$ grep -rn "Thickness" native/Adapter/ native/Wire/
(nothing)
```

There is no field in any record for those to be wrong in, so setting them would grow the
corpus and assert nothing. `PROVENANCE.md` records the exemption, because it is the
direction somebody enriches in by reflex.

## Proven readable by both implementations

Written by ACadSharp's `DwgWriter`, then read back by acadrust 0.5.5, an unrelated
pure-Rust reader, at exactly the counts they were built with, two inserts per file:

```
g13_mesh           AC1032 OK n=5 [Insert:2 Mesh:2 Viewport:1]
g13_point          AC1032 OK n=9 [Insert:2 Point:6 Viewport:1]
g13_polyface_mesh  AC1032 OK n=6 [Insert:2 PolyfaceMesh:3 Viewport:1]
g13_polygon_mesh   AC1032 OK n=6 [Insert:2 PolygonMesh:3 Viewport:1]
g13_ray_xline      AC1032 OK n=8 [Insert:2 Line:1 Ray:2 Viewport:1 XLine:2]
g13_solid          AC1032 OK n=7 [Insert:2 Solid:4 Viewport:1]
g13_tolerance      AC1032 OK n=6 [Insert:2 Tolerance:3 Viewport:1]
```

That is the property that makes them useful here: one input, two implementations.

## Notes

No expectations are committed yet. These kinds are all refused today, so an expectation
would record warnings, and it should be written by whichever issue actually implements the
kind. `test_adapter_stream.py` already tolerates a fixture with no recording.

`PROVENANCE.md` carries all seven with their digests. Full suite green: 2002 passed, 126
skipped, the skips pre-existing.

Closes #93


spdrman:
Worth knowing before this merges: these seven are **thinner than they should be in one specific way**, and I have written the enrichment up on #92.

Every one of them holds its entity at top level only, so every record they will ever produce has `flags=0`. `docs/WIRE.md` line 144 defines bit 0 as set when a record came from expanding a nested insertion, and nothing here reaches it. The corpus already takes that seriously for the kinds the flattener handles (`g13_insert`, `g13_ocs_rotated` and others carry `flags=1` records); these do not.

The failure that lets through is the one that ships: a kind gets implemented, the top-level case works, the fixture is green, and instances inside an INSERT are dropped or emitted untransformed.

Two options, and I do not mind which:

- **Merge as is and enrich in a follow-up.** These are strictly better than having nothing, and nothing downstream depends on them yet.
- **Hold, and I add the block-expanded instances here**, plus a non-uniform and a mirrored insert, plus a non-Z normal on the four that currently sit at the default +Z.

The second keeps the corpus from carrying a known-thin round, which is the thing that tends not to get revisited. Say which and I will do it.

spdrman:
## Merging this, with four corrections folded in rather than filed

A four-expert panel reviewed this PR alongside the other twelve open issues. The
files are good and re-cutting them would only churn digests, so this lands as is.
Four things they measured are worth carrying, and I am putting them here rather
than opening issues.

**1. The claim about `DwgObjectWriter` is wrong, and five issues are scoped around
it.** This PR and #93 both say the writer "has no case for MULTILEADER, MLINE,
3DFACE, 3DSOLID, REGION or PDFUNDERLAY, and the final arm throws
`NotImplementedException`". Reading `DwgObjectWriter.Entities.cs` in the pinned
3.7.1, there are explicit cases with real write methods for `Face3D` (line 545),
`MLine` (563), `MultiLeader` (569), `PdfUnderlay` (575) and `Solid3D` (618).
`writeFace3D` writes all four corners plus the invisible-edge flags, for R13/R14
and for R2000+. Only **REGION** genuinely reaches the default arm, because
`Region : ModelerGeometry` and `Solid3D : ModelerGeometry` are siblings so
`case Solid3D` does not catch it.

So isolated fixtures are producible for five of the six kinds written off here,
and #84, #86, #87, #88 and #89 can each be held to a higher bar than "it stays on
the real drawing".

**2. The two mesh files' second copy proves nothing.** The body says the four +Z
files "gained a non-Z case too". For `g13_polyface_mesh` and `g13_polygon_mesh`
that did not happen: decoded, the +Z copy and the skew-normal copy are
geometrically identical and both report `normal=[(0,0,1)]`. The reason is in the
writer: `writePolyfaceMesh` and `writePolygonMesh` write no extrusion field at all,
which is correct, because POLYLINE_PFACE and POLYLINE_MESH have none in the DWG
object format. Only `writePolyline2D` writes one. A non-Z normal cannot round trip
for these two kinds by construction, so the second entity in each file is a
duplicate. Worth a line in `PROVENANCE.md` so nobody adds it back.

Related: `PROVENANCE.md` does not claim a non-Z case for `g13_mesh`, the PR body
does, and `Mesh` has no `Normal` at all.

**3. `g13_solid.dwg` discriminates, and the assertion has to be on the first
solid.** Corners `(0,0)`, `(10,1)`, `(2,5)`, `(11,7)`, checked by shoelace: the
correct 1,2,4,3 order is a simple quad of area 50, and the 1,2,3,4 bow-tie has
signed area 3.5. The third solid is the rectangle `(0,0),(4,0),(0,3),(4,3)`, whose
bow-tie has signed area exactly **0**, so a shoelace control there is degenerate.
Assert on the first.

**4. These seven files are read by nothing.** `test_adapter_stream.py` builds its
parametrisation from `MANIFEST["fixtures"]`, and this PR adds no manifest entry, so
the DWGs get no digest check, no size check and no expectation.
`test_no_dump_is_orphaned` catches a dump with no manifest entry; nothing catches
the other direction, which is why seven files can land outside every check. The
`sha256` values added to `PROVENANCE.md` are also verified by nothing: a grep of
the whole suite for `PROVENANCE` returns zero hits, which is exactly the
hand-carried digest `docs/LINKINFO.md` warns about.

Deferring the expectations is the right call and the PR argues it well: an
expectation of a refusal belongs to whoever implements the kind. The missing guard
is not. A test that fails when a `.dwg` under `tests/fixtures/` has no manifest
entry, with an explicit allow-list of "carried, not yet recorded" names, closes the
hole without pinning warnings nobody wants pinned. That is going into the campaign's
first lane along with a `PROVENANCE.md` digest check.

Issue #95, opened 2026-09-13
The macOS dylib's install name is @rpath/viprs_acadsharp.dylib and no such file ships, so a consumer links it and cannot load it


spdrman:
The macOS archive ships `lib/libacadsharp_native.dylib`, and that file's install
name is `@rpath/viprs_acadsharp.dylib`. No file of that name is in the archive,
so anything that links it records a name the loader cannot resolve and the
binary dies before `main`.

Measured on `acadsharp-mac-arm64.tgz` from `acadsharp-3.7.1-viprs.1`
(sha256 `b809dc5e8a488f6bb71a065386225be9e706225873a244969165c374b8c59347`):

```
$ ls -1 acadsharp-mac-arm64/lib/
libacadsharp_native.dylib

$ otool -D acadsharp-mac-arm64/lib/libacadsharp_native.dylib
acadsharp-mac-arm64/lib/libacadsharp_native.dylib:
@rpath/viprs_acadsharp.dylib

$ find acadsharp-mac-arm64 -name 'viprs_acadsharp*'
acadsharp-mac-arm64/include/viprs_acadsharp.h
```

`metadata/LINKINFO.json` says `"shared_library": "lib/libacadsharp_native.dylib"`,
which is the file that is really there, so a consumer that follows the manifest
links the right path and still gets the wrong recorded name.

## Why it matters more on mac than anywhere else

`aarch64-apple-darwin` is the one published target with
`"static_certified": false`, so the shared library is the **only** link mode
there. On Linux this cannot happen: the `.so` carries no `SONAME` at all, so the
linker records whatever filename it was handed. I verified that by linking and
running a probe against `acadsharp-linux-arm64.tgz` in a container, which works.

So today a consumer can link and run the published artifact on four of the five
targets, and on the fifth it links and then cannot start. That is the one case
EPIC G's definition of done does not cover, because nothing in the release
pipeline runs a mac binary.

## Root cause

`build_acadsharp.py` renames the published assembly on purpose, and says so at
line 184:

```python
# Names inside the archive. The publish emits viprs_acadsharp.so (the
# assembly name); the archive ships a name `-lacadsharp_native` can take.
SHARED_LIBRARY_STEM = "libacadsharp_native"
...
PUBLISHED_ASSEMBLY = "viprs_acadsharp"
```

The rename is right and should stay: `-lacadsharp_native` needs it. What the
[...]

Issue #89, opened 2026-09-13
IMAGE, PDFUNDERLAY and WIPEOUT are refused; dropping a WIPEOUT shows what should be hidden


spdrman:
`IMAGE`, `PDFUNDERLAY` and `WIPEOUT` are refused. One of each on `real_AC1032.dwg`.

All three are placed rectangles rather than strokes, and all three share the property
that makes them different from everything else on the refused list: **what they display
is not in the DWG.** An IMAGE names an external raster, a PDFUNDERLAY names an external
PDF, and both are resolved through a definition object that holds a file path.

## What acadrust models

```rust
pub struct Wipeout {
    pub common: EntityCommon,
    pub class_version: ...,
    pub insertion_point: Vector3,
    pub u_vector: Vector3,
    // ...
}
```

The same placement shape (insertion point, u and v vectors, clip boundary) covers
RasterImage and Underlay.

## Why WIPEOUT is the one to do first

A WIPEOUT is a mask: an opaque region that hides what is under it. Its geometry is
entirely in the file, it needs no external resource, and **skipping it is not a neutral
omission**. Every other refusal loses something that should have been drawn. Dropping a
wipeout shows things that should have been hidden, which is the failure direction that
produces a plausible-looking wrong drawing rather than an obviously incomplete one.

That makes it a correctness concern wearing a missing-feature coat, and it lowers to an
existing `Polygon` record plus a flag or convention saying it masks.

## IMAGE and PDFUNDERLAY

These need a policy decision before any code. A decoder that resolves external file
paths out of a drawing and reads them is a decoder that will be handed a drawing
pointing at `/etc/passwd` or a UNC path, which is a security boundary this repository
does not currently have anywhere.

The safe shape is to emit the placement without resolving anything: the frame, the
transform, and the referenced name as an opaque string, leaving any fetch to a consumer
that has a policy. That is useful on its own, since a renderer can draw the frame and a
placeholder, and it keeps the decoder from ever opening a path a file told it to.

## Acceptance

- [ ] WIPEOUT emits its boundary, with the masking intent expressed rather than implied.
- [ ] A decision recorded on whether IMAGE and PDFUNDERLAY emit a placement record.
- [ ] If they do: the decoder resolves no path and opens no external file, and a fixture
      pointing at an absolute path proves it.

spdrman:
## Fixtures

| Kind | Writable? | Fixture |
| --- | --- | --- |
[...]

Issue #82, opened 2026-09-13
POLYFACE_MESH and POLYGON_MESH are silently flattened as a polyline by the IPolyline arm


spdrman:
`Flattener.Map`'s `case IPolyline poly:` arm catches more than it was written for, and the
two extra kinds are not polylines at all.

In ACadSharp 3.7.1:

```csharp
public abstract class Polyline<T> : Entity, IPolyline   // src/ACadSharp/Entities/PolyLine.cs:16
public class PolyfaceMesh : Polyline<VertexFaceMesh>    // PolyfaceMesh.cs:14
public class PolygonMesh  : Polyline<PolygonMeshVertex> // PolygonMesh.cs
```

So a POLYFACE_MESH and a POLYGON_MESH both match `IPolyline` and are emitted as a single
open `Polyline` record threaded through the vertex list. Neither is a polyline. A
polyface mesh is a set of faces defined by a vertex list plus face records that index
into it, and a polygon mesh is an M×N grid.

## Why this is worse than refusing them

Every other entity kind this version does not flatten produces an `UNSUPPORTED_ENTITY`
warning, so a consumer is told. These two produce a geometry record that looks exactly
like every other polyline record, and a consumer has no way to know it is holding a
wandering line rather than a surface. Silent wrong geometry is the one outcome
`docs/WIRE.md`'s contract is meant to prevent.

It is likely worse still for POLYFACE_MESH specifically. Its `Vertices` collection holds
both the coordinate vertices and the face records, and a face record is a vertex whose
payload is four indices rather than a position. Threading a polyline through all of them
puts points wherever those records' coordinates happen to sit.

## Evidence

`real_AC1032.dwg` holds 23 LWPOLYLINE, 1 POLYLINE_3D and 1 POLYFACE_MESH at top level.
`tests/expectations/real_AC1032.txt` emits exactly **25** `Polyline` records and carries
no `POLYFACE_MESH` line in its 88 `UNSUPPORTED_ENTITY` warnings. 23 + 1 + 1 = 25, so the
mesh went out as a polyline.

The top-level census is from acadrust 0.5.5, a pure-Rust reader, over the same file:
it models `PolyfaceMesh` as its own type and does not conflate it with `Polyline3D`.

## What to do

Decide, and make the decision visible either way:

- Refuse both explicitly, ahead of the `IPolyline` arm, with `UNSUPPORTED_ENTITY`. That
  is a one-line change and it converts silent wrong geometry into an honest refusal.
[...]

PR #96, opened 2026-09-14
Flatten SOLID and MESH onto the existing Polygon record

### LINKED ISSUE #84: SOLID and 3DFACE are refused; both lower to the existing Polygon record
`SOLID` and `3DFACE` are both refused. On `real_AC1032.dwg` that is 14 SOLID and 1
3DFACE, 15 of the 88 refused entities.

Both are the same shape of problem and should be one piece of work: a planar filled
quadrilateral given by four corners, where a degenerate fourth corner makes it a
triangle. `SOLID` is filled, `3DFACE` is a face with per-edge visibility flags.

## What acadrust models

```rust
pub struct Solid {
    pub common: EntityCommon,
    pub first_corner: Vector3, pub second_corner: Vector3,
    pub third_corner: Vector3, pub fourth_corner: Vector3,
    pub normal: Vector3, pub thickness: f64,
}

pub struct Face3D {
    pub common: EntityCommon,
    pub first_corner: Vector3, pub second_corner: Vector3,
    pub third_corner: Vector3, pub fourth_corner: Vector3,
    pub invisible_edges: ...,
}
```

Four corners each, already read out of these files.

## The trap worth writing down before anyone implements this

**SOLID's corner order is not the polygon's winding order.** DXF stores the third and
fourth corners swapped relative to the traversal order, so a flattener that emits
corners 1, 2, 3, 4 as a closed polygon draws a bow-tie rather than a quad. The order to
emit is 1, 2, 4, 3.

That defect passes every test built out of squares and symmetric shapes, because a
bow-tie over a symmetric quad can still cover a plausible-looking area. A fixture for
this has to use an asymmetric quad, and the assertion has to be on the emitted point
order rather than on an area or a bounding box.

## Mapping

`Polygon` already exists in the wire format and is what a hatch boundary lowers to, so
neither needs a new record. A degenerate fourth corner, equal to the third, is the
triangle case and should emit three points rather than four coincident ones.

`3DFACE`'s invisible-edge flags have no representation in `Polygon`. Dropping them is
defensible for a first pass, and should be said out loud in the record rather than left
for a consumer to discover.

## Acceptance

- [ ] SOLID emits a `Polygon` in 1, 2, 4, 3 order.
- [ ] 3DFACE emits a `Polygon`, with the invisible-edge loss documented.
- [ ] A degenerate fourth corner emits a triangle, not a coincident fourth point.
- [ ] Fixtures use an **asymmetric** quad, and assert the point order, not an area.
- [ ] `real_AC1032.txt` regenerated: 15 warnings become records.

### LINKED ISSUE #88: 3DSOLID, REGION and MESH are refused; MESH is tractable and the other two need ACIS

spdrman:
SOLID and MESH were both refused and both are a filled face given by corners the
file already holds, so both lower to `Polygon`, record 9, with no wire change.
The two arm bodies go in one new file, `native/Adapter/Flatten.Faces.cs`, with
only the `case` labels left in `Flattener.cs` where their order is semantics the
compiler enforces.

Closes #84
Closes #88

## What each closure actually covers, and what it does not

**#84: the SOLID half only.** 3DFACE is deliberately left out, and the issue's
own trap paragraph is half wrong in the dangerous direction, so the corrected
reasoning is written into `Flatten.Faces.cs` rather than left for the next
person to rediscover:

- `Solid` exposes FirstCorner, SecondCorner, ThirdCorner and FourthCorner at DXF
  group codes 10, 11, 12 and 13 verbatim, with no reordering. The swap is DXF's
  and ACadSharp passes it through, so the bow-tie is real and the order is
  **1, 2, 4, 3**.
- `Face3D` exposes the same four names at the same codes **and** carries
  `InvisibleEdgeFlags { First = 1, Second = 2, Third = 4, Fourth = 8 }`. Per-edge
  visibility only means anything if edge 1 is corners 1 to 2, edge 2 is 2 to 3,
  edge 3 is 3 to 4 and edge 4 is 4 to 1, which is traversal order, so a 3DFACE
  emits **1, 2, 3, 4**.

Implementing the issue as written, with one shared `QuadPoints(a, b, c, d)`
helper, fixes SOLID and introduces the bow-tie into 3DFACE. They also differ in
coordinate space: `Solid : Entity, IOrientable` carries a Normal at DXF 210 and
is lifted through `place.WithOcs(solid.Normal, 0.0)` the way the Arc and Circle
arms are, while `Face3D : Entity` carries no normal and its own doc comments say
every corner is in WCS, so it must not be lifted and its record's normal has to
be computed from its own corners rather than taken from the placement basis.

3DFACE is deferred because there is no isolated 3DFACE fixture and the one
instance on `real_AC1032.dwg` cannot check a corner order. The fixture is
producible: `DwgObjectWriter` does have a `case Face3D` with a real
`writeFace3D`, contrary to #93's claim, so that is a follow-up rather than a
dead end.

**#88: the MESH half only.** 3DSOLID and REGION stay refused and that refusal is
the decisions lane's to record, not this one's. Nothing here touches them.

## SOLID

- Corners emitted 1, 2, 4, 3, with the permutation commented at the call site
  naming the group codes.
- A fourth corner equal to the third is the triangle case and emits three
  vertices rather than a coincident fourth. The comparison is exact and is made
  on what the file holds, before any transform, because the format stores a
  literal copy.
- Lifted through the entity's own OCS.
- No bulge array at all: a solid's edges are straight, and record 9 leaves the
  array out when every span is. That is also why there is no
  `NON_UNIFORM_BLOCK_SCALE` on this path, matching the hatch arm: a polygon's
  vertices transform exactly whatever the scale.

## MESH

- One `Polygon` per face of the base mesh, read out of the explicit vertex list
  and the face indices.
- The subdivision level is **ignored, out loud**, through new warning code 110
  `MESH_SUBDIVISION_IGNORED`, emitted once per mesh when the level is above
  zero. Evaluating it is a smoothing algorithm that invents vertices the drawing
  does not hold, and "these are the faces in the file" is a promise this layer
  can keep while "this is what AutoCAD displays" is not. A fixture at level zero
  could not show that anything was ignored, which is why `g13_mesh.dwg` is at
  level 2.
- New warning code 111 `MESH_FACE_UNREADABLE` names a face with fewer than three
  vertices or an index outside its own vertex list, drops that face, and lets
  the rest of the mesh cross. Both numbers are file-controlled, so without it an
  `IndexOutOfRangeException` is reachable from a drawing. It follows
  `HATCH_LOOP_NOT_POLYGON`'s precedent: a per-sub-part condition gets its own
  code. It has no fixture, because producing one needs a malformed mesh in the
  corpus and lanes in this campaign do not regenerate the corpus, so it is
  `None` in `CODE_FIXTURES` the way `DIMENSION_WITHOUT_BLOCK` is.
- `CheckPointCount` runs per face before the array is allocated, which is the
  rule the Spline and LwPolyline arms follow.
- A mesh carries no normal and its vertices are WCS, so each face's plane is
  measured off the face itself with Newell's method rather than taken from the
  placement. That is what makes a mirrored insertion's winding come out right
  with no correction of its own: the normal is computed from the points as
  emitted, so it follows the mirror.

**Codes 110 and 111 sit next to the decisions lane's 109.** If that lane
allocates more than one number, check the two do not collide before composing.

## The prediction

Written before the generator ran, and this is the line the single regeneration
confirms or refutes.

**`g13_solid`, the first solid**, whose corners are `(0,0) (10,1) (2,5) (11,7)`:

```
Polygon flags=0 n=4 closed=1
pts=[(0,0,0);(10,1,0);(11,7,0);(2,5,0)] bulges=[] normal=[(0,0,1)]
```

Shoelace on that order is 50. The 1, 2, 3, 4 bow-tie is 3.5. The assertion is on
the **first** solid, not the third: the third is the rectangle
`(0,0) (4,0) (0,3) (4,3)` whose bow-tie has signed area exactly 0, so a control
there would be measuring a fixed point.

Rest of the file: 5 `Polygon` records in total, 3 at top level and 2 from the two
insertions with flags=1, the second of them the triangle at n=3, the third
naming the plane `(0.333333, 0.666667, 0.666667)`, and no
`UNSUPPORTED_ENTITY` naming SOLID anywhere.

**`g13_mesh`**: 6 `Polygon` records, every one n=3, plus 3
`MESH_SUBDIVISION_IGNORED` warnings, one per mesh. Top-level faces
`(0,0,0);(10,0,0);(10,10,0)` and `(0,0,0);(10,10,0);(0,10,4)`.

**`g13_unsupported`** goes from 2 `UNSUPPORTED_ENTITY` warnings to 1: its SOLID
has four default corners, so it becomes a degenerate three-vertex polygon at the
origin.

### What I measured

I ran `regenerate.py --only expectations` against a **copy of this tree in my
own scratch**, nothing committed, and every line above came back exactly:

```
00004 Polygon handle=49 flags=0 n=4 closed=1 pts=[(0.000000,0.000000,0.000000);
      (10.000000,1.000000,0.000000);(11.000000,7.000000,0.000000);
      (2.000000,5.000000,0.000000)] bulges=[] normal=[(0.000000,0.000000,1.000000)]
```

`g13_solid` 9 records (4 notifications + 5 Polygon), `g13_mesh` 13 records
(4 notifications + 3 subdivision warnings + 6 Polygon), `g13_unsupported` 6
records with one POINT refusal and one Polygon. The mirrored mesh insertion's
faces come back with normals `(0,0,-1)` and `(0.192450,0.192450,-0.962250)`,
which is the winding following the mirror, which is the thing that fixture's
mirrored insert exists to make testable.

Against that regenerated copy the whole suite is **2375 passed, 148 skipped, 0
failed**.

## One thing PR #94 left undone

`g13_solid.dwg` and `g13_mesh.dwg` were in `tests/fixtures` and in
`PROVENANCE.md` and in **neither** `regenerate.py`'s `DUMPED` tuple nor
`MANIFEST.json`, so nothing ever dumped them and no test could read them. I
added the two entries. Every other lane that implements a refused kind needs the
same line for its own fixture, or the integrator's regeneration will quietly
produce no expectation for it.

## This PR is red, on purpose

No `MANIFEST.json` and nothing under `tests/expectations/` is committed, per the
campaign rule, so `test_shim_digest.py` is stale by construction. 22 tests fail
on this branch and all 22 are the pre-regeneration set:

- 15 in the new `test_flatten_faces.py`, which reads expectations that do not
  exist yet
- 3 in `test_flatten_warnings.py`: the two `g13_unsupported` assertions I
  updated, and the new `MESH_SUBDIVISION_IGNORED` fixture row
- 4 in `test_shim_digest.py`

All 22 are green in the regenerated copy above. The rest of the gate is clean on
this branch as it stands: `ruff check` 0, `ruff format --check` 0,
`tools/shellcheck-all.sh` 0, and pytest 2340 passed / 147 skipped beside those
22.

Everything ran in containers on `--platform linux/arm64`.

## One test row removed rather than updated

`TestAnEmptyView::test_a_damaged_view_carries_it_with_company` was parametrised
over three fixtures, one of which was `g13_unsupported.dwg`: it held a POINT and
a SOLID, both refused, so its view was empty and it carried `EMPTY_VIEW`. The
SOLID is flattened now, so that view has geometry and `EMPTY_VIEW` is correctly
absent. The other two rows cover the claim from two different reasons, which is
what the row was for, and the reason is in the test.


spdrman:
Three things from this lane that are not really about this lane, so they are
here rather than buried in the diff.

**1. The `DUMPED` gap hits every lane, not just this one.** PR #94 added seven
refused-kind fixtures to `tests/fixtures` and to `PROVENANCE.md`, and nothing
added any of them to `regenerate.py`'s `DUMPED` tuple. `expectations()` iterates
`DUMPED` and nothing else: it is what writes each `tests/expectations/*.txt` and
what fills `MANIFEST.json["fixtures"]`. So a fixture that is not in it is never
dumped, never gets a manifest entry, and no test can read it, including the
per-fixture parametrisations in `test_adapter_stream.py` that everything else in
the corpus gets for free. I added `g13_solid.dwg` and `g13_mesh.dwg`. Whoever
implements POINT, POLYFACE_MESH, POLYGON_MESH or any of the rest needs the same
line for their own fixture, or the single regeneration at the end will produce
nothing for it and the lane's tests will fail looking like a code bug.

Worth knowing what it is worth: those two entries alone added 35 collected tests
from existing parametrisations, on top of the 15 in the new file.

**2. Warning codes.** I took **110** `MESH_SUBDIVISION_IGNORED` and **111**
`MESH_FACE_UNREADABLE`, on the assumption the decisions lane takes only 109 for
deliberate refusals. If that lane allocates more than one number, the two need
checking before composing. Both of mine have a row in WIRE.md's table and an
entry in `test_flatten_warnings.py`'s `CODE_FIXTURES`, which is where a
collision shows up: `test_the_declared_codes_are_the_ones_this_file_accounts_for`
compares the two sets, and all three files are places two lanes will have edited
adjacent lines.

**3. For whoever picks up 3DFACE.** Beyond the corner-order and coordinate-space
correction in the PR body, the Z of corners 2, 3 and 4 does not survive a DWG
round trip: the format stores one elevation for the entity. That is correct
behaviour for a SOLID and it did not come up here because every corner in
`g13_solid.dwg` is at z=0, but a 3DFACE is genuinely three-dimensional and an
assertion written against a fixture's authored Z values will fail for a reason
that has nothing to do with the flattener. Check what comes back out of the
writer before pinning anything to it.

spdrman:

red and push anyway.

consumers to take.

## Mandatory, in order

   be sufficient on its own.
   silently to code 100.
   it and did not.
     body should too.

## Should fix

PR #97, opened 2026-09-14
Refuse a mac dylib whose install name is not the file beside it

### LINKED ISSUE #95: The macOS dylib's install name is @rpath/viprs_acadsharp.dylib and no such file ships, so a consumer links it and cannot load it

spdrman:
`acadsharp-3.7.1-viprs.1` ships `lib/libacadsharp_native.dylib` whose `LC_ID_DYLIB`
reads `@rpath/viprs_acadsharp.dylib`, and no file of that name is anywhere in the
archive. That string is what a consumer's link copies into its own
`LC_LOAD_DYLIB`, so on `aarch64-apple-darwin`, the one published target where
shared is the only link mode, a consumer links cleanly and the binary dies before
`main`.

Closes #95.

## Why nothing saw it

The build's shared smoke is `dlopen(path, RTLD_NOW)` on an absolute path, and
`dlopen` by absolute path never consults `LC_ID_DYLIB`. It loaded a library whose
recorded name is nonsense and reported success. `verify_archive.sh`'s
link-and-run probe is gated on `static_certified`, which is `false` on mac. So on
the one target where this can happen there was no link-and-run check at all, and
the archive passed every invariant. I ran `main`'s verifier against the published
tarball to be sure of that rather than to assume it:

```
$ bash acadsharp/scripts/verify_archive.sh acadsharp-mac-arm64.tgz     # on main
  exports every entry point the header declares
verify_archive.sh: all invariants hold for acadsharp-mac-arm64.tgz
exit=0
```

## The guard

`verify_archive.sh` already walks Mach-O load commands in POSIX shell to find
`LC_SYMTAB`. Reading `LC_ID_DYLIB` is the same loop with `cmd == 13`, plus the
name slice from `off + nameoff` to `off + cmdsize` cut at the first NUL. About
forty lines with the comments, no `otool`, so it runs on `ubuntu-latest`.

Not `otool -D` as the check: otool is mac-only, so on every Linux runner that
check would skip itself, and a host-capability skip is the same colour as a pass.
This repo has shipped a broken recipe behind exactly that kind of skip before
(the musl cargo link, which is why `VIPRS_REQUIRE_LINK_TEST` exists). `otool -D`
is asked the same question as a **second** check inside the mac-host branch, so
the release runner gets the real tool on the real bytes and every other runner
still gets a check.

Same command, this branch, same tarball:

```
  install name: @rpath/viprs_acadsharp.dylib
FAIL: libacadsharp_native.dylib records the install name '@rpath/viprs_acadsharp.dylib',
    whose basename is not libacadsharp_native.dylib. No file of that name is in this
    archive, so anything that links this library records a name the loader cannot
    resolve and dies before main. [...]
exit=1
```

That is the pure-shell reader, on a Linux container, refusing the real published
Mach-O.

There is a second arm: an install name whose basename is right but which is not
under `@rpath` (a build machine's absolute path, say) is refused too. The
basename comparison is the invariant #95 asks for and it does not catch that one,
and it is the same defect, a recorded name the consumer's loader cannot follow.

## The red-first test

`test_verify_archive.py` synthesises a 64-bit Mach-O because no Linux container
can produce a real one. It now emits an `LC_ID_DYLIB` with a caller-chosen name,
laid out first so the symbol reader has to walk past a command to find its own
rather than finding `LC_SYMTAB` at offset 32 every time.

Four cases: the name 3.7.1 shipped is refused, the matching name is accepted and
reported, an absolute name is refused, and a dylib with no `LC_ID_DYLIB` at all is
refused.

Against `main`'s source, watched failing:

```
6 failed, 1 passed
FAILED ...TestTheMachOInstallName::test_the_name_3_7_1_actually_shipped_is_refused
FAILED ...TestTheMachOInstallName::test_the_matching_name_is_accepted_and_reported
FAILED ...TestTheMachOInstallName::test_an_install_name_that_is_not_under_rpath_is_refused
FAILED ...TestTheMachOInstallName::test_a_dylib_that_records_no_name_at_all_is_refused
FAILED ...TestTheMacPublishSetsTheInstallName::test_the_install_name_is_the_file_the_archive_ships
FAILED ...TestTheMacPublishSetsTheInstallName::test_the_mac_publish_passes_it
```

(the one that passed is the control, "no Linux publish carries the property",
which is true before and after.)

On this branch: `7 passed`.

## The fix, and why it is not the one the issue picked

The issue is wrong about where this lives. `build_acadsharp.py:186` is a constant;
the rename is one line of tracked shell, `scripts/stage.sh:81`. And the issue
prefers `install_name_tool -id` because it needs no extra tool. Measured on the
published binary, in a container:

| | |
| --- | --- |
| `@rpath/viprs_acadsharp.dylib` | 28 bytes, `LC_ID_DYLIB` cmdsize 56 |
| `@rpath/libacadsharp_native.dylib` | 32 bytes, needs cmdsize 64 |
| growth `install_name_tool` must find | 8 bytes |
| load commands end at | 3328 (`sizeofcmds` 3296) |
| `__TEXT`'s first section starts at | 3400 |
| **headerpad free** | **72 bytes** |

So the rewrite would work today, with 64 bytes to spare, on a margin the AOT link
happened to leave and nothing asked for. The failure mode is a build error on the
one target with no local reproduction, so I took the link-time route instead.

Then the interesting part. My first attempt was a `LinkerArg` item of our own, and
it is wrong. `Microsoft.NETCore.Native.Unix.targets:68` already defaults

```
<SharedLibraryInstallName Condition="'$(SharedLibraryInstallName)' == '' and
    '$(_IsApplePlatform)' == 'true' and '$(NativeLib)' == 'Shared'"
    >@rpath/$(TargetName)$(NativeBinaryExt)</SharedLibraryInstallName>
```

and line 237 turns that, unconditionally for an Apple shared library, into the one
`-Wl,-install_name,...` the link gets. `TargetName` is the assembly name, which is
the whole root cause in one line. An extra `LinkerArg` does not replace that, it
adds a second `-install_name` to the same command.

I captured the link line to check, by publishing for `osx-arm64` in a Linux
container with a shell script standing in for `clang`:

```
A default                        -> -Wl,-install_name,@rpath/viprs_acadsharp.dylib
B plus a LinkerArg of our own    -> -Wl,-install_name,@rpath/libacadsharp_native.dylib
                                    -Wl,-install_name,@rpath/viprs_acadsharp.dylib   <- last wins
C -p:SharedLibraryInstallName    -> -Wl,-install_name,@rpath/libacadsharp_native.dylib
```

So the fix is C: `publish_command()` passes
`-p:SharedLibraryInstallName=@rpath/libacadsharp_native.dylib` on the mac cell
only, and the name is built from `SHARED_LIBRARY_STEM`, the same constant the
archive is named from, so the two cannot drift. The csproj needs nothing; the
property is the SDK's own, and its default carries a `== ''` condition precisely
so it can be named from outside.

`A` is also the root cause reproduced without a Mac: the default really does
produce the exact string the shipped binary carries.

## What I am not claiming

**Nothing in this org runs a macOS binary.** Docker does not run macOS, this Mac's
toolchain is off limits, and the NAS is x86_64 Linux. So:

* the guard is **proven**, red then green, and proven again against the real
  published archive's real bytes;
* `A`, `B` and `C` above are **measured**, but they are measurements of the *link
  command line*, captured with a stand-in for clang. No Mach-O was produced;
* the build-side fix is therefore **reviewed and not executed**. Nothing has
  loaded a dylib built this way.

The real-bytes half fires at the next release, in `release-acadsharp.yml`'s
`build-mac` job, which already runs this verifier on `macos-15` and will now
refuse the archive if the property did not take. I am deliberately not writing
"fixed" about a binary nothing has loaded.

## Also

`docs/LINKINFO.md` says to link the `.dylib` "the ordinary way", and does not
mention that a mac consumer has to supply an rpath for `@rpath/...` to expand
against. #95 assumes that paragraph already says so. It did not. It does now,
with the `dlopen`-ignores-the-recorded-name note beside it, because that is the
fact that let this ship.

## Gate

Container, `--platform linux/arm64`, `CI=1` so the compiled fixtures are strict.

| | baseline (`main`) | this branch |
| --- | --- | --- |
| `ruff check .` | pass | pass |
| `ruff format --check .` | pass | pass |
| `tools/shellcheck-all.sh` | pass | pass |
| `pytest -q` | 2494 passed | 2501 passed |

No `.cs` file is touched, so `shim_digest` does not move and `test_shim_digest`
stays green: this lane needs no regeneration and pushes no stale manifest.


spdrman:

red and push anyway.

consumers to take.

## Mandatory, in order

   be sufficient on its own.
   silently to code 100.
   it and did not.
     body should too.

## Should fix

PR #98, opened 2026-09-14
Say which refusals are decisions and which are just unfinished

### LINKED ISSUE #85: RAY and XLINE are refused, and the wire format has no unbounded record to put them in

### LINKED ISSUE #88: 3DSOLID, REGION and MESH are refused; MESH is tractable and the other two need ACIS

### LINKED ISSUE #89: IMAGE, PDFUNDERLAY and WIPEOUT are refused; dropping a WIPEOUT shows what should be hidden

### LINKED ISSUE #90: SHAPE is refused, and its geometry is in an external SHX this decoder should not resolve

### LINKED ISSUE #91: The DWG version floor is AC1014 with no recorded reason; ACadSharp and acadrust both read R13
`AbiConstants.DwgVersionMin` is `1014`, so `VersionGate` refuses anything below R14.

```csharp
// native/Abi.cs:108
public const uint DwgVersionMin = 1014u;
public const uint DwgVersionMax = 1032u;
```

acadrust 0.5.5 declares read and write support from **AC1012 (R13)** upward
(`src/lib.rs:143`), and refuses AC1009 with an explicit error, which is the same floor
this repository's `g13_ac1009.dwg` fixture exists to prove. So the two implementations
agree about where the bottom is, except for two versions in between:

| Version | Release | libviprs-dep | acadrust |
| --- | --- | --- | --- |
| AC1009 | R11/R12 | refused | refused |
| AC1012 | R13 | **refused** | supported |
| AC1013 | R13c3 | **refused** | supported |
| AC1014 | R14 | supported | supported |
| AC1015+ | 2000+ | supported | supported |

## Is this a gap worth closing

Not obviously, and that is the point of filing it rather than quietly doing it. R13 is
from 1994. The question is whether any drawing a consumer will hand this decoder is
still in it, and nobody here has data either way.

What is worth deciding is whether `1014` was chosen or inherited. There is no comment at
the constant and nothing in `docs/` explains it, so a reader today cannot tell whether
R13 was ruled out after testing or simply never tried. ACadSharp itself reads R13, which
means the floor is this repository's, not the underlying library's.

## The cheap experiment

`tests/fixtures/gen/regenerate.py` drives ACadSharp's `DwgWriter` in a pinned container
and the writer can target R13. Writing one R13 fixture and lifting `DwgVersionMin` to
`1012` behind a branch would answer it in an afternoon: either the reader handles it and
the floor was arbitrary, or it does not and the constant gets the comment it is missing.

## Acceptance

- [ ] `DwgVersionMin` carries a comment saying why it is what it is.
- [ ] Either the floor moves to 1012 with a fixture proving R13 decodes, or `docs/`
      records that R13 was tried and what failed.

### LINKED ISSUE #92: Differential comparison against acadrust: 88 refused entities on a real drawing, and one silent wrong-geometry defect
A differential comparison against [acadrust](https://github.com/hakanaktt/acadrust) 0.5.5,
an independent pure-Rust DWG reader, run over this repository's own fixture corpus. The
harness and its method are in
[spdrman/acadrust#2](https://github.com/spdrman/acadrust/pull/2).

acadrust is useful here precisely because it shares no code with this adapter. Where the
two agree, the agreement is evidence. Where they disagree, one of them is wrong and it is
worth knowing which.

## What agrees

acadrust opens 42 of the 43 fixtures; the refusal is `g13_ac1009.dwg`, which this
repository also refuses. On the 17 fixtures whose geometry the harness can compare
without reimplementing the flattener, **15 match to 1e-6**, including every OCS fixture,
which means the arbitrary-axis lift in `Flattener` is right.

`g13_scale_1x` is the strongest single result: 128 records, 64 circles and 64 lines, every
value agreeing to six decimal places.

## The refused-entity gap

On `real_AC1032.dwg`, a real AutoCAD drawing rather than one written by our own fixture
generator, this adapter emits 222 geometry records and refuses 88 entities:

```
  40 POINT          3 TOLERANCE       1 XLINE     1 REGION
  15 MULTILEADER    3 MLINE           1 WIPEOUT   1 RAY
  14 SOLID          2 MESH            1 SHAPE     1 PDFUNDERLAY
                    2 3DSOLID                     1 LEADER
                                                  1 IMAGE
                                                  1 3DFACE
```

**acadrust reads all 88, with matching counts for every type**, and carries real geometry
for each rather than just a handle. So none of these gaps are blocked on what ACadSharp
exposes; they are all flattener and wire-format work.

| Issue | Kinds | Count | Note |
| --- | --- | --- | --- |
| #83 | POINT | 40 | Biggest single gap and the cheapest. Needs a wire record. |
| #84 | SOLID, 3DFACE | 15 | Lower to the existing `Polygon`. Corner-order trap inside. |
| #86 | MULTILEADER, LEADER, TOLERANCE | 19 | Composites; DIMENSION is the precedent. |
| #87 | MLINE | 3 | First kind needing a style-table lookup. |
| #88 | 3DSOLID, REGION, MESH | 5 | MESH is tractable; the other two need ACIS. |
| #89 | IMAGE, PDFUNDERLAY, WIPEOUT | 3 | External resources; WIPEOUT is the urgent one. |
| #85 | RAY, XLINE | 2 | Unbounded; no record can hold them today. |
| #90 | SHAPE | 1 | Needs SHX. Probably `wontfix`. |

Two of those are not really "missing features" and are worth pulling out:

- **#89's WIPEOUT.** Every other omission loses something that should have been drawn.
  Dropping a wipeout *shows* something that should have been hidden, which produces a
  plausible wrong drawing rather than an obviously incomplete one.
- **#88's 3DSOLID and REGION, and #90's SHAPE.** These cannot be done on top of ACadSharp
  at all. They deserve a recorded decision rather than an open backlog item implying they
  are merely unscheduled.

## Where the outputs disagree

**#82, ours.** `POLYFACE_MESH` and `POLYGON_MESH` are silently emitted as a `Polyline`
threaded through their vertices, because `Polyline<T> : Entity, IPolyline` in ACadSharp
and both inherit it, so they match the `case IPolyline` arm. No warning, so a consumer
cannot tell. This is the only defect the comparison found in this repository, and it is
the kind the whole wire contract exists to prevent.

**Theirs, recorded for completeness, not actionable here.** acadrust returns extrusion
normals un-normalized: `(1,2,2)` where this adapter gives `(0.333,0.667,0.667)`. The
circle centre agrees to every decimal, so its arbitrary-axis maths is right and only the
returned vector is raw. And acadrust emits non-finite geometry: on `g13_nan_bulge` it
passes a NaN bulge and an Infinity straight through, where `docs/WIRE.md`'s promise makes
this adapter refuse both. **This adapter is the correct one in both cases**, which is
worth stating plainly given how much of the rest of this issue is a list of our gaps.

**#91, neither.** The version floor here is AC1014 with no recorded reason; ACadSharp and
acadrust both read R13.

## What is still unmeasured, and it is the important part

16 of the 43 fixtures could not be compared at all, because they need block expansion,
hatch lowering or dimension expansion, and the harness deliberately does not reimplement
any of that: a second flattener would mean every difference it found might be its own
bug.

So the headline should not be read as "15 of 17 agree, therefore the adapter is 88%
right". Block expansion, hatch boundaries and dimension flattening are the parts a CAD
renderer leans on hardest, and this comparison says nothing about them. Closing that gap
means recording a second oracle **after** flattening, which is the natural next piece of
work and is not filed yet.

spdrman:
Closes #85, closes #88, closes #89, closes #90, closes #91, closes #92

Every entity kind this build does not flatten left the same warning: code 100,
`X is not a primitive this version flattens`. That sentence covers two
completely different situations and a consumer cannot tell them apart, because
`WIRE.md` forbids parsing the message and 100 was the only number either of
them carried. "Nobody has got to MLINE yet" and "no decoder built on this
reader will ever render a 3DSOLID" are different facts, and the second is the
one a consumer wants when it decides whether to draw a placeholder or go and
find another tool.

## Warning codes this lane claims

**109 only.** `ENTITY_REFUSED_BY_DESIGN`, and nothing else. No collision with
the MESH lane's 110 and 111.

Its `CODE_FIXTURES` entry is `real_AC1032.dwg`, which carries seven of the
eight kinds in the table below, so it does not need the `None`-with-a-reason
pattern.

## What landed

**Warning code 109, `ENTITY_REFUSED_BY_DESIGN`.** For a kind somebody looked at
and decided against. Adding it is not a wire version bump: `WIRE.md` already
says a consumer skips a code it does not know rather than refusing the stream,
and `test_warning_codes.py` already proves the C# enumeration and the
specification table name the same set with no .NET present. One constant, one
`Name` case, one documentation row.

**`native/Adapter/RefusedKinds.cs`.** Eight rows, consulted by `Flattener`'s
`default:` arm before it falls back to 100.

| Kind | Code | Why |
| --- | --- | --- |
| 3DSOLID, REGION | 109 | Embedded ACIS, a boundary representation nothing on this boundary evaluates |
| SHAPE | 109 | The outline is in an external SHX this decoder does not open |
| IMAGE, PDFUNDERLAY | 109 | The content is an external file and this decoder opens no path a drawing names |
| RAY, XLINE | 109 | Unbounded, and no record in wire 2 can hold one |
| WIPEOUT | **100** | Deferred work, not a decision, with its own sentence |

WIPEOUT stays on 100 deliberately. Its geometry is in the file, it lowers to
the `Polygon` record that already exists, so 100 is honest about it. What it
gains is its own message, because it is the one refusal that reveals rather
than loses: a wipeout is a mask, so dropping it shows what the drawing meant to
hide, and the blast radius is the area it covers rather than the one entity it
is.

**`docs/adr/0002-what-this-decoder-refuses.md`.** Verdict line, a row per kind
with disposition, code, reason and revisit condition, a "not decided" section
naming the three records a wire 3 should carry (Point, Unbounded, Placement),
and the trap that comes with any of them. `WireFormat.IsGeometry` is the
contiguous range 3 to 10 and the numbering is frozen, so a new geometry record
lands at 14, falls outside `Flattener.Finite()`, and `WIRE.md`'s finiteness
guarantee quietly stops covering it while staying literally true. Nothing fails
to compile and no test of either half notices.

**`tests/test_refusal_decisions.py`.** Holds the ADR's table, the C# table and
`WIRE.md`'s warning-code table to the same kinds and the same codes, same shape
as `test_warning_codes.py` and needing no .NET.

**#92** closes as the report it is, with its headline numbers copied into ADR
0002 so the census has a home that does not depend on an issue staying open.

## Where I put the lookup, and why it is not a `case`

The `default:` arm consults `RefusedKinds` keyed on `ObjectName`, rather than
eight `case` arms on concrete ACadSharp types.

Two reasons. A `case Entity refused when ...` is a case on a base type, which
is exactly the construct the POLYFACE_MESH lane is adding a source rule to
forbid in this switch. And naming eight CLR types (`Solid3D`, `RasterImage`,
`PdfUnderlay`, ...) cannot be verified here, because nothing in this
repository's gate compiles C#, so a wrong type name would land unnoticed.
`ObjectName` is the string the wire already carries and every expectation dump
in `tests/expectations` names these kinds with it.

#85 and #90 ask for an explicit arm rather than a fall-through so the next
reader can see somebody looked. A named table with a reason and a revisit
condition per row, plus a test tying it to the ADR and the specification, does
that better than eight empty labels would. Flagging the deviation rather than
burying it.

## What I found about AC1012 (#91)

The two reviewers disagreed about whether an AC1012 fixture can be forged, so I
settled it by running it, in the pinned SDK container on arm64: build a
one-line document at three versions with ACadSharp 3.7.1's own `DwgWriter` and
read each back.

```
WRITE AC1012: ACadSharp.Exceptions.CadNotSupportedException: File version not supported: AC1012
WRITE AC1014: OK, 7360 bytes
READ  AC1014: OK, version=AC1014, entities=1
WRITE AC1015: OK, 7582 bytes
READ  AC1015: OK, version=AC1015, entities=1
```

So the reviewer who found `DwgWriter` throwing is right. The two halves of
ACadSharp disagree: the reader takes AC1012 (`CreateFileHeader` sends it to
`DwgFileHeaderAC15`, `GetStreamHandler` to `DwgStreamReaderAC12`) and
`DwgWriter.getFileHeaderWriter` lists it in the same arm as AC1009 and throws.
It refuses AC1021 from the same switch, which is why `tests/ac21_forge.py`
exists at all.

**The floor does not move**, and not because R13 fails. #91's cheap experiment
does not exist: there is no way to write the fixture that would prove R13
decodes, short of forging an R13 file by hand the way `ac21_forge.py` forges
AC1021, and R13's object stream differs from R14's by considerably more than
AC1021's page metadata differs from AC1018's.

Two corrections recorded with it. **AC1013 is not in ACadSharp's `ACadVersion`
enum**, which goes `AC1009`, `AC1012 = 19`, `AC1014 = 21`, so #91's table row
for it invents a version and there are exactly two candidates below 1014, which
are 1012 and nothing. And the comment beside `DwgVersionMin` cited ADR 0001 for
"upstream's own reader table"; **ADR 0001 contains no reader table, no 1014 and
no 1012**. The citation pointed at nothing, which is worse than no comment
because it stops the next reader looking. It now says what was tried and cites
ADR 0002, and `test_refusal_decisions.py` has a control asserting ADR 0001
really does lack those strings, so the fix cannot be satisfied by deleting a
citation that was in fact correct.

The asymmetry is why leaving it alone is safe: lowering this floor is additive
and needs no version to move, raising it is breaking, cannot be expressed on
this wire, and a consumer finds out one drawing at a time.

## Honest scope on #88 and #89

Both issues are split across two lanes and this PR is half of each.

* **#88.** The ACIS half landed: 3DSOLID and REGION now carry 109 with a reason,
  and ADR 0002 records that as a decision about ACIS rather than a backlog item.
  **MESH is another lane's** and is deliberately absent from the table, because
  a row here would say the opposite of what is true about it: its faces are an
  explicit index list, it needs no evaluator, and it lowers to one `Polygon` per
  face. ADR 0002 says so in as many words.
* **#89.** The external-resource half landed: IMAGE and PDFUNDERLAY carry 109
  and the decision is recorded, including `WIRE.md`'s new "geometry that needs a
  lookup" rule saying no path out of a drawing is ever opened and what a later
  placement record would carry. **WIPEOUT stays refused this round**, moved from
  a silent fall-through to a table row at code 100 with a message naming what
  dropping it reveals. Its implementation is not here.

So `Closes #88` and `Closes #89` are fair, but the acceptance boxes those
issues carry for MESH and for WIPEOUT's boundary are met by other work, not by
this PR. Worth a look before merge if either issue should stay open instead.

## `g13_ray_xline.dwg` was never readable

#94 added seven fixtures and none of them reached `regenerate.py`'s `DUMPED`
tuple, which `expectations()` iterates and nothing else. So those DWGs get no
expectation dump and no `MANIFEST.json` entry: files in the tree the suite
cannot read, which is why they landed green and inert.

I added `g13_ray_xline.dwg`, the one this branch is about. It carries a RAY and
an XLINE at the top level plus a block holding one of each, instanced twice, so
it evidences the refusal on a drawing carrying almost nothing else and
exercises the block-expansion flag on the same warning. The other six belong to
the lanes that own their kinds, and `g13_point.dwg` and `g13_tolerance.dwg`
have no lane in this batch, so they stay inert unless somebody adds them.

**Prediction for the new dump.** One `Line` record, then two
`ENTITY_REFUSED_BY_DESIGN` warnings at `flags=0` (RAY then XLINE, in write
order), then four more at `flags=1` from the two insertions of the block, two
per insertion, each carrying the block entity's own handle. Plus the reader
notifications the generator's document always produces.

No `EMPTY_VIEW`: the fixture holds a real `Line`, unlike `g13_unsupported.dwg`,
`g13_xref.dwg` and `g13_nan_bulge.dwg`. And no `NON_UNIFORM_BLOCK_SCALE`
despite the insertions being 3.0 by 1.5 and -2.0 by 2.0, because only the arms
that read the basis raise it and a refusal arm does not.

## Expected reds, and a prediction

This PR touches `native/` and `regenerate.py` and does not touch
`tests/expectations/` or `MANIFEST.json`, per the campaign's serialisation
rule. Six failures, all of
them the stale-capture alarm:

* `test_shim_digest.py`, 5 failures. `native/Abi.cs`,
  `native/Adapter/Flattener.cs` and `native/Adapter/WarningCodes.cs` moved, and
  `native/Adapter/RefusedKinds.cs` is a new source the generator's glob picks up.
* `test_flatten_warnings.py::test_the_fixture_produces_the_code[ENTITY_REFUSED_BY_DESIGN-real_AC1032.dwg]`,
  1 failure. Every code the shim declares has to have a fixture behind it, and
  `real_AC1032.dwg` carries seven of the eight kinds. It is red against the
  expectation dump recorded before this change and goes green on regeneration.

Everything else is green: ruff check, ruff format, shellcheck, and 2502 passed
against a baseline of 2489 passed with 0 failures. The 19 new tests are 18 in
`test_refusal_decisions.py` plus the new parametrised case above.

**Prediction for the single regeneration.** In `real_AC1032.txt` and
`real_AC1018.txt`, **9 lines change in each and nothing else does**. Same
handles, same flags, same record indices, same record count, so the monotonic
5-digit index does not renumber.

| Fixture | Line (1032 / 1018) | Handle | Code after | Message |
| --- | --- | --- | --- | --- |
| XLINE | 59 / 52 | `389` | `ENTITY_REFUSED_BY_DESIGN` | new |
| RAY | 60 / 53 | `38A` | `ENTITY_REFUSED_BY_DESIGN` | new |
| WIPEOUT | 76 / 69 | `44D` | `UNSUPPORTED_ENTITY` (unchanged) | new |
| SHAPE | 220 / 213 | `889` | `ENTITY_REFUSED_BY_DESIGN` | new |
| IMAGE | 221 / 214 | `8C6` | `ENTITY_REFUSED_BY_DESIGN` | new |
| PDFUNDERLAY | 329 / 322 | `D07` | `ENTITY_REFUSED_BY_DESIGN` | new |
| 3DSOLID | 331 / 324 | `D65` | `ENTITY_REFUSED_BY_DESIGN` | new |
| REGION | 332 / 325 | `D69` | `ENTITY_REFUSED_BY_DESIGN` | new |
| 3DSOLID | 333 / 326 | `D6A` | `ENTITY_REFUSED_BY_DESIGN` | new |

The other 79 `UNSUPPORTED_ENTITY` lines on each fixture (40 POINT, 15
MULTILEADER, 14 SOLID, 3 TOLERANCE, 3 MLINE, 2 MESH, 1 LEADER, 1 3DFACE) keep
their exact text from this change. No other expectation file carries any of
these kinds, and `g13_scenarios.json` carries none of these messages.

In `MANIFEST.json`, `Warning` counts, `records`, `notification_count`,
`view_extents` and `batches` are all unchanged, and `output_bytes` grows by
exactly **1236** on each fixture:

* `real_AC1032.dwg` 35716 -> **36952**
* `real_AC1018.dwg` 34536 -> **35772**

That is the sum over the nine records of the change in
`32 + len(message) + padding to 4`. Still one batch either way, since
`TargetBatchBytes` is 65536.

If the regeneration disagrees with any of that, the disagreement is the
interesting thing, not the number.

## One thing for the integrator

`g13_scenarios.json` holds two entries for `real_AC1032.dwg`
(`warnings/reader_notification` and `declared/real_ac1032_control`) and both
record `output_bytes: 35716`, the same number `MANIFEST.json` carries. A
regeneration of `--only expectations` moves the manifest's copy to 36952 and
leaves the scenarios' copies at 35716, so the two capture files end up
disagreeing about the same decode. **No test compares them**, so nothing goes
red and nothing tells you. I checked every test function that reads both and
there is none.

Not fixing it here, because that means running the `scenarios` stage and this
lane does not regenerate anything. Worth a run of `--only scenarios` alongside
`--only expectations`, or an assertion that the two agree.


spdrman:
Two discoveries from this lane that do not belong in the diff, recorded here so
they are not re-derived.

## The AC1012 probe, reproducible

Reviewers disagreed about whether the fixture generator can forge an R13
drawing, so I ran it rather than reading it. Pinned SDK image, arm64, ACadSharp
3.7.1 source with the CSUtilities submodule, a console project referencing
`src/ACadSharp/ACadSharp.csproj` and this `Main`:

```csharp
foreach (ACadVersion v in new[] { ACadVersion.AC1012, ACadVersion.AC1014, ACadVersion.AC1015 })
{
    CadDocument doc = new CadDocument();
    doc.Header.Version = v;
    doc.Entities.Add(new Line(new XYZ(0, 0, 0), new XYZ(10, 0, 0)));
    using (DwgWriter w = new DwgWriter(path, doc)) { w.Write(); }
    CadDocument back = DwgReader.Read(path);
}
```

```
WRITE AC1012: ACadSharp.Exceptions.CadNotSupportedException: File version not supported: AC1012
WRITE AC1014: OK, 7360 bytes
READ  AC1014: OK, version=AC1014, entities=1
WRITE AC1015: OK, 7582 bytes
READ  AC1015: OK, version=AC1015, entities=1
```

The source behind it, for anyone checking without building:

* `IO/DwgWriter.cs:137`, `getFileHeaderWriter`, lists `AC1012` in the same
  throwing arm as `AC1009`, and throws separately for `AC1021`.
* `IO/DWG/FileHeaders/DwgFileHeader.cs:23`, `CreateFileHeader`, sends `AC1012`
  to `DwgFileHeaderAC15`.
* `IO/DWG/DwgStreamReaders/DwgStreamReaderBase.cs:38`, `GetStreamHandler`,
  sends `AC1012` to `DwgStreamReaderAC12`.
* `ACadVersion.cs` goes `AC1009`, `AC1012 = 19`, `AC1014 = 21`. The gaps are
  unnamed, so there is no `AC1013` member to support or refuse.

The reader takes R13 and the writer will not produce it. That is why the floor
does not move, and it is a different reason from "R13 fails".

## `g13_scenarios.json` and `MANIFEST.json` will disagree after regeneration

Both `warnings/reader_notification` and `declared/real_ac1032_control` record
`output_bytes: 35716` for `real_AC1032.dwg`, which is the number
`MANIFEST.json` carries today. A regeneration of `--only expectations` moves
the manifest's copy and leaves the scenarios' copies where they are, so the two
capture files end up describing the same decode with two different sizes.

Nothing catches it. I walked every test function that reads both a `scenario()`
and `MANIFEST` and there is none, so this is silent by construction rather than
by accident. Any lane that changes a warning message on `real_AC1032.dwg` or
`real_AC1018.dwg` widens it.

Cheapest fix is running `--only scenarios` alongside `--only expectations` in
the single regeneration. The durable one is an assertion that a scenario naming
a fixture agrees with the manifest entry for it, which would have caught this
without anybody looking.

spdrman:

red and push anyway.

consumers to take.

## Mandatory, in order

   be sufficient on its own.
   silently to code 100.
   it and did not.
     body should too.

## Should fix

Issue #87, opened 2026-09-13
MLINE is refused, and it is the first kind whose geometry needs a style-table lookup


spdrman:
`MLINE` is refused. 3 on `real_AC1032.dwg`.

An MLINE is a run of parallel lines sharing one path: a wall, a road, a pipe. The path
is a vertex list and the parallel offsets come from an MLINE style, which is a table
object rather than something on the entity.

## What acadrust models

```rust
pub struct MLine {
    pub common: EntityCommon,
    pub flags: ..., pub justification: ...,
    pub normal: Vector3, pub scale_factor: f64,
    pub start_point: Vector3,
    pub style_handle: ...,
}
```

Note `style_handle`. The offsets are not on the entity in either implementation: they
live in the MLINE style the handle points at, and the entity carries a scale factor and
a justification that say how to apply them.

## Why that matters here

This is the first refused kind whose geometry cannot be computed from the entity alone.
Every other one on the list is self-contained. An MLINE needs a table lookup, and the
flattener has to decide what to do when the style is missing, which on a real drawing it
sometimes is.

The failure mode to avoid is emitting the path as a single `Polyline` and calling it
done. That renders a wall as a centre line: plausible, wrong, and no warning to say so.
It is the same defect shape as #82.

## Sketch

Resolve the style, take its element offsets, apply `scale_factor` and the justification
to place each element relative to the path, emit one `Polyline` per element. A missing or
unresolvable style is an `UNSUPPORTED_ENTITY` warning, not a fallback to the centre line.

Joints, caps and fills are a second pass and should be left out of the first, with the
omission stated.

## Acceptance

- [ ] One `Polyline` per style element, offset correctly for the justification.
- [ ] A missing style warns rather than emitting the bare path.
- [ ] A fixture with at least three elements and a non-zero scale factor, so an
      implementation that ignores either is red.
- [ ] `real_AC1032.txt` regenerated.

spdrman:
## Fixtures

**MLINE is not writable.** `MLine` has no case in `DwgObjectWriter.Entities.cs` and the
final arm throws `NotImplementedException`, so no isolated fixture can be generated from
this repository's own writer.

| Source | Handles |
| --- | --- |
| `real_AC1032.dwg` | **`38B`, `3A5`, `3A6`** |

Three MLINEs, already refused and recorded in `tests/expectations/real_AC1032.txt`.

### Why the style-table half makes this harder than the others

[...]

PR #99, opened 2026-09-14
Refuse POLYFACE_MESH and POLYGON_MESH instead of flattening them as polylines

### LINKED ISSUE #82: POLYFACE_MESH and POLYGON_MESH are silently flattened as a polyline by the IPolyline arm

spdrman:
Closes #82

`case IPolyline poly:` caught more than it was written for. `PolyfaceMesh` and
`PolygonMesh` both derive from `Polyline<T>`, `Polyline<T>` implements
`IPolyline`, so both reached that arm and went out as one open polyline
threaded through their own vertices. In `real_AC1032.dwg` that is handle
`0x422`:

```
00070 Polyline handle=422 flags=0 n=5 closed=0
      pts=[(213.692029,2.252867,0.000000);(222.457789,3.348587,0.000000);
           (215.603726,6.635747,0.000000);(215.944553,12.075878,0.000000);
           (211.562496,8.827187,0.000000)]
```

A five point open polyline through the mesh's vertices, no warning, and
identical in shape to the polylines beside it. A consumer draws a wandering
line and reports success, which is the one outcome `docs/WIRE.md`'s contract
exists to prevent. `acadsharp-rs` decodes this record today.

## What I changed

Two arms ahead of the `IPolyline` one, each emitting `UNSUPPORTED_ENTITY` with
the entity's handle and carrying `flags` the way the `default:` arm does. They
go through a new `Refused` helper beside `NonUniform`.

The message is the DXF subclass marker and not `ObjectName`. All four kinds
under `Polyline<T>` are the entity `POLYLINE` in DXF and are told apart by the
subclass alone, so `ObjectName` would refuse a polyface mesh by saying
`POLYLINE`. What comes out is `AcDbPolyFaceMesh` and `AcDbPolygonMesh`, which
are upstream's `DxfSubclassMarker` constants and the names DXF itself uses in
group 100. Worth being explicit that the census therefore gains an
`AcDbPolyFaceMesh` row rather than a `POLYFACE_MESH` one: the first token is
the kind either way, and this way the string is the format's rather than ours.

## The source rule, which is the part that outlives the instance

`acadsharp/tests/test_shim_source_rules.py` gains
`TestNoArmSwallowsAKindNobodyWroteItFor`. It reads the entity switch as text,
with no .NET, and refuses an arm that matches a kind nothing ahead of it names.

The compiler is worth being precise about here, because it looks like it should
have caught this and it cannot. It orders the arms that exist: a subtype after
its base, or a class after an interface it implements, is CS8120 and a hard
build error, so these arms could not have been put in the wrong place. What
nothing checks is whether the arm is there at all, and an absent arm is the
defect. That is the question the rule asks.

The subtype table is ACadSharp 3.7.1's entity model read off the pinned
tarball, and a sibling test fails the moment `acadsharp/VERSION` moves, because
a new kind under an existing base is exactly what this is for and the rule
cannot see one that is not in the table. Leaves are not in the table and do not
need to be, so a lane adding `case Point` or `case Solid3D` is not held up by
it.

I watched it fail first, on the tree as it stood:

```
FAILED test_shim_source_rules.py::TestNoArmSwallowsAKindNobodyWroteItFor::
  test_no_arm_matches_a_kind_with_no_arm_ahead_of_it
AssertionError: these arms match kinds nothing ahead of them names:
  `case IPolyline` also matches PolyfaceMesh,
  `case IPolyline` also matches PolygonMesh.
assert not [('IPolyline', 'PolyfaceMesh'), ('IPolyline', 'PolygonMesh')]
1 failed, 26 passed
```

Every control in the class passed at that point: the extraction control, the
pin, the negative control on a fabricated pre-#82 arm list, and the ordering
control that `case Arc` ahead of `case Circle` satisfies the rule while the
reverse does not.

## The rule in WIRE.md

New section, `Geometry that needs a lookup`, beside the refusal table:

> When an entity's geometry depends on something outside the entity, a producer
> emits the resolved geometry or it emits `UNSUPPORTED_ENTITY` and no geometry
> record at all. It never emits the part it could compute without the lookup.

The part computable without the lookup is the dangerous output, because it is
type-compatible with a correct one and the consumer cannot tell. That is a
defect and not a degradation. #87's MLINE and #89's external resources turn on
the same rule, and `UNRESOLVED_BLOCK` already follows it.

## The two fixtures were there, and nothing could read them

`g13_polyface_mesh.dwg` and `g13_polygon_mesh.dwg` are already in
`tests/fixtures`, `Corpus.cs` writes them, and `PROVENANCE.md` carries a row
and a digest for each. Neither was in `regenerate.py`'s `DUMPED` tuple.
`expectations()` iterates that tuple and nothing else, so neither produced a
dump, neither got a `MANIFEST.json` entry, and nothing in the suite could open
them. They looked green because they were inert. I added the two lines.

They are built for exactly this defect: the polyface mesh fixture orders its
vertices so a line threaded through them in storage order self-intersects, and
both carry a third instance inside a block, so the from-block flag goes through
the refusal too.

Heads up for the compose: the SOLID/MESH lane edits the same tuple, so those
two hunks will want resolving together.

No new warning code. Both arms emit 100, `UNSUPPORTED_ENTITY`, so there is
nothing here to collide with 109, 110 or 111.

## My prediction, written before I ran anything

- `real_AC1032.txt`: record `00070` stops being a `Polyline` and becomes
  `00070 Warning handle=422 flags=0 code=UNSUPPORTED_ENTITY message="AcDbPolyFaceMesh is not a primitive this version flattens"`.
  `Polyline` 25 to 24, `Warning` 119 to 120, `UNSUPPORTED_ENTITY` 88 to 89, 341
  records before and after, and no index past `00070` moves.
- `real_AC1018.txt`: record `00063`, same handle, same message. `Polyline` 25 to
  24, 334 records before and after.
- Neither real file gains an `AcDbPolygonMesh` line, because `entities_by_type`
  in both captures has `PolyfaceMesh: 1` and no `PolygonMesh` at all.
- The regeneration writes two new expectations. Each is four
  `READER_NOTIFICATION` records, then four `UNSUPPORTED_ENTITY` records, then
  `EMPTY_VIEW`, with `flags=1` on the last two of the four because they come
  out of the block:

```
00004 Warning handle=49 flags=0 code=UNSUPPORTED_ENTITY message="AcDbPolyFaceMesh is not a primitive this version flattens"
00005 Warning handle=53 flags=0 code=UNSUPPORTED_ENTITY message="AcDbPolyFaceMesh is not a primitive this version flattens"
00006 Warning handle=60 flags=1 code=UNSUPPORTED_ENTITY message="AcDbPolyFaceMesh is not a primitive this version flattens"
00007 Warning handle=60 flags=1 code=UNSUPPORTED_ENTITY message="AcDbPolyFaceMesh is not a primitive this version flattens"
00008 Warning handle=0 flags=0 code=EMPTY_VIEW message="this view holds no geometry"
```

  and the same shape for `g13_polygon_mesh.txt` with `AcDbPolygonMesh` and
  handles `49`, `57`, `68`, `68`.
- No other expectation changes.

## What the generator actually showed

I built fixturegen against the pinned SDK and ACadSharp 3.7.1 in my own
scratch and replayed the whole corpus with `fixturegen decode --check`.
Nothing was written into the tree.

```
31/33 identical, 2 red

RED  real_AC1032.dwg: record 70 differs
  expected: 00070 Polyline handle=422 flags=0 n=5 closed=0 pts=[...]
  actual:   00070 Warning handle=422 flags=0 code=UNSUPPORTED_ENTITY message="AcDbPolyFaceMesh is not a primitive this version flattens"
RED  real_AC1018.dwg: record 63 differs
  expected: 00063 Polyline handle=422 flags=0 n=5 closed=0 pts=[...]
  actual:   00063 Warning handle=422 flags=0 code=UNSUPPORTED_ENTITY message="AcDbPolyFaceMesh is not a primitive this version flattens"
```

Counts off the dumps: `Polyline` 25 to 24 in both, `UNSUPPORTED_ENTITY` 88 to 89
in both, `Warning` 119 to 120 in AC1032 and 112 to 113 in AC1018, 341 and 334
records before and after, one line changed in each file and no renumbering. The
build succeeding is itself the check that the arms are legal where they sit.

Then `--only expectations` against a copy of this tree in scratch, which is
where the two new fixture dumps come from. Both match the prediction line for
line.

And the control, which is the half that matters: the same `--only expectations`
against `origin/main` with only the two `DUMPED` lines added, so the arms are
the only difference. On `main` the same fixtures dump geometry.

```
00004 Polyline handle=49 flags=0 n=6 closed=0 pts=[(0,0,0);(10,0,0);(0,10,0);(10,10,6);(20,0,6);(20,10,0)]
...
00004 Polyline handle=49 flags=0 n=12 closed=0 pts=[(0,0,0);(0,3,1.5);(0,6,0);(0,9,1.5);(5,0,1.5);...]
```

A six point line through the polyface mesh's vertices and a twelve point line
through the polygon mesh's 3 by 4 grid in storage order, four of each, with no
warning anywhere in either file. That is the defect, on an isolated input, and
it is gone on this branch. The prediction holds in every part.

## Corrections to #82 recorded here

- **The "likely worse still" paragraph does not hold for 3.7.1.**
  `PolyfaceMesh : Polyline<VertexFaceMesh>` keeps positions in `Vertices` and
  face records in a separate `Faces` collection, and `Polyline<T>`'s explicit
  `IPolyline.Vertices` returns only `this.Vertices`. The defect is a wandering
  line through real vertices, not points at index-shaped garbage. Severity is
  unchanged, the shape is not, and a fixture assertion written from the issue's
  description would assert something that cannot happen.
- **The acceptance criterion "25 to 23" cannot be met, and the naive fix cannot
  meet it either.** `real_AC1032.dwg` holds one `PolyfaceMesh` and no
  `PolygonMesh`, so exactly one record changes. The measured move is 25 to 24.
- **The naive one-line fix also names the wrong thing.** `PolyfaceMesh.ObjectName`
  is `DxfFileToken.EntityPolyline`, the string `"POLYLINE"`, so reusing the
  `default:` arm's message would emit `POLYLINE is not a primitive this version
  flattens` and the census would gain a `POLYLINE` row. Hence `SubclassMarker`.
- **The comment on the `Arc` arm was wrong.** It claimed the other order
  silently turns every arc into a full circle. It does not, it is CS8120 and the
  build fails. I corrected it, and said there what the compiler does and does
  not hold, because that distinction is the whole argument for the source rule.
- The acceptance's "a fixture for each, with an expectation" is met by the two
  `DUMPED` lines rather than by new DWGs: the fixtures already existed and only
  the dump was missing. The expectation files themselves come out of the
  integrator's single regeneration, not this branch.

## Things I found and did not act on

- **`TableEntity : Insert` in 3.7.1, and `Walk` dispatches on `item.Entity is
  Insert`.** Both real captures hold `TableEntity: 2`, neither expectation
  carries a warning for them, so those two tables are being expanded as block
  insertions today with nobody having decided that. It is the same shape as
  this bug in a different dispatch, and the block a table references is its
  rendered content so the output may well be reasonable, but it is undeclared.
  I kept the source rule scoped to the entity switch rather than widen it into
  a decision that is not this lane's.
- **`case TextEntity` matches `AttributeDefinition` as well as
  `AttributeEntity`.** `InsertBody` yields the insertion's `Attributes` and then
  every entity in the block record, and a block's ATTDEFs are in that list, so
  one insertion can reach the definition and the instance. Nothing in the corpus
  has an attribute, so nothing exercises it. I recorded it as the reason string
  on that pair in `CARRIED` rather than guess.
- **The `default:` arm still spells the refusal sentence inline.** It should
  fold onto the new `Refused` helper. I left it alone because another lane is
  inserting immediately above it, and the comment on the helper says so.
- **`WIRE.md`'s row for code 100 says the message "names the source format's
  type".** A subclass marker is the source format's name for the kind, so I read
  that as satisfied, but somebody may want the row to say so outright. I left
  the table alone to keep out of lane 4's way.

## Gate

Everything in a container, `--platform linux/arm64`.

- `ruff check .`: All checks passed.
- `ruff format --check .`: 91 files already formatted.
- `tools/shellcheck-all.sh`: exit 0, 12 scripts.
- `pytest -v`: **2 failed, 2494 passed, 5 skipped**. Both failures are
  `test_shim_digest.py`, the expected staleness alarm, because I changed
  `Flattener.cs` and did not touch `MANIFEST.json`.

The delta against `origin/main`, same image, same flags, baseline exported to
scratch and run first:

| | baseline | this branch |
| --- | --- | --- |
| passed | 2484 | 2489 |
| failed | 1 | 3 |
| skipped | 9 | 9 |

Total collected goes 2494 to 2501, which is my 7 new tests exactly. The 2 extra
failures are the two `test_shim_digest.py` ones. The 1 failure both runs share
is `zstd/tests/test_ci_coverage.py`, which shells out to `git ls-files` and had
no git in that container; with the repository's git directory mounted the branch
run is the 2 failed / 2494 passed / 5 skipped above and that one passes.

Nothing under `tests/expectations/` and no `MANIFEST.json` is in this branch.


spdrman:
Four things out of this lane that are somebody else's to act on, so they are
here rather than buried in the body.

**1. `TableEntity : Insert`, and `Walk` dispatches on `item.Entity is Insert`.**
This is the same shape as #82 in a different dispatch. Both real captures hold
`TableEntity: 2`, neither expectation carries a warning for either of them, so
two ACAD_TABLEs are being expanded as block insertions right now with nobody
having decided that. It may even be reasonable, since the block a table
references is its rendered content, but it is undeclared and untested. I kept
the new source rule scoped to the entity switch rather than widen it into a
decision that is not this lane's to make.

**2. `case TextEntity` matches `AttributeDefinition` as well as
`AttributeEntity`.** `InsertBody` yields the insertion's `Attributes` and then
every entity in the block record, and a block's ATTDEFs are in that list, so
one insertion can reach the definition and the instance and emit text for both.
Nothing in the corpus has an attribute so nothing exercises it. I recorded it
as the reason string on that pair in `CARRIED` instead of guessing.

**3. Nothing checks that a fixture on disk is in `DUMPED`.** That is the hole
that let `g13_polyface_mesh.dwg` and `g13_polygon_mesh.dwg` sit in the tree with
`PROVENANCE.md` rows and no expectation. The guard belongs with the corpus
replay lane, which already owns the missing-fixture check in the other
direction, and I did not write a second one here to stay out of its way. Worth
checking whether any other DWG is in the same state.

**4. The `default:` arm still spells the refusal sentence inline.** There is a
`Refused` helper next to `NonUniform` now. The default arm should fold onto it
next time anything touches that arm; I left it alone because another lane is
inserting immediately above it.

And one note on the composition: this branch adds two entries to `DUMPED` and so
does the SOLID/MESH lane, in the same tuple, so those hunks want resolving
together. No new warning code from here, both arms emit 100.

spdrman:

red and push anyway.

consumers to take.

## Mandatory, in order

   be sufficient on its own.
   silently to code 100.
   it and did not.
     body should too.

## Should fix

PR #100, opened 2026-09-14
Replay the committed corpus in CI, so the captures stop being forgeable

spdrman:
## What was wrong

The binding between this repository's committed captures and the code that produced
them is forgeable, and I reproduced the forgery before writing anything.

`MANIFEST.json` carries a sha256 per source the fixture generator compiles and
`test_shim_digest.py` recomputes them, which is meant to be what stops a dump from
being a recording of a flattener nobody runs any more. But `regenerate.py`'s
`shim_block()` computes that block with the same `g13_support.shim_digest` the test
verifies it with, so the block is a pure function of the source tree. It carries no
information about whether a decode ever ran, and four lines rewrite it:

```python
import g13_support as g
m = json.load(open(g.MANIFEST))
m["shim"]["sources"] = {n: g.sha256_file(os.path.join(g.ACAD_ROOT, n)) for n in g.shim_sources()}
m["shim"]["sha256"] = g.shim_digest()
```

CI did not close it. `ci.yml` is ruff, pytest and shellcheck, and neither conformance
consumer opens a fixture or reads an expectation: both drive `SyntheticSource`
through `open_memory`. Nothing on any runner had ever decoded a committed DWG.

## Reproduction 1: the positive control, 33 of 33 identical

`python3 acadsharp/tests/fixtures/gen/replay.py --image acadsharp-builder-3.7.1-viprs.1-linux-arm64:latest`
against this branch's tip, in the builder image, 7.8 seconds warm and 13.6 from a
cold generator build:

```
identical  g11_codepage.dwg
identical  g11_shapes.dwg
identical  g13_arc.dwg
identical  g13_bad_extents.dwg
identical  g13_circle.dwg
identical  g13_deep_blocks.dwg
identical  g13_dimension.dwg
identical  g13_dimension_shallow.dwg
identical  g13_ellipse.dwg
identical  g13_empty_view.dwg
identical  g13_hatch.dwg
identical  g13_insert.dwg
identical  g13_line.dwg
identical  g13_long_text.dwg
identical  g13_mirrored_bulge.dwg
identical  g13_nan_bulge.dwg
identical  g13_nonuniform.dwg
identical  g13_ocs_mirror.dwg
identical  g13_ocs_plane.dwg
identical  g13_ocs_rotated.dwg
identical  g13_ocs_skew.dwg
identical  g13_polyline.dwg
identical  g13_scale_1x.dwg
identical  g13_slot.dwg
identical  g13_slot_block.dwg
identical  g13_spline.dwg
identical  g13_text.dwg
identical  g13_two_entities.dwg
identical  g13_unsupported.dwg
identical  g13_wide_polyline.dwg
identical  g13_xref.dwg
identical  real_AC1018.dwg
identical  real_AC1032.dwg

33 of 33 fixtures replay identical to what is committed
```

## Reproduction 2: the same replay against a mutated flattener

I applied #82's refusal to `Flattener.cs`, a `case PolyfaceMesh: case PolygonMesh:`
arm ahead of the `IPolyline` one emitting a `SubclassMarker` warning, and reverted it
afterwards. `test_shim_digest.py` noticed, as designed:

```
FAILED test_shim_digest.py::TestTheCapturesAreOfTheShimInTheTree::test_each_source_is_byte_for_byte_what_recorded_the_captures[native/Adapter/Flattener.cs]
FAILED test_shim_digest.py::TestTheCapturesAreOfTheShimInTheTree::test_the_rollup_covers_the_whole_set
2 failed, 37 passed in 0.10s
```

The replay named which fixtures and which record, matching the panel's measurement
exactly:

```
  DIFFERS  real_AC1018.dwg
  DIFFERS  real_AC1032.dwg

real_AC1018.dwg: the committed dump is not what the shim produces now
  record 63 differs
  expected: 00063 Polyline handle=422 flags=0 n=5 closed=0 pts=[(213.692029,2.252867,0.000000);(222.457789,3.348587,0.000000);(215.603726,6.635747,0.000000);(215.944553,12.075878,0.000000);(211.562496,8.827187,0.000000)] bulges=[] normal=[(0.000000,0.000000,1.000000)]
  actual:   00063 Warning handle=422 flags=0 code=UNSUPPORTED_ENTITY message="AcDbPolyFaceMesh is not a primitive this version flattens"
real_AC1032.dwg: the committed dump is not what the shim produces now
  record 70 differs
  expected: 00070 Polyline handle=422 flags=0 n=5 closed=0 pts=[(213.692029,2.252867,0.000000);(222.457789,3.348587,0.000000);(215.603726,6.635747,0.000000);(215.944553,12.075878,0.000000);(211.562496,8.827187,0.000000)] bulges=[] normal=[(0.000000,0.000000,1.000000)]
  actual:   00070 Warning handle=422 flags=0 code=UNSUPPORTED_ENTITY message="AcDbPolyFaceMesh is not a primitive this version flattens"

2 of 33 replayed fixtures disagree with what is committed.
```

Then I ran the forgery on that same mutated tree, the four lines above, and this is
the part that matters:

```
=== the forgery: four lines rewrite the shim block ===
477 passed in 0.28s

=== the replay, same forged tree ===
  DIFFERS  real_AC1018.dwg
  DIFFERS  real_AC1032.dwg
```

`test_shim_digest.py` and `test_adapter_stream.py` went fully green over expectations
recording two records the shim had stopped emitting, and the replay stayed red. That
is the hole, closed.

## What I built

**1. The corpus replay, in CI.** `acadsharp/tests/fixtures/gen/replay.py` builds the
generator in the builder image the archive build already made, decodes every fixture
the manifest names, and compares each against the dump beside it with
`fixturegen decode --check`, which prints the first differing record. One step in the
`consumers` job, ahead of the AbiTest publish because it is the cheaper half.

The driver's own comparison is unit tested, because a guard whose failure path has
never run is the shape this PR is about: five separate refusals for the five ways a
replay goes wrong, including the one that would otherwise read as a pass, a decode
that compared nothing at all. The fixture list comes out of the manifest and the
driver names no fixture of its own, which is asserted. Both ends of "CI runs it" are
guarded and I watched both fail with the step deleted.

The loop inside the container is `sh`, not Python: the builder image's python3 is
`python3-minimal` and has no `json` module, so the reading happens on the host side
across a marker protocol.

**2. The missing-fixture guard.** `test_no_dump_is_orphaned` catches a dump with no
fixture. Nothing caught the other direction, which is how #94's seven DWGs landed
outside every check in the repository.

Following the note from the SOLID/MESH lane, the guard asks `regenerate.py`'s `DUMPED`
rather than only the manifest, because `expectations()` walks that tuple and nothing
else: a fixture missing from it can never get a manifest entry however often anybody
regenerates, so the manifest is the symptom and `DUMPED` is the cause. The manifest's
agreement with `DUMPED` is a case of its own, which cleanly separates "nobody told the
generator about this file" from "the generator knows and has not been rerun".

Neither allow-list skips. A skip is the same colour as a pass and both lists are meant
to shrink, so an excused fixture is asserted to still need excusing: the day one is
recorded, the red test is the one telling you to take it off the list, which is a
one-line edit that does not wait for a regeneration. I watched it fail with the lists
empty (16 red naming exactly the nine gaps), with a recorded fixture excused, with a
name whose file is gone, with a `DUMPED` entry that has no DWG, and in the compose
state where a lane has added to `DUMPED` and not regenerated.

**3. The `PROVENANCE.md` digest check.** A grep of the suite for `PROVENANCE` returned
nothing, so every sha256 in that file was hand-carried and verified by nothing, which
is the failure `docs/LINKINFO.md` warns about one directory over. `hashlib`, no
container, no .NET, runs in the existing pytest job. The table reader keeps a row it
cannot parse rather than dropping it and a separate case refuses it, because a row
silently dropped is a digest silently unchecked. Watched failing on a digest with one
digit changed, on a row whose digest column is not one, and on a row for a file nobody
committed.

**4. The docstring that oversold.** `test_shim_digest.py` read as though the shipped
library were covered. The generator lists its sources; the native project lists none
and takes the SDK's default glob, so the library compiles one file the generator does
not, `native/Exports.cs`. That exclusion is right, and it is only right while it stays
one file, so the set difference is now asserted in both directions and there is a case
on the reason itself: `Exports.cs` is excused for being the ABI surface, so it has to
still hold entry points. Watched failing on a second `.cs` under `native/` and on the
project opting out of the default glob.

## Folded in

* **The streaming benchmark measures two fixtures it did not write down.**
  `g13_scale_4x.dwg` and `g13_scale_16x.dwg` are named by `amplification.json` with no
  `fixture_sha256`, and `test_adapter_benchmarks.py` asserts on their numbers,
  including that the 16x point is sixteen times the 1x one. Those assertions are about
  whatever file is on disk. Fixing it needs a `fixture_sha256` per streaming entry,
  which is a benchmark regeneration, so it is named in `NAMED_BUT_NOT_PINNED` instead.
* **`PROVENANCE.md` covers 41 of 47 DWGs.** It opens with "Every DWG in this
  directory, where it came from, and under what licence" and misses six. All six are
  pinned by `MANIFEST.json` so the gap is the documentation rather than the evidence,
  but the file is where a fixture's source and licence live. I cannot write the rows,
  because the third column describes what the drawing holds, so they are in
  `UNDOCUMENTED` and that list shrinks like the others.
* **A driver imported by path leaves a `.pyc` beside it.** Gitignored, so never
  committed, but a cached copy is one more thing that can be of a different file than
  the one on disk, and it bit me once while I was watching these fail. The shared
  `load_script()` turns bytecode writing off around the load.

## Two things about merging this

**This PR closes no issue.** It is the campaign's foundation, so there is no
`Closes #N` line and that is deliberate rather than an omission.

**It is green, and that is not the campaign's rule.** Every other lane in this batch
pushes with a stale manifest and a red `test_shim_digest.py`, because they change shim
sources. This one changes no C# at all, so the digest still matches and the full suite
passes: 2834 passed, 5 skipped, up from a 2489-passed baseline on `main` in the same
container. Nothing here is exempt from the compose gate, which is still the merge
signal.

Gate run in a `python:3.12-slim` arm64 container with git, shellcheck, gcc, binutils
and pyyaml: `ruff check .` clean, `ruff format --check .` clean on 94 files,
`tools/shellcheck-all.sh` clean, `pytest -q` at 2834 passed / 5 skipped / 0 failed.


spdrman:
Two things I hit while making `PROVENANCE.md` testable that belong to other lanes, not
to this PR.

**The prose in `PROVENANCE.md` carries the claim the panel corrected.** Under "The
refused-entity corpus" it says:

> Six refused kinds are absent here and stay on the real drawing, because ACadSharp's
> `DwgObjectWriter` has no case for them and its final arm throws
> `NotImplementedException`: MULTILEADER, MLINE, 3DFACE, 3DSOLID, REGION and
> PDFUNDERLAY.

`DwgObjectWriter` has explicit cases with real write methods for five of those six.
Only REGION reaches the default arm. #94 is already merged so the correction has
nowhere else to land, and I did not touch the paragraph because the refusal semantics
belong to the lanes working #82 and #84 through #90, and rewriting it correctly means
deciding what actually replaces it. Whoever writes the corrected sentence: this file is
now under test, so the digests below it will tell you if you disturb a row.

**The replay driver will cover new fixtures for free, but only once they are dumped.**
`replay.py` takes its list from `MANIFEST["fixtures"]` and names no fixture of its own,
so a lane adding a fixture to `DUMPED` gets it replayed in CI the moment the integrator
regenerates. Before that regeneration the fixture is in `DUMPED` and not in the
manifest, which my `test_the_generator_and_the_manifest_agree_on_what_is_dumped` reports
as one red with the exact command to fix it. That is the expected state for every lane
in this batch, and it clears at compose.

spdrman:

red and push anyway.

consumers to take.

## Mandatory, in order

   be sufficient on its own.
   silently to code 100.
   it and did not.
     body should too.

## Should fix

PR #106, opened 2026-09-14
Lower 3DFACE onto record 9, with the invisible edges said out loud

### LINKED ISSUE #84: SOLID and 3DFACE are refused; both lower to the existing Polygon record

spdrman:
3DFACE lands on record 9, the `Polygon` that already exists, so no wire version
moves and no consumer is rebuilt. The SOLID half of this landed already and the
3DFACE half did not, because 3DFACE looks like the same work and is not: DXF
swaps a SOLID's third and fourth corners relative to traversal order and does
not swap a 3DFACE's, and Face3D carries no normal at all where Solid carries
one at DXF 210. So `Face3DPolygon` is its own method rather than a shared
`QuadPoints(a, b, c, d)`, its corners go out 1, 2, 3, 4 through `place.Apply`
with no OCS lift, and its normal is Newell's over the emitted points. The one
thing record 9 cannot carry is per-edge visibility, and warning 113 is where
that is said out loud rather than dropped in silence.

Closes #84

- **SOLID emits a Polygon in 1, 2, 4, 3 order.** Met already, before this
  branch. Untouched here: `SolidPolygon` is not edited.
- **3DFACE emits a Polygon, with the invisible-edge loss documented.** Met.
  The arm is `Flatten.Faces.cs`, the code is 113
  `FACE_EDGE_VISIBILITY_IGNORED`, its row is in WIRE.md's warning table, and
  the sentence beside record 9 says which order a 3DFACE's corners come out in
  and why its normal is measured rather than read.
- **A degenerate fourth corner emits a triangle, not a coincident fourth
  point.** Met. `face.FourthCorner == face.ThirdCorner`, compared exactly on
  what the file holds. That is safe here rather than optimistic: `writeFace3D`
  writes the fourth corner as a 3DD delta against the third and `read3dFace`
  reads it with the third as its default, so an equal pair round-trips equal.
- **Fixtures use an asymmetric quad, and assert the point order, not an area.**
  Met. `g13_face3d.dwg`'s first face is the same asymmetric quad
  `g13_solid.dwg` holds, listed in traversal order rather than DXF's, and
  `TestFace3DCornerOrder` compares the emitted points one at a time. The
  control beside it measures both orders: 50 and 3.5.
- **`real_AC1032.txt` regenerated: 15 warnings become records.** Met at
  composition. 14 of the 15 are the SOLIDs and already crossed; the fifteenth
  is handle `322` and is in the prediction below. The composer runs the single
  regeneration.

## The prediction

`g13_face3d.dwg`, handles replaced by `H`:

```
00004 Polygon handle=H flags=0 n=4 closed=1 pts=[(0.000000,0.000000,0.000000);(10.000000,1.000000,0.000000);(11.000000,7.000000,0.000000);(2.000000,5.000000,0.000000)] bulges=[] normal=[(0.000000,0.000000,1.000000)]
00005 Polygon handle=H flags=0 n=3 closed=1 pts=[(20.000000,0.000000,0.000000);(26.000000,2.000000,0.000000);(22.000000,6.000000,0.000000)] bulges=[] normal=[(0.000000,0.000000,1.000000)]
00006 Polygon handle=H flags=0 n=4 closed=1 pts=[(30.000000,0.000000,0.000000);(34.000000,0.000000,0.000000);(34.000000,0.000000,3.000000);(30.000000,0.000000,3.000000)] bulges=[] normal=[(0.000000,-1.000000,0.000000)]
00007 Warning handle=H flags=0 code=FACE_EDGE_VISIBILITY_IGNORED message="3DFACE marks its first and third edges invisible, and record 9 has no per-edge visibility, so the face crosses whole and the flags are ignored rather than the face being split"
00008 Polygon handle=H flags=0 n=4 closed=1 pts=[(40.000000,0.000000,0.000000);(46.000000,1.000000,0.000000);(47.000000,5.000000,0.000000);(41.000000,4.000000,0.000000)] bulges=[] normal=[(0.000000,0.000000,1.000000)]
00009 Polygon handle=H flags=1 n=4 closed=1 pts=[(40.000000,0.000000,0.000000);(58.000000,1.500000,0.000000);(61.000000,7.500000,0.000000);(43.000000,6.000000,0.000000)] bulges=[] normal=[(0.000000,0.000000,1.000000)]
00010 Polygon handle=H flags=1 n=4 closed=1 pts=[(80.000000,0.000000,0.000000);(68.000000,2.000000,0.000000);(66.000000,10.000000,0.000000);(78.000000,8.000000,0.000000)] bulges=[] normal=[(0.000000,0.000000,-1.000000)]
```

Eleven records in all, and I am quoting the count I measured rather than the
one I expected: `{'Warning': 5, 'Polygon': 6}`. Six polygons, because a block
member is emitted once per INSERTION and `AddBlockInstances` inserts its block
twice, so the fixture's one block face is four top-level faces plus two
expanded copies. Five warnings, because four are the reader notifications
every fixture in this corpus carries and the fifth is the 113. No `EMPTY_VIEW`
either, and that is correct rather than missing: this view holds geometry.

Four records at the top of that list are the four top-level faces, at `n` = 4,
3, 4 and 4; the standing one names `(0, -1, 0)` because it is in the XZ plane
and the placement's basis would have said `+Z`; the flagged one is preceded by
its 113 under the same handle; the two `flags=1` records are the block copies
and the mirrored one's normal is `(0, 0, -1)`, which is record 9's documented
MESH behaviour and is the opposite of what the SOLID in the same position
does.

On `real_AC1032.dwg` the line

```
00046 Warning handle=322 flags=0 code=UNSUPPORTED_ENTITY message="3DFACE is not a primitive this version flattens"
```

becomes

```
00046 Polygon handle=322 flags=0 n=4 closed=1 pts=[(75.039418,8.050392,0.000000);(80.721042,8.050392,0.000000);(82.190427,1.492336,0.000000);(76.606762,1.492336,0.000000)] bulges=[] normal=[(0.000000,0.000000,-1.000000)]
```

Four vertices, so that face's fourth corner is not its third, and no 113
beside it, so it marks no edge invisible. Geometry goes 371 to 372, warnings
104 to 103, entity refusals 73 to 72 with the whole drop on code 100 and code
109 unchanged at 8. `real_AC1018.dwg` moves the same way, one line earlier.

## What I measured

This lane has no external oracle and does not need one: the question is a
permutation of numbers the fixture itself holds, and what it needs is a probe
that can go red. So I wrote the tests first and then ran them against a
deliberately wrong arm before writing the real one, in the scratch copy only.
The wrong arm was the plausible one: SOLID's 1, 2, 4, 3 permutation, the
placement's basis normal, and the edge flags dropped in silence. Nine
assertions went red and each named one of those three:

```
test_the_first_face_is_emitted_in_1_2_3_4_order      vertex 2 is (2.0, 5.0, 0.0) and traversal order puts (11.0, 7.0, 0.0) there
test_the_two_orders_are_different_polygons           the emitted order encloses 3.5 and the correct quad is 50
test_a_standing_face_names_its_own_plane             the standing face names the plane (0.0, 0.0, 1.0)
test_its_corners_are_the_ones_the_file_holds         vertex 2 is (30.0, 0.0, 3.0) and not (34.0, 0.0, 3.0)
test_a_face_with_invisible_edges_still_crosses_whole_and_says_so   0 faces said their edge flags were ignored
test_the_message_names_which_edges                   IndexError, there is no warning to read
test_each_block_copy_is_its_own_insertion            insertion 0 vertex 2 is (43.0, 6.0, 0.0) and not (61.0, 7.5, 0.0)
test_the_mirrored_copy_flips_the_measured_normal     the mirrored insertion names (0.0, 0.0, 1.0)
test_the_fixture_produces_the_code[FACE_EDGE_VISIBILITY_IGNORED-g13_face3d.dwg]   produced ['READER_NOTIFICATION']
```

The two that stayed green under the wrong arm are the ones that should:
`test_a_fourth_corner_equal_to_the_third_is_a_triangle`, because three points
are the same three points in either order, and `test_the_other_faces_say
_nothing`, which is a control separating a per-face warning from a per-drawing
one rather than a discriminator on the order.

Then the correct arm, and the same scratch copy is green apart from the two
numbers the composer re-measures. Green here is a comparison rather than a
zero, because a plain `python:3.12-slim` has no `git` and an rsynced copy has
no `.git` either, so `build_acadsharp.driver_commit()` takes a chunk of
`test_verify_archive.py` and `test_build_manifests.py` down with it. I ran a
pristine tree at `de43c6f` in the same image as the control:

| tree | `git` in the image | result |
| --- | --- | --- |
| pristine `de43c6f` | no | 23 failed, 2559 passed, 128 skipped, 13 errors |
| this branch, regenerated | no | 26 failed, 2594 passed, 128 skipped, 13 errors |
| pristine `de43c6f` | yes | 2594 passed, 129 skipped |
| this branch, regenerated | yes | 3 failed, 2629 passed, 129 skipped |

So 35 new passing tests, the same 23 failures and 13 errors in both columns
where the image is the cause, and three real failures which are ADR 0002's
census (73 refusals becomes 72, twice) and `SHIM_DIGESTS["3.7.1-viprs.2"]`.
Both are on the composer's re-measure list.

One thing I could not exercise from a fixture and guarded anyway: the flag
word is a `BitShort` the file controls and `read3dFace` casts it straight to
the enum, so a drawing can set a bit above 8 that names none of the four
edges. Without a guard `EdgeList` would index an empty list from inside the
walk. It warns and says which bits instead, and no fixture reaches that
branch.

## Red until composition

Nothing here is green on this branch and all of it is the same fact: the dump
for the new fixture does not exist until the composer regenerates, and the
shim digest has moved.

```
acadsharp/tests/test_acadsharp_version.py::TestTheShimRevisionMovesWhenTheShimDoes::test_the_shim_in_the_tree_is_the_one_that_version_names
acadsharp/tests/test_adapter_stream.py::TestNoFixtureIsOutsideEveryCheck::test_the_generator_and_the_manifest_agree_on_what_is_dumped
acadsharp/tests/test_flatten_faces.py::TestFace3DCornerOrder::test_the_first_face_is_emitted_in_1_2_3_4_order
acadsharp/tests/test_flatten_faces.py::TestFace3DCornerOrder::test_the_two_orders_are_different_polygons
acadsharp/tests/test_flatten_faces.py::TestFace3DCornerOrder::test_a_fourth_corner_equal_to_the_third_is_a_triangle
acadsharp/tests/test_flatten_faces.py::TestAFaceMeasuresItsOwnPlane::test_a_standing_face_names_its_own_plane
acadsharp/tests/test_flatten_faces.py::TestAFaceMeasuresItsOwnPlane::test_its_corners_are_the_ones_the_file_holds
acadsharp/tests/test_flatten_faces.py::TestInvisibleEdges::test_a_face_with_invisible_edges_still_crosses_whole_and_says_so
acadsharp/tests/test_flatten_faces.py::TestInvisibleEdges::test_the_other_faces_say_nothing
acadsharp/tests/test_flatten_faces.py::TestInvisibleEdges::test_the_message_names_which_edges
acadsharp/tests/test_flatten_faces.py::TestFace3DInsideABlock::test_the_block_copies_cross_and_are_labelled
acadsharp/tests/test_flatten_faces.py::TestFace3DInsideABlock::test_each_block_copy_is_its_own_insertion
acadsharp/tests/test_flatten_faces.py::TestFace3DInsideABlock::test_the_mirrored_copy_flips_the_measured_normal
acadsharp/tests/test_flatten_faces.py::TestFace3DIsNoLongerRefused::test_nothing_still_refuses_a_3dface
acadsharp/tests/test_flatten_faces.py::TestNeitherFixtureProducesAnythingElse::test_only_polygons_and_warnings[g13_face3d.dwg]
acadsharp/tests/test_flatten_warnings.py::TestEveryCodeHasAFixture::test_the_fixture_produces_the_code[FACE_EDGE_VISIBILITY_IGNORED-g13_face3d.dwg]
acadsharp/tests/test_shim_digest.py::TestTheCapturesAreOfTheShimInTheTree::test_each_source_is_byte_for_byte_what_recorded_the_captures[native/Adapter/Flatten.Faces.cs]
acadsharp/tests/test_shim_digest.py::TestTheCapturesAreOfTheShimInTheTree::test_each_source_is_byte_for_byte_what_recorded_the_captures[native/Adapter/Flattener.cs]
acadsharp/tests/test_shim_digest.py::TestTheCapturesAreOfTheShimInTheTree::test_each_source_is_byte_for_byte_what_recorded_the_captures[native/Adapter/WarningCodes.cs]
acadsharp/tests/test_shim_digest.py::TestTheCapturesAreOfTheShimInTheTree::test_the_rollup_covers_the_whole_set
```

That is 20, and nothing else. Two more go red **after** the regeneration and
are the composer's to fix, not this branch's, because they read the committed
`real_AC1032.txt` and it has not moved yet:
`test_acadsharp_adr.py::TestTheCensusIsRecomputed::test_every_stated_count_is_the_one_in_the_corpus`
and `::test_the_point_bullet_counts_the_same_refusals`. Both are the 73 to 72
step and both are on the re-measure list.

## What I did not do

- No wire change. No header, no `WireFormat.cs`, no record type, no
  `IsGeometry`, no consumer.
- No `VERSION` bump and no `SHIM_DIGESTS` edit. The `3.7.1-viprs.2` row is the
  composer's single re-measure on the composed tree.
- Nothing under `acadsharp/tests/expectations/`, and `regenerate.py` never ran
  against this worktree. Every measurement above is from an rsync'd copy in my
  own scratch.
- `SolidPolygon` is not touched, there is no shared quad helper, and
  `RefusedKinds.cs` is not touched: a 3DFACE was refused on code 100 through
  the default arm and never had a row in that table.
- The edge flags are not turned into geometry. No face is split into lines, no
  edge is dropped from the vertex run, and nothing is drawn for the
  invisibility.
- The `MarkerKind` comment in `Flattener.cs` still lists 3DFACE. It is
  explicitly the historical refusal census of `real_AC1032` and reads as one,
  and another lane in this wave is told to leave it alone, so I have too.

One finding on the way past, for whoever is next in `PROVENANCE.md`. The
paragraph saying six refused kinds cannot be forged because ACadSharp's
`DwgObjectWriter` has no case for them was wrong about 3DFACE: there has been
a `case Face3D` with a real `writeFace3D` behind it the whole time, which is
how this fixture exists at all. I corrected that sentence for 3DFACE only, and
left the rest, but the same writer has `case MLine` at line 563,
`case MultiLeader` at 569, `case PdfUnderlay` at 575 and `case Solid3D` at
618, so the reason that paragraph gives looks wrong for four of the remaining
five as well. Worth reading the writer rather than the list before assuming
any of them is unwritable.


spdrman:

PR #107, opened 2026-09-14
MLINE lowers to one Polyline per style element, offset along each vertex's miter

### LINKED ISSUE #87: MLINE is refused, and it is the first kind whose geometry needs a style-table lookup

spdrman:
MLINE lands on record 4 as one Polyline per element of its style, offset along
each vertex's miter, and nothing about the wire moves: no header change, no
consumer change, one new warning code through the channel `docs/WIRE.md`
already sanctions.

The reason this kind is worth the care is in the issue. An MLINE is a centre
path plus a handle to an MLINESTYLE, and the lines a drawing shows sit at the
offsets that style holds. The part that can be computed without the lookup is
the centre path, and emitting it alone is a `Polyline` record with a plausible
point count and real coordinates that no consumer could tell from a real
polyline. So either every element crosses, or nothing does and a warning says
why.

Closes #87

* **One `Polyline` per style element, offset correctly for the justification.**
  Met. Per element, in style order, the path is offset by
  `(offset − reference) × scale` along each vertex's miter, where the reference
  is 0 for Zero, the largest offset for Top and the smallest for Bottom, and a
  positive offset lies to the left of the direction about the entity's normal.
  Every element carries the entity's handle, which is record 9's MESH
  precedent.
* **A missing style warns rather than emitting the bare path.** Met, and the
  finding under it matters: a dangling style handle is not observable at this
  layer at all. `CadMLineTemplate.build` assigns `Style` only when the handle
  resolves and otherwise leaves the entity holding `MLineStyle.Default`, whose
  `AssignDocument` then runs `TryAdd` and hands back the document's own
  "Standard". So a file whose style is missing decodes as though its style were
  Standard with offsets +0.5 and -0.5, and nothing here can tell the two apart.
  What this arm does refuse, on code 100 with a sentence naming the reason, is a
  style that resolves and holds no elements, and a path with fewer than two
  vertices. The arm's comment says where a later pin that exposes the
  unresolved handle would change that.
* **A fixture with at least three elements and a non-zero scale factor.** Met.
  `g13_mline.dwg` carries three styles; the working one has offsets +1, 0 and
  -1.5 (asymmetric on purpose, so Top and Bottom are different numbers and
  neither is zero) and the entities using it have `ScaleFactor` 2.5.
* **`real_AC1032.txt` regenerated.** At composition. I measured it in my scratch
  copy and the numbers are under "The prediction".

## The prediction

`g13_mline.dwg`, with handles replaced by `H`:

```
Polyline handle=H flags=0 n=3 closed=0 pts=[(0.000000,2.500000,0.000000);(7.500000,2.500000,0.000000);(7.500000,8.000000,0.000000)] bulges=[] normal=[(0.000000,0.000000,1.000000)]
Polyline handle=H flags=0 n=3 closed=0 pts=[(0.000000,0.000000,0.000000);(10.000000,0.000000,0.000000);(10.000000,8.000000,0.000000)] bulges=[] normal=[(0.000000,0.000000,1.000000)]
Polyline handle=H flags=0 n=3 closed=0 pts=[(0.000000,-3.750000,0.000000);(13.750000,-3.750000,0.000000);(13.750000,8.000000,0.000000)] bulges=[] normal=[(0.000000,0.000000,1.000000)]
Polyline handle=H flags=0 n=3 closed=0 pts=[(0.000000,0.000000,0.000000);(10.000000,0.000000,0.000000);(10.000000,8.000000,0.000000)] bulges=[] normal=[(0.000000,0.000000,1.000000)]
Polyline handle=H flags=0 n=3 closed=0 pts=[(0.000000,-2.500000,0.000000);(12.500000,-2.500000,0.000000);(12.500000,8.000000,0.000000)] bulges=[] normal=[(0.000000,0.000000,1.000000)]
Polyline handle=H flags=0 n=3 closed=0 pts=[(0.000000,-6.250000,0.000000);(16.250000,-6.250000,0.000000);(16.250000,8.000000,0.000000)] bulges=[] normal=[(0.000000,0.000000,1.000000)]
Polyline handle=H flags=0 n=3 closed=0 pts=[(0.000000,6.250000,0.000000);(3.750000,6.250000,0.000000);(3.750000,8.000000,0.000000)] bulges=[] normal=[(0.000000,0.000000,1.000000)]
Polyline handle=H flags=0 n=3 closed=0 pts=[(0.000000,3.750000,0.000000);(6.250000,3.750000,0.000000);(6.250000,8.000000,0.000000)] bulges=[] normal=[(0.000000,0.000000,1.000000)]
Polyline handle=H flags=0 n=3 closed=0 pts=[(0.000000,0.000000,0.000000);(10.000000,0.000000,0.000000);(10.000000,8.000000,0.000000)] bulges=[] normal=[(0.000000,0.000000,1.000000)]
Polyline handle=H flags=0 n=4 closed=1 pts=[(32.500000,2.500000,0.000000);(37.500000,2.500000,0.000000);(37.500000,7.500000,0.000000);(32.500000,7.500000,0.000000)] bulges=[] normal=[(0.000000,0.000000,1.000000)]
Polyline handle=H flags=0 n=4 closed=1 pts=[(30.000000,0.000000,0.000000);(40.000000,0.000000,0.000000);(40.000000,10.000000,0.000000);(30.000000,10.000000,0.000000)] bulges=[] normal=[(0.000000,0.000000,1.000000)]
Polyline handle=H flags=0 n=4 closed=1 pts=[(26.250000,-3.750000,0.000000);(43.750000,-3.750000,0.000000);(43.750000,13.750000,0.000000);(26.250000,13.750000,0.000000)] bulges=[] normal=[(0.000000,0.000000,1.000000)]
Warning handle=H flags=0 code=UNSUPPORTED_ENTITY message="MLINE names style VIPRS_G13_MLS_EMPTY which carries no elements, so there is no offset to place a line at and the path alone is not the entity"
Warning handle=H flags=0 code=MLINE_STYLE_FEATURES_IGNORED message="MLINE style VIPRS_G13_MLS_CAPS asks for fill, caps and this version draws the element lines only; the fill, the joint lines and the caps are not emitted"
Polyline handle=H flags=0 n=2 closed=0 pts=[(60.000000,2.500000,0.000000);(70.000000,2.500000,0.000000)] bulges=[] normal=[(0.000000,0.000000,1.000000)]
Polyline handle=H flags=0 n=2 closed=0 pts=[(60.000000,0.000000,0.000000);(70.000000,0.000000,0.000000)] bulges=[] normal=[(0.000000,0.000000,1.000000)]
Polyline handle=H flags=0 n=2 closed=0 pts=[(60.000000,-3.750000,0.000000);(70.000000,-3.750000,0.000000)] bulges=[] normal=[(0.000000,0.000000,1.000000)]
Polyline handle=H flags=1 n=2 closed=0 pts=[(38.658359,1.341641,0.000000);(56.658359,5.841641,0.000000)] bulges=[] normal=[(0.000000,0.000000,1.000000)]
Polyline handle=H flags=1 n=2 closed=0 pts=[(40.000000,0.000000,0.000000);(58.000000,4.500000,0.000000)] bulges=[] normal=[(0.000000,0.000000,1.000000)]
Polyline handle=H flags=1 n=2 closed=0 pts=[(42.012461,-2.012461,0.000000);(60.012461,2.487539,0.000000)] bulges=[] normal=[(0.000000,0.000000,1.000000)]
Polyline handle=H flags=1 n=2 closed=0 pts=[(80.894427,1.788854,0.000000);(68.894427,7.788854,0.000000)] bulges=[] normal=[(0.000000,0.000000,1.000000)]
Polyline handle=H flags=1 n=2 closed=0 pts=[(80.000000,0.000000,0.000000);(68.000000,6.000000,0.000000)] bulges=[] normal=[(0.000000,0.000000,1.000000)]
Polyline handle=H flags=1 n=2 closed=0 pts=[(78.658359,-2.683282,0.000000);(66.658359,3.316718,0.000000)] bulges=[] normal=[(0.000000,0.000000,1.000000)]
```

That is 21 Polylines and two entity warnings, plus the four reader
notifications every fixture in this corpus carries. The block half is SIX
polylines and not three, because `AddBlockInstances` inserts the block twice
and a three-element style crosses once per insertion.

On `real_AC1032.dwg`, measured in the scratch copy rather than guessed: the
three `UNSUPPORTED_ENTITY` lines at handles `38B`, `3A5` and `3A6` become six
Polylines, two per entity, because all three of that drawing's MLINEs use
`Standard` and `Standard` carries two elements. Geometry goes 371 to 377,
Polyline records 24 to 30, warnings 104 to 101, code-100 refusals 65 to 62,
code-109 unchanged at 8, `READER_NOTIFICATION` unchanged at 31. No 115: none of
the three styles sets a fill, joint or cap bit. `real_AC1018.dwg` is the same
drawing saved older and moves the same way.

## What I measured

Two oracles, both on `real_AC1032.dwg`, which AutoCAD wrote, neither sharing a
line of code with the arm. Both agree with the offset side, the justification
references and the miter division, so nothing here was derived and then
believed.

**1. The parameters AutoCAD baked in.** A real file carries, per vertex per
element, `Segments[i].Parameters[0]`, which per the ODA specification is the
distance along the miter to that element's line: exactly the `t` the arm
computes. A throwaway probe in the scratch copy printed both side by side for
all three MLINEs, 12 vertices and 24 element parameters in total. Every pair
agreed to all nine printed decimals. A sample:

```
MLINE handle=38B style=STANDARD elements=2 just=Zero scale=1.500000000 styleflags=None verts=8
  offsets=[0.500000000,-0.500000000]
  v0 dir=(0,1,0) miter=(-1,0,0) dot(M,N)=1.000000000
    seg0 baked[0]=0.750000000  computed_t=0.750000000
    seg1 baked[0]=-0.750000000 computed_t=-0.750000000
  v1 dir=(1,0,0) miter=(-0.707106781,0.707106781,0) dot(M,N)=0.707106781
    seg0 baked[0]=1.060660172  computed_t=1.060660172
    seg1 baked[0]=-1.060660172 computed_t=-1.060660172
MLINE handle=3A5 style=STANDARD elements=2 just=Top scale=2.130943316
  v0 seg0 baked[0]=0.000000000  computed_t=0.000000000
     seg1 baked[0]=-2.130943316 computed_t=-2.130943316
MLINE handle=3A6 style=STANDARD elements=2 just=Zero scale=1.598207487
  v0 seg0 baked[0]=0.799103744  computed_t=0.799103744
     seg1 baked[0]=-0.799103744 computed_t=-0.799103744
```

The sign is the half this pins hardest. Take `38B` v0: the side is
`cross(normal, direction) = cross((0,0,1),(0,1,0)) = (-1,0,0)`, the file's miter
is `(-1,0,0)`, so `dot` is +1 and `t` is +0.75 for the +0.5 offset. The other
side convention, `cross(direction, normal)`, gives `dot` of -1 and `t` of -0.75,
which is not what AutoCAD wrote. Handle `3A5` pins Top the same way: its
reference is `max(0.5, -0.5) = 0.5`, so the first element's `t` is 0 and the
second's is `-1 × 2.130943316`, which is the pair in the file.

**2. ezdxf's `MLine.virtual_entities()`** on a DXF export of the same drawing,
made with ACadSharp's own `DxfWriter` in the scratch copy. The export did not
throw. ezdxf resolved all three MLINEs and reconstructed 18 LINE segments. Every
endpoint matches the corresponding vertex of the Polylines this branch emits:

| Entity | ezdxf | this branch |
| --- | --- | --- |
| 38B element 0 start | `(137.07861533797256, 1.0145784496462795)` | `(137.078615, 1.014578)` |
| 38B element 0 next | `(137.07861533797256, 9.681727104928427)` | `(137.078615, 9.681727)` |
| 38B element 1 start | `(138.57861533797256, 1.0145784496462795)` | `(138.578615, 1.014578)` |
| 3A5 element 0 | `(132.5468878910341, 37.54063454975548)` | `(132.546888, 37.540635)` |
| 3A5 element 1 | `(134.6778312074528, 37.54063454975548)` | `(134.677831, 37.540635)` |
| 3A6 element 0 | `(132.5468878910341, 42.60162492624995)` | `(132.546888, 42.601625)` |
| 3A6 element 1 | `(132.5468878910341, 41.003417438935905)` | `(132.546888, 41.003417)` |

The dump prints six decimals, so that is agreement to everything the dump can
say. The element counts the plan could not predict are settled by the same run:
all three real styles are `STANDARD` with two elements, so three refusals become
six records.

The baked parameters stay an oracle and never enter the product. Two ways of
computing one shape is how the two drift, and a file is free to carry parameters
that disagree with its own style.

**The tests are falsifiable, and I proved it by building the wrong arms.** Not by
reasoning about them: I applied each defect to the arm in a second scratch copy,
rebuilt, redumped and ran this file's 26 tests against the result.

Each row is a build of this arm with one thing changed, dumped into a fresh
copy of this fixture and run against the 26 tests in
`tests/test_flatten_mline.py`. The named test is the one written for that
defect; the count is how many of the 26 it took down.

| The wrong arm | Tests red | The named one |
| --- | --- | --- |
| the bare centre path, as one Polyline, which is the #87 defect | 16 of 26 | `TestTheFixtureCanFail::test_the_bare_path_is_a_different_record_from_every_element` |
| `offset − reference` without `× ScaleFactor` | 7 of 26 | `TestZeroJustification::test_the_scale_factor_is_in_the_offset` |
| offset along `cross(normal, direction)` rather than along `Vertex.Miter` | 6 of 26 | `TestZeroJustification::test_the_bend_follows_the_miter_and_not_the_perpendicular` |

The separation between the last two is the part worth reading. Dropping the
scale moves every vertex, so it takes down the closed-square and both
justification tests as well. Taking the perpendicular instead of the miter is
exactly right at the ends of a straight run and wrong at every bend, so
`test_the_scale_factor_is_in_the_offset` stays GREEN under it, because that
assertion is about the first vertex where the two agree. Two defects, two
distinguishable signatures, and neither test is standing in for the other.

## Two upstream findings

**`PROVENANCE.md` was wrong about MLINE.** It said ACadSharp's
`DwgObjectWriter` has no case for the kind. It does:
`DwgObjectWriter.Entities.cs` dispatches `case MLine mLine: this.writeMLine(mLine);`
and `writeMLine` writes the scale, the justification, the base point, the
normal, the open/closed flag, every vertex with its direction and miter, one
segment per style element per vertex, and a hard pointer to the style. The kind
was forgeable the whole time. I corrected the sentence for MLINE only and left
the other names in that list alone; `PDFUNDERLAY` is on it and
`case PdfUnderlay pdfUnderlay:` is at line 575 of the same file, which is
somebody else's row to check.

**For the composer:** that sentence is a merge conflict waiting to happen. The
3DFACE lane found the same thing about `case Face3D` and edited the same line,
so after both land the count is four and the list is MULTILEADER, 3DSOLID,
REGION and PDFUNDERLAY, with both correction paragraphs kept. Nothing else in
this branch touches a line another lane is likely to.

**`MLine.Clone` in ACadSharp 3.7.1 empties both copies.** `base.Clone` is a
`MemberwiseClone`, so the clone's `Vertices` is the same `List` object as the
original's; the method then calls `clone.Vertices.Clear()` and refills it from
`this.Vertices`, which is by then the list it just emptied. Both come out with
no vertices. `new Insert(block)` clones the block record whenever the record
already belongs to a document, so an MLINE placed in a block before the
insertions exist is silently emptied and writes as an entity carrying nothing.
The first run of this fixture produced exactly that, and the arm reported
`MLINE carries 0 vertices` twice, which is the guard doing its job on a file
that really did say zero. Filling the path after the insertions are made is the
whole of the workaround and it is commented where it happens. Nothing in the
adapter is involved and nothing about this needs a patch in `patches/` for the
shipped library, because the library never clones.

## Red until composition

`git status` on this branch is clean and the committed dumps are untouched, so
the branch is red exactly where the plan says a lane branch is red:

* `acadsharp/tests/test_flatten_mline.py` (all 26) and
  `test_flatten_warnings.py::TestEveryCodeHasAFixture::test_the_fixture_produces_the_code[MLINE_STYLE_FEATURES_IGNORED-g13_mline.dwg]`,
  because `expectations/g13_mline.txt` does not exist until the composer
  regenerates.
* `test_adapter_stream.py::TestNoFixtureIsOutsideEveryCheck::test_the_generator_and_the_manifest_agree_on_what_is_dumped`,
  same reason.
* `test_shim_digest.py` cases naming `native/Adapter/Flatten.MLine.cs`,
  `native/Adapter/Flattener.cs`, `native/Adapter/WarningCodes.cs` and
  `tests/fixtures/gen/Corpus.cs`.
* `test_acadsharp_version.py::TestTheShimRevisionMovesWhenTheShimDoes::test_the_shim_in_the_tree_is_the_one_that_version_names`,
  because `SHIM_DIGESTS["3.7.1-viprs.2"]` is the composer's to re-measure and
  this branch does not touch `VERSION` or that table.
* After the composer regenerates, `test_acadsharp_adr.py::TestTheCensusIsRecomputed`
  (both cases) go red until ADR 0002's census sentence is recomputed, which is
  step 4 of the compose plan. They are green on this branch as it stands.

The scratch copy is how I know nothing else moved. Same container, same
command, four runs, `python:3.12-slim`:

| Tree | `git` in the image | Result |
| --- | --- | --- |
| pristine `de43c6f` | no | 23 failed, 2559 passed, 128 skipped, 13 errors |
| this branch, regenerated | no | 26 failed, 2609 passed, 128 skipped, 13 errors |
| pristine `de43c6f` | yes | **2594 passed, 129 skipped, 0 failed** |
| this branch, regenerated | yes | **3 failed, 2644 passed, 129 skipped** |

The first two rows are the trap. Those 23 failures and 13 errors are identical
line for line in both and all of them are `test_verify_archive.py` and
`test_build_manifests.py` going through `build_acadsharp.driver_commit()`,
which shells out to a `git` that base image does not have. Install `git` and
the pristine tree is a real green, which is the only baseline worth comparing
against. On that baseline this branch adds 50 passing tests and the
only three failures left are the two ADR 0002 census cases and the
`SHIM_DIGESTS` one, which are exactly the two classes listed above as the
composer's to re-measure. Nothing else in the suite moved. `ruff check` and `ruff format --check` pass on the
whole tree in every run.

## What I did not do

* No wire version bump, no header change, no consumer change. 115 is a new
  warning code, which `docs/WIRE.md` says is not a version bump because a
  consumer skips a code it does not know.
* No fill, no joint lines, no caps. Warning 115 says so when the style asks for
  them, which is what 110 already does for a MESH's subdivision level.
* The product code never reads `Segments[].Parameters`. It is an oracle in the
  lane's scratch and nothing else.
* No bulges. An MLINE's elements are straight between its vertices, so record 4
  leaves the array out rather than carrying a run of zeroes.
* Vertices are not lifted through an OCS. Every coordinate on an MLINE is
  already world; the record still names the entity's plane the way the
  `Polyline3D` branch of the `IPolyline` arm does.
* I did not touch `SHIM_DIGESTS`, `VERSION`, `RefusedKinds.cs`, ADR 0002 or
  anything under `tests/expectations/`.
* The dump carries no delimiter between two insertions of one block, so six
  consecutive records under one handle could be two insertions of a
  three-element MLINE or one insertion of a six-element one. That is the
  ambiguity a MESH already has under block expansion rather than a new one, and
  I did not invent a way to say it on wire 2. The test chunks on the element
  count and says so.


spdrman:

PR #108, opened 2026-09-14
Flatten LEADER onto record 4, and say when its arrowhead is missing

### LINKED ISSUE #102: LEADER lowers to an open Polyline through its vertices, with a missing arrowhead said out loud
This is the first step of the group #86 describes. That issue covers MULTILEADER, LEADER and
TOLERANCE together, and this one is only LEADER. MULTILEADER and TOLERANCE are deliberately not
in it: both stay refused, and the design for them goes on #86 as the remainder.

LEADER is the smallest of the three and the only one whose geometry is already in the file as
geometry. `Leader.Vertices` is a run of WCS points, so it lowers onto record 4 as one open
`Polyline` with no new record type, no wire version bump and nothing synthesised. The hook line,
where the drawing recorded one, is already the last vertex in that run, so there is nothing to
work out and nothing to invent.

Two things the arm has to say out loud rather than guess at. An arrowhead is a glyph the dimension
style names, this version draws none, and a consumer that never hears about it draws a leader
that points at nothing, so an enabled arrowhead gets its own warning code beside the polyline. A
spline-fit leader keeps only its fit points, never the curve, so a polyline threaded through those
points is a tessellation by another name and the entity is refused on code 100 instead of being
straightened.

On `real_AC1032.dwg` this is the one LEADER at handle `512`.

## Acceptance

- [ ] A straight LEADER emits one open `Polyline` through the vertices the file holds, hook line
      included when the file recorded one, nothing synthesised, and the entity's plane named on
      the record without the vertices being lifted through it.
- [ ] An enabled arrowhead emits a warning of its own beside the polyline, and a leader with the
      arrowhead off emits none.
- [ ] A spline-fit LEADER is refused on code 100 rather than straightened through its fit points.
- [ ] An isolated fixture carries all three cases at top level and inside a block, and
      `docs/WIRE.md` carries the new warning row and the record 4 paragraph.

spdrman:
LEADER lands on record 4, the `Polyline` wire version 2 already defines, as one open run
through the vertices the file holds. No new record, no header change, no wire version bump:
a LEADER's vertices are world coordinates that are already the geometry, so the only new
thing on the wire is warning code 114, and adding a warning code is the sanctioned
extension channel `docs/WIRE.md` describes ("a consumer skips a code it does not know and
carries on"). A spline-fit leader keeps only fit points, never the curve, so it is refused
on 100 rather than straightened, and an enabled arrowhead gets 114 beside the polyline
because the glyph and its size live in the dimension style rather than in the drawing.

Closes #102

The scoped issue's four boxes, all met on this branch except the dump, which is the
composer's regeneration:

- A straight LEADER emits one open `Polyline` through the vertices the file holds, plane
  named, vertices not lifted: met, `Flatten.Leaders.cs` and
  `test_flatten_leaders.py::TestAStraightLeaderIsOneOpenPolyline` plus
  `TestThePlaneIsNamedAndTheVerticesAreNot`.
- An enabled arrowhead warns and one with the flag off does not: met,
  `TestTheArrowhead`, both halves.
- A spline-fit LEADER is refused rather than straightened: met, `TestASplineFitLeader`,
  with a control asserting its four fit points reach no record at all.
- An isolated fixture and the `WIRE.md` rows: met. `g13_leader.dwg` is committed with its
  PROVENANCE row, and `WIRE.md` carries the 114 row and the record 4 paragraph.

This is the first step of the group issue #86 describes, and only that step. MULTILEADER
and TOLERANCE are untouched and still refused on 100. Their design is the remainder at the
bottom of this body, and it belongs on the group issue rather than here.

## The prediction

Written before the composer's regeneration runs. From my scratch dump of `g13_leader.dwg`,
verbatim except for the four reader notifications every fixture in this corpus opens with:

```
00004 Warning handle=49 flags=0 code=ARROWHEAD_NOT_DRAWN message="LEADER asks for an arrowhead at its first vertex and this version draws the vertex run only; the arrowhead is a glyph the dimension style names and it is not emitted"
00005 Polyline handle=49 flags=0 n=3 closed=0 pts=[(0.000000,0.000000,0.000000);(5.000000,3.000000,0.000000);(9.000000,3.000000,0.000000)] bulges=[] normal=[(0.000000,0.000000,1.000000)]
00006 Polyline handle=4A flags=0 n=2 closed=0 pts=[(20.000000,0.000000,0.000000);(24.000000,5.000000,0.000000)] bulges=[] normal=[(0.000000,0.000000,1.000000)]
00007 Warning handle=4B flags=0 code=UNSUPPORTED_ENTITY message="LEADER is a spline-fit leader and its curve is not in the file as a curve: a polyline through its fit points would be a tessellation by another name, so it is not emitted"
00008 Polyline handle=4C flags=0 n=2 closed=0 pts=[(30.000000,0.000000,0.000000);(34.000000,2.000000,1.000000)] bulges=[] normal=[(0.333333,0.666667,0.666667)]
00009 Warning handle=50 flags=1 code=ARROWHEAD_NOT_DRAWN message="LEADER asks for an arrowhead at its first vertex and this version draws the vertex run only; the arrowhead is a glyph the dimension style names and it is not emitted"
00010 Polyline handle=50 flags=1 n=3 closed=0 pts=[(40.000000,0.000000,0.000000);(49.000000,3.000000,0.000000);(58.000000,3.000000,0.000000)] bulges=[] normal=[(0.000000,0.000000,1.000000)]
00011 Warning handle=50 flags=1 code=ARROWHEAD_NOT_DRAWN message="LEADER asks for an arrowhead at its first vertex and this version draws the vertex run only; the arrowhead is a glyph the dimension style names and it is not emitted"
00012 Polyline handle=50 flags=1 n=3 closed=0 pts=[(80.000000,0.000000,0.000000);(74.000000,4.000000,0.000000);(68.000000,4.000000,0.000000)] bulges=[] normal=[(0.000000,0.000000,1.000000)]
```

13 records: 4 notifications, 5 `Polyline`, 3 `ARROWHEAD_NOT_DRAWN`, 1 `UNSUPPORTED_ENTITY`.
The block half is one leader inserted twice, so it is two of each and not one, and the two
records are told apart by their coordinates rather than by their count: handle `50` comes
out at `(40,0,0)` running `+x` under the 3 by 1.5 insertion and at `(80,0,0)` running `-x`
under the mirrored one. One entity visited twice would be two identical records.

On `real_AC1032.dwg`: line `00210`, `Warning handle=512 ... "LEADER is not a primitive this
version flattens"`, becomes

```
00210 Warning handle=512 flags=0 code=ARROWHEAD_NOT_DRAWN message="LEADER asks for an arrowhead at its first vertex and this version draws the vertex run only; the arrowhead is a glyph the dimension style names and it is not emitted"
00211 Polyline handle=512 flags=0 n=3 closed=0 pts=[(717.868715,2.941179,0.000000);(729.268117,48.538786,0.000000);(740.398750,14.071812,0.000000)] bulges=[] normal=[(0.000000,0.000000,1.000000)]
```

That drawing's one leader has its arrowhead enabled, so the record count goes 475 to 476,
geometry 371 to 372, the warning count stays 104 (one 100 leaves, one 114 arrives), and the
refusal census goes 73 to 72 with code 100 at 64 and code 109 unchanged at 8. POINT stays at
40. `real_AC1018.dwg` is the same drawing saved older and moves identically: 468 to 469
records, geometry 371 to 372, warnings 97.

Those census numbers are ADR 0002's, and re-measuring them is on the composer's list
already. My branch does not touch the ADR and its census tests are green here, because the
committed real dumps have not moved yet.

## What I measured

I have no oracle for LEADER and did not need one: every coordinate in the prediction is a
coordinate the file holds, transformed by a placement the corpus already exercises, so
there is no convention to confirm against AutoCAD the way a wipeout's pixel flip or an
MLINE's offset sign would be. What I measured instead is that the assertions can go red.

Three mutations, each built and dumped in the scratch copy, never in the worktree:

| mutation | what went red |
| --- | --- |
| lift the vertices through `place.WithOcs(leader.Normal, 0.0)` before applying | exactly one test, `test_vertices_are_not_lifted_but_the_plane_is_named`, at `vertex 0 is (-26.832816, 13.416408, 0.0) and the file holds (30.0, 0.0, 0.0)`. Nothing else moved, which is the point of `TestTheLiftWouldBeVisible`: for a normal of `+Z` the algorithm is the identity, so the other three leaders are fixed points of the defect and cannot detect it. |
| straighten the spline-fit leader, close every run, drop the 114 | 12 tests, including both spline ones, all three arrowhead ones and the `closed == 0` assertion |
| walk `AssociatedAnnotation` by calling `Map` from inside the arm | 2 tests, and **only** the two source rules in `TestTheArmExpandsNothing`. The other 19 passed, on the correct dump, which is exactly the claim: an arm that recursed inside `Map` emits byte-identical records and no assertion on a dump can see it. |

Suite numbers, same container both sides, `python:3.12-slim` with `git` added (without `git`
a pristine tree already shows 23 failures and 13 errors, because `build_acadsharp.driver_commit()`
shells out to it):

| tree | result |
| --- | --- |
| pristine `de43c6f` | 2594 passed, 129 skipped, 0 failed |
| this branch, rsynced and regenerated | 2639 passed, 129 skipped, 3 failed |

45 tests added, nothing broken. The three are ADR 0002's two census assertions and
`SHIM_DIGESTS["3.7.1-viprs.2"]`, all three on the composer's re-measure list.

One trap worth passing on: `rsync -a` preserves mtimes, so restoring a mutated `.cs` from
the worktree left it older than the build output and MSBuild skipped the rebuild. The dump
came out silently carrying the previous mutation's behaviour and looked like a real
measurement. `find ... -name '*.cs' -exec touch {} +` after the rsync is the fix, and every
number above was taken after it.

## Red until composition

On this branch, and nothing else:

```
acadsharp/tests/test_flatten_leaders.py (17 of its 21, every one that reads the dump)
acadsharp/tests/test_flatten_warnings.py::TestEveryCodeHasAFixture::test_the_fixture_produces_the_code[ARROWHEAD_NOT_DRAWN-g13_leader.dwg]
acadsharp/tests/test_adapter_stream.py::TestNoFixtureIsOutsideEveryCheck::test_the_generator_and_the_manifest_agree_on_what_is_dumped
acadsharp/tests/test_shim_digest.py::TestTheCapturesAreOfTheShimInTheTree::test_the_recorded_set_is_the_set_the_generator_compiles
acadsharp/tests/test_shim_digest.py::TestTheCapturesAreOfTheShimInTheTree::test_each_source_is_byte_for_byte_what_recorded_the_captures[native/Adapter/Flattener.cs]
acadsharp/tests/test_shim_digest.py::TestTheCapturesAreOfTheShimInTheTree::test_each_source_is_byte_for_byte_what_recorded_the_captures[native/Adapter/WarningCodes.cs]
acadsharp/tests/test_shim_digest.py::TestTheCapturesAreOfTheShimInTheTree::test_the_rollup_covers_the_whole_set
acadsharp/tests/test_acadsharp_version.py::TestTheShimRevisionMovesWhenTheShimDoes::test_the_shim_in_the_tree_is_the_one_that_version_names
```

24 failed, 2599 passed, 128 skipped, measured on a clone of this branch. The four tests in
`test_flatten_leaders.py` that pass here are the source rules, which read the arm rather
than a dump. `native/Adapter/Flatten.Leaders.cs` is a new file and therefore not in the
recorded digest set, which is why the first and last `test_shim_digest` cases are red as
well as the two named ones.

## What is left of the group issue, for whoever takes it next

MULTILEADER. Its geometry is in `MultiLeader.ContextData`: `LeaderRoots[].Lines[].Points`
are WCS and lower onto record 4 the way this arm's vertices do, and
`LeaderRoots[].ConnectionPoint` with `LandingDistance` and `Direction` give the dogleg when
`EnableDogleg` or `EnableLanding` is set. Text content is `HasTextContents` plus
`TextLabel`, `TextLocation`, `TextHeight` and `TextRotation`, which is record 10 under the
MTEXT arm's rules. Block content is `HasContentsBlock` plus `BlockContent` with a location,
a scale, a rotation and a normal, and **that one is the expansion**: it goes onto `Walk`'s
own stack as a `Pending` carrying a composed `Placement` at `Depth + 1` under `CheckDepth`,
never inside `Map`, for the reason the comment above `case Dimension` gives. Arrowheads say
114, which now exists.

The risk that kept it out of this window has not gone away: ACadSharp 3.7.1's MLEADER
context-data reader is the least trusted reader in the pin, and a fixture forged by the
same library round-trips that reader against itself, so a green fixture would say nothing
about the fifteen instances on `real_AC1032.dwg`. The oracle has to be ezdxf's
`MultiLeader.virtual_entities()` over a DXF export of the real drawing, and it has to run
before a single coordinate is pinned.

One correction to the record, verified in the pinned source this round and written into
`tests/fixtures/PROVENANCE.md` as an added paragraph rather than an edit to the sentence it
corrects: that file says six refused kinds are absent from the corpus because
`DwgObjectWriter` has no case for them. That is wrong for MULTILEADER.
`src/ACadSharp/IO/DWG/DwgStreamWriters/DwgObjectWriter.Entities.cs` carries
`case MultiLeader multiLeader:` at line 569 and a real `writeMultiLeader` at line 1611, so
the writer can forge one. The reason to be careful is the reader, not the writer.

TOLERANCE stays on 100. Its frame is computed from the dimension style's text height rather
than read out of the file, so "this is what the drawing holds" is not a promise it can keep,
and the group issue rates that promise weaker itself.

## What I did not do

- No MULTILEADER and no TOLERANCE arm, and nothing touched in `RefusedKinds.cs`.
- No arrowhead drawn, no hook line synthesised, no annotation followed.
- No `VERSION` bump and no `SHIM_DIGESTS` edit. The composer re-measures the viprs.2 row.
- Nothing committed under `acadsharp/tests/expectations/`, and `regenerate.py` was never run
  against the worktree.
- No ADR edit. ADR 0002's census sentence needs the composer's recount, not mine.
- No prologue bit, no header change, no `IsGeometry` change, no new record type.


spdrman:

PR #109, opened 2026-09-14
Flatten WIPEOUT onto record 9, with warning 112 saying it masks

### LINKED ISSUE #89: IMAGE, PDFUNDERLAY and WIPEOUT are refused; dropping a WIPEOUT shows what should be hidden

spdrman:
WIPEOUT lands on record 9, the `Polygon` record that already existed, with
warning 112 immediately before it under the same handle saying that the polygon
is a mask rather than a face. No wire version moves, no record is added and no
consumer has to be rebuilt: a warning code is not a wire bump, because
`docs/WIRE.md` promises a consumer skips a code it does not know and carries on.
WIPEOUT was the one row in `RefusedKinds.cs` on code 100 rather than 109, which
ADR 0002 recorded as "not a decision at all, just nobody has done it", so the
row comes out of the table and the ADR's opening goes from eight kinds to seven.

Closes #89

* **A WIPEOUT emits its boundary with the masking intent expressed.** Met. The
  clip boundary crosses as a closed `Polygon` and warning 112 sits beside every
  one of them under the same handle. On `real_AC1032.dwg` handle `44D` stops
  being a refusal.
* **A decision recorded on IMAGE and PDFUNDERLAY.** Already met before this
  lane, and this restates rather than moves it: both stay on code 109 with the
  rows in `docs/adr/0002-what-this-decoder-refuses.md` and the paragraph in
  `docs/WIRE.md` about the external ones. Neither row is touched here. What
  those two need is a placement record on a later wire, whose byte layout is
  pinned in ADR 0003 as record 16.
* **A placement record carrying the frame.** Not this lane and not this wire
  version. It is wire 3, ADR 0003, and the epic filed beside it.

## The prediction

`g13_wipeout.dwg`, handles replaced by `H` and warning text elided:

```
Warning handle=H flags=0 code=POLYGON_MASKS message="WIPEOUT is a mask, ..."
Polygon handle=H flags=0 n=4 closed=1 pts=[(0.000000,4.000000,0.000000);(10.000000,4.000000,0.000000);(10.000000,0.000000,0.000000);(0.000000,0.000000,0.000000)] bulges=[] normal=[(0.000000,0.000000,-1.000000)]
Warning handle=H flags=0 code=POLYGON_MASKS message="WIPEOUT is a mask, ..."
Polygon handle=H flags=0 n=3 closed=1 pts=[(20.000000,6.000000,0.000000);(30.000000,6.000000,0.000000);(26.000000,0.000000,0.000000)] bulges=[] normal=[(0.000000,0.000000,-1.000000)]
Warning handle=H flags=0 code=POLYGON_MASKS message="WIPEOUT is a mask, ..."
Polygon handle=H flags=0 n=4 closed=1 pts=[(36.000000,3.000000,0.000000);(42.000000,11.000000,0.000000);(46.000000,8.000000,0.000000);(40.000000,0.000000,0.000000)] bulges=[] normal=[(0.000000,0.000000,-1.000000)]
Warning handle=H flags=1 code=POLYGON_MASKS message="WIPEOUT is a mask, ..."
Polygon handle=H flags=1 n=3 closed=1 pts=[(61.000000,0.000000,0.000000);(70.000000,6.000000,0.000000);(40.000000,7.500000,0.000000)] bulges=[] normal=[(0.000000,0.000000,1.000000)]
Warning handle=H flags=1 code=POLYGON_MASKS message="WIPEOUT is a mask, ..."
Polygon handle=H flags=1 n=3 closed=1 pts=[(66.000000,0.000000,0.000000);(60.000000,8.000000,0.000000);(80.000000,10.000000,0.000000)] bulges=[] normal=[(0.000000,0.000000,-1.000000)]
```

Four `READER_NOTIFICATION` lines precede those, as they do for every fixture
this generator writes. Five masks, not four: `AddBlockInstances` inserts one
block twice, so the one wipeout inside it is two records and two warnings.

On `real_AC1032.dwg`, line `00209`, which reads

```
Warning handle=44D flags=0 code=UNSUPPORTED_ENTITY message="WIPEOUT is not a primitive this version flattens, ..."
```

becomes

```
Warning handle=44D flags=0 code=POLYGON_MASKS message="WIPEOUT is a mask, ..."
Polygon handle=44D flags=0 n=4 closed=1 pts=[(271.921129,3.988123,0.000000);(298.113345,4.239772,0.000000);(295.720786,26.510687,0.000000);(272.676675,35.695866,0.000000)] bulges=[] normal=[(0.000000,0.000000,1.000000)]
```

Measured on the regenerated scratch copy, not predicted, for both real drawings:
geometry 371 to 372, `Polygon` 158 to 159, warnings unchanged at 104 (1032) and
97 (1018) because one code 100 leaves and one 112 arrives, code 100 refusals 65
to 64, code 109 unchanged at 8, total refusals 73 to 72, POINT refusals
unchanged at 40. So ADR 0002's census sentence needs 371/104/73/65/8 recomputed
to 372/104/72/64/8 and its Point bullet needs "40 of the 73" to become "40 of
the 72". I have deliberately left both for the composer's single re-measure,
because this lane is not the only thing moving them.

## What I measured

The y-flip in `insert + u·(px + 0.5) + v·(size_y − py − 0.5)` is a convention,
and this fixture cannot prove a convention: the same generator writes the
numbers and reads them back, so a round trip agrees with itself whichever way
the flip goes. So ezdxf 1.4.4 was asked the same question three ways, all in a
container, nothing committed.

**One, the fixture's own frames, forwards.** A WIPEOUT built in ezdxf with the
insert point, the two per-pixel vectors, the image size and the pixel boundary
each fixture entity carries, then `boundary_path_wcs()`:

| entity | ezdxf says | y-flipped | un-flipped |
| --- | --- | --- | --- |
| polygonal, insert (20,0,0) | (20,6,0) (30,6,0) (26,0,0) | same, max diff 0 | (20,0,0) (30,0,0) (26,6,0), off by 6 |
| block, insert (0,0,0) | (7,0,0) (10,4,0) (0,5,0) | same, max diff 0 | (7,5,0) (10,1,0) (0,0,0), off by 5 |
| rectangular, insert (0,0,0) | (0,4,0) (10,4,0) (10,0,0) (0,0,0) | same four, same order | same four, other order |
| rotated, insert (40,0,0) | (36,3,0) (42,11,0) (46,8,0) (40,0,0) | same four, same order | same four, other order |

The two rectangular rows are also where ezdxf confirms the corner expansion:
handed two opposite corners it returns four, in the order (a.x, a.y),
(b.x, a.y), (b.x, b.y), (a.x, b.y), which is what the arm does.

**Two, backwards.** `msp.add_wipeout()` on the world triangle
(20,6,0) (30,6,0) (26,0,0) encodes it as insert (20,0,0), u (10,0,0),
v (0,6,0), image_size (1,1) and pixel boundary (-0.5,-0.5) (0.5,-0.5)
(0.1,0.5). That is character for character the second entity of this fixture,
which I had written by hand from the plan's formula before running this. Two
implementations sharing one sign error would not agree in both directions.

**Three, on data AutoCAD wrote.** `real_AC1032.dwg`'s WIPEOUT, handle `44D`,
read straight out of the DWG: insert (271.9211288808192, 3.988123351850959, 0),
u (31.707742926072683, 0, 0), v (0, 31.707742926072683, 0), size (1, 1),
polygonal, definition null, four vertices (-0.5, 0.5),
(0.3260511008800897, 0.49206350233605894),
(0.2505944934505404, -0.21031745948902658),
(-0.47617156637515023, -0.5). Handed to ezdxf unchanged it answers exactly what
this build emits for it, to 3.5e-7, which is the six decimals the dump prints.
The un-flipped mapping is 31.707743 away from that.

So the oracle agrees with the plan's mapping, on my numbers and on AutoCAD's,
and there is nothing here to stop on.

Two things fell out of that AutoCAD entity that are worth carrying forward.
Its `Size` is (1, 1), so the plan's guess that a wipeout's image size is one
pixel each way is right for a real file and the fixture matches it. And the
IMAGE next to it, handle `8C6`, has `Size` (994, 965) with rectangular corners
(-0.5, -0.5) and (993.5, 964.5), which is upstream's documented default
(-0.5, -0.5) to (size − 0.5) exactly, and is where the `size_y` term in the
mapping does real work rather than cancelling to a half. **That bears on ADR
0003.** Its record 16 `Placement` carries clip vertices as fractions along `u`
and `v` with the pixel flip applied by the producer, which is the same
transform this arm applies, so this run is evidence for that decision as well.
A producer writing record 16 for handle `8C6` has to divide by 994 and 965 and
flip against 965, and getting the flip wrong there moves a boundary by the
height of the image rather than by the half pixel a wipeout with `Size` (1,1)
would forgive.

**The test is red against the wrong answer, measured rather than asserted.** I
built the arm with `v * (py + 0.5)` in the scratch copy, regenerated, and ran
the suite. Four cases go red and eight stay green:

```
FAILED test_flatten_masks.py::TestThePixelMapping::test_the_polygonal_mask_lands_where_the_pixel_mapping_puts_it
FAILED test_flatten_masks.py::TestThePixelMapping::test_a_rectangular_clip_is_four_corners
FAILED test_flatten_masks.py::TestThePixelMapping::test_a_rotated_frame_follows_u_and_v
FAILED test_flatten_masks.py::TestBlockCopies::test_the_block_copies_cross_and_the_mirror_flips_the_measured_normal
4 failed, 8 passed
```

The un-flipped build emitted `(20,0,0) (30,0,0) (26,6,0)` for the polygonal
entity, which is the named wrong answer, and `(0,0,0) (10,0,0) (10,4,0)
(0,4,0)` for the rectangular one, which is the same rectangle in a different
order. `TestTheFixtureCanFail` is the control for both halves: it asserts the
two mappings put the polygonal boundary at least 1.0 apart and are not
permutations of each other, and separately that the rectangle IS a fixed point
of the flip, so nobody later moves the mapping assertion onto it.

## Red until composition

Measured on the branch tree in a container, not guessed: 18 failed, 2595
passed, 129 skipped. The expectations are the committed ones, because a lane
never regenerates. Every one of the 18 is in the classes section 4.0 names:

```
test_acadsharp_version.py::TestTheShimRevisionMovesWhenTheShimDoes::test_the_shim_in_the_tree_is_the_one_that_version_names
test_adapter_stream.py::TestNoFixtureIsOutsideEveryCheck::test_the_generator_and_the_manifest_agree_on_what_is_dumped
test_flatten_masks.py::TestThePixelMapping::test_the_polygonal_mask_lands_where_the_pixel_mapping_puts_it
test_flatten_masks.py::TestThePixelMapping::test_a_rectangular_clip_is_four_corners
test_flatten_masks.py::TestThePixelMapping::test_the_rectangular_mask_covers_the_area_it_should
test_flatten_masks.py::TestThePixelMapping::test_a_rotated_frame_follows_u_and_v
test_flatten_masks.py::TestThePixelMapping::test_no_mask_carries_a_bulge_array
test_flatten_masks.py::TestTheMaskIsSaidToBeAMask::test_every_mask_polygon_has_a_112_beside_it_with_the_same_handle
test_flatten_masks.py::TestTheMaskIsSaidToBeAMask::test_the_code_says_what_to_do_with_it
test_flatten_masks.py::TestBlockCopies::test_the_block_copies_cross_and_the_mirror_flips_the_measured_normal
test_flatten_masks.py::TestWipeoutIsNoLongerRefused::test_nothing_in_this_fixture_refuses_a_wipeout
test_flatten_masks.py::TestWipeoutIsNoLongerRefused::test_the_fixture_is_geometry_and_warnings_and_nothing_else
test_flatten_warnings.py::TestEveryCodeHasAFixture::test_the_fixture_produces_the_code[POLYGON_MASKS-g13_wipeout.dwg]
test_shim_digest.py::TestTheCapturesAreOfTheShimInTheTree::test_the_recorded_set_is_the_set_the_generator_compiles
test_shim_digest.py::TestTheCapturesAreOfTheShimInTheTree::test_each_source_is_byte_for_byte_what_recorded_the_captures[native/Adapter/Flattener.cs]
test_shim_digest.py::TestTheCapturesAreOfTheShimInTheTree::test_each_source_is_byte_for_byte_what_recorded_the_captures[native/Adapter/RefusedKinds.cs]
test_shim_digest.py::TestTheCapturesAreOfTheShimInTheTree::test_each_source_is_byte_for_byte_what_recorded_the_captures[native/Adapter/WarningCodes.cs]
test_shim_digest.py::TestTheCapturesAreOfTheShimInTheTree::test_the_rollup_covers_the_whole_set
```

Nine of those are the new module, which reads a dump that does not exist on a
lane branch. `test_flatten_masks.py::TestTheFixtureCanFail` passes there, which
is the point of it: it compares two mappings rather than a dump.

The three-way measurement, all in the same container with `git` installed,
because without `git` in the image `build_acadsharp.driver_commit()` shells out
to something that is not there and 23 unrelated cases go red for no reason:

| tree | result |
| --- | --- |
| pristine `de43c6f` | 2594 passed, 129 skipped, 0 failed |
| this branch, committed expectations | 2595 passed, 129 skipped, 18 failed |
| this branch, rsynced and regenerated | 2630 passed, 129 skipped, 3 failed |

The three on the regenerated copy are
`test_acadsharp_adr.py::TestTheCensusIsRecomputed::test_every_stated_count_is_the_one_in_the_corpus`,
`::test_the_point_bullet_counts_the_same_refusals` and
`test_acadsharp_version.py::TestTheShimRevisionMovesWhenTheShimDoes::test_the_shim_in_the_tree_is_the_one_that_version_names`,
which are exactly the two numbers on the composer's re-measure list and
nothing else. So this branch adds 36 passing cases and breaks none.

One trap worth recording for whoever runs this protocol next. `rsync -a`
preserves mtimes, so restoring a `.cs` file from the worktree over an
already-built copy can leave it older than the build output, MSBuild skips the
rebuild, and the dump you take is the previous build's behaviour wearing a
fresh timestamp. I re-took every number in this section and in the two above
after `find copy/acadsharp/native -name '*.cs' -exec touch {} +` and a forced
rebuild. `g13_wipeout.txt` came back byte for byte identical and every census
count was unchanged, so nothing here was measuring the wrong binary, but the
only reason I can say that is that I checked.

## What I did not do

* No prologue flag bit. Record 9's spare bits and the prologue's are a semantic
  change under an unchanged wire version, which is the one thing the version
  number and the ABI fingerprint exist to prevent, and `payload.rs:124` in the
  Rust conformance consumer asserts prologue `flags == 0` on the probe stream,
  so at least one consumer author already read the prologue's silence as a
  rule. ADR 0003 reserves bit 1 on wire 3 for exactly this, so the code and the
  bit can agree later.
* No change to the IMAGE or PDFUNDERLAY rows in `RefusedKinds.cs` or ADR 0002,
  and no change to the ADR's census numbers.
* No path opened, nothing resolved. A wipeout names no file and this arm reads
  none.
* No `VERSION` bump and no `SHIM_DIGESTS` edit.
* Nothing under `acadsharp/tests/expectations/`.
* Nothing degenerate in the fixture. The arm refuses a boundary of fewer than
  three vertices on code 100 with a sentence that starts `WIPEOUT`, and no
  entity in this corpus triggers it: it is a bound on hostile input rather than
  a case, and code 100 itself is covered by `g13_unsupported.dwg`.

## Two things the composer needs from this lane

**`PROVENANCE.md` is edited additively.** I did not touch the contended
sentence that begins "Six refused kinds are absent here". WIPEOUT was never on
that list, so I had nothing to change in it. I added a paragraph after it
instead, which says two things: PDFUNDERLAY does not belong on the list either,
and the WIPEOUT fixture needed a detour. The PDFUNDERLAY half I verified myself
in the pinned source: `DwgObjectWriter.Entities.cs` has
`case PdfUnderlay pdfUnderlay:` at line 575, `writePdfUnderlay` at 1920 writes
the normal, insertion point, rotation, three scales, flags, contrast, fade,
definition handle and clip boundary, and `UnderlayEntity` adds no `IsValid`
override, so nothing filters one out. With Lane A taking 3DFACE out, Lane C
taking MLINE out and PDFUNDERLAY going too, that sentence should end up reading
three kinds: MULTILEADER, 3DSOLID and REGION. Its stated reason is wrong for
those three as well (`case MultiLeader` is at 569 and `case Solid3D` at 618,
and `isEntitySupported` filters Solid3D and Region rather than the final arm
throwing), so it is worth one more pass by whoever owns it next.

**`DwgWriter` does not take a `Wipeout` with a null `Definition`, and that is
not an exception.** `CadWipeoutBase.IsValid` refuses any instance whose
`Definition` or `DefinitionReactor` is null, and
`DwgObjectWriter.isEntitySupported` calls it and drops the entity from the
block's entity list, so the file writes cleanly and comes back empty. Measured:
both null wrote 10507 bytes and read back 0 entities; a throwaway
`ImageDefinition` wrote 10604 bytes and read back 0 entities, because
`DefinitionReactor` is still null and its setter and both of its constructors
are internal to `ACadSharp.dll`. That rule is `RasterImage`'s and it is simply
wrong about a wipeout, which has no raster by definition and which AutoCAD
writes with a zero definition handle (handle `44D` on the real drawing has
`definition=null`). So `Corpus.cs` writes one through a subclass that overrides
`IsValid` and changes nothing else: the same `writeCadImage` runs, and the
reader gives back a plain `Wipeout`. Worth knowing because ACadSharp's DXF
writer applies the same gate at `DxfSectionWriterBase.Entities.cs:141`, which is
why the first ezdxf run found zero wipeouts in a DXF export of this fixture and
why the oracle asks ezdxf directly instead.


spdrman:

